PR #2, opened 2019-05-28
[utils] fixing how cprint handles tuples

llevitis:
This PR fixes how cprint handles tuples as previously the following code would cause the ADNI to BIDS conversion to crash if there were multiple `key_preferred_visit`. https://github.com/aramis-lab/clinica/blob/fbdb62724c02033abb2b26c970d10569d6f2f823/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_t1.py#L523-L526

### DIFF
--- a/clinica/utils/stream.py
+++ b/clinica/utils/stream.py
@@ -34,11 +34,14 @@ class FilterOut(object):
 def active_cprint():
     sys.stdout = FilterOut(sys.stdout)
 
-
 def cprint(msg):
     global clinica_verbose
     if clinica_verbose is True:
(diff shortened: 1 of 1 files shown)


arnaudmarcoux:
Hello, 
The converters have been greatly improved (including your suggestion). We recommand that you perform the developper installation ( http://www.clinica.run/doc/Installation/ ) in order to use the **dev** branch of clinica that contains those modifications. They are now less buggy and support multiprocessing.
Thank you for your interest in Clinica !

Arnaud

PR #10, opened 2019-07-19
Remove duplicates during oasis-to-bids conversion

14thibea:
- remove duplicates from participants.tsv
- ensure that the image converted corresponds to the alternative_id_1


alexandreroutier:
Thanks for the PR. It is now merged on the `dev` branch of Clinica.

PR #16, opened 2019-08-26
Sort diagnosis list in ML datasets

14thibea:
Will avoid to find 1 - accuracy when computing the metrics on a test set.


alexandreroutier:
Thanks for this bugfix!

In the future, do not forget to to sync your fork with the `aramis-lab` version of Clinica. Otherwise, when you will do e.g. a new PR, your previous contributions will be present (in your case the commit regarding "Remove duplicates during oasis-to-bids conversion").

PR #19, opened 2019-08-30
t1-volume-existing-template update with iterables 

arnaudmarcoux:
This pull request make some modification to allow the pipeline `T1VolumeExistingTemplate` to fail for some subjects without affecting the others. 

The files are grabbed with the use of `glob`, allowing to have a list of input files, eventually leading to the utilization of `iterables`.

The main problem to deal with were the containers of the datasink. A trick had to be found. Container variable of datasink is now linked to input T1w file through a function that give away the container name based on the filename.

Work to do left (but not necessary for this merge request) :  
- replace MapNode by node when necessary.
- @alexandreroutier needs to test wether the intended behavior is working (crash on one of the file : do the other keep running ?

Attached : 
- new diagram
- working directory with an axemple of unnecessary MapNode (creating mapflows even though there will always be only one instance here thanks to previous iterable)
t1volexTemplate
Screen Shot 2019-08-30 at 18 44 43

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -863,10 +863,10 @@ def convert_flair(folder_input, folder_output, name, fixed_file=False):
             flair_path = glob(path.join(flair_lst[0], '*.nii.gz*'))[0]
             copy(flair_path, path.join(folder_output, name + (get_bids_suff('Flair')) + '.nii.gz'))
         elif len(flair_lst) == 0:
-                return -1
+            return -1
         elif len(flair_lst) > 1:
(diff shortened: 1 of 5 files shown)


alexandreroutier:
I replaced a T1w image by an empty file and this PR managed to save images where pipeline was successful to the CAPS folder!

It is not clear for me why we don't need to move the content of `build_output_node` to `build_core_nodes` (cf PR #14) nor we don't need to convert MapNodes into Nodes for `build_core_nodes` but it works.

To investigate when `t1-volume*`  will be refactored one day...

PR #28, opened 2019-09-09
[WIP] New DWI-Preprocessing pipelines using FSL eddy

alexandreroutier:
This PR contains new versions of DWI-Preprocessing pipelines. They all rely on `eddy` tool [[Andersson et al., 2016](https://dx.doi.org/10.1016%2Fj.neuroimage.2015.10.019)] from FSL

- A very close version of `dwi-preprocessing-using-t1w` was evaluated in [[Wen et al., Submitted](https://arxiv.org/abs/1812.11183)]
- `dwi-preprocessing-using-fmap` needs to be tested on more subjects.


alexandreroutier:
Follow up on PR #59.

PR #31, opened 2019-09-18
pet-surface update for docker

arnaudmarcoux:
pet-surface can now be used with SPM standalone on mac and Linux (tested on latest version of SPM standalone for mac and linux, and MCR v713 for mac and MCR v2018b for linux)

SPM standalone is used when `SPMSTANDALONE_HOME` and `MCR_HOME` are found in the `os.environ.keys()`

The key idea is to create a batch file that SPM can read. The command to run a batch file with spm standalone is :
`$SPMSTANDALONE_HOME/run_spm12.sh /path/to/MCR batch /path/to/batch/file.m`

BUT :  
In Mac Os : the script `run_spm12.sh` must be run from its very own folder, leading to a command like : 
`cd $SPMSTANDALONE_HOME && ./run_spm12.sh /path/to/MCR batch /path/to/batch/file.m`  

The batch file is created on the fly when the function `runApplyInverseDeformationField_SPM_standalone` is run.

The file ApplyInverseDeformationField.m  could not be used (like before), as it does not have the syntax to be recognised as a spm batch file. 

If SPM standalone is not used, the pipeline is executed as before.

To use SPM_STANDALONE in the rest of the pipeline (like for the Coregister, or Normalize12), we use the standard method : 
```python
matlab_cmd = ('cd ' + os.path.expandvars('$SPMSTANDALONE_HOME') + ' && ' + './run_spm12.sh'
                                  + ' ' + os.environ['MCR_HOME']
                                  + ' script')
spm.SPMCommand.set_mlab_paths(matlab_cmd=matlab_cmd, use_mcr=True)
```
 and we change `matlab_cmd` in the same way as previously explained if we are on a Linux machine

### DIFF
--- a/clinica/pipelines/pet_surface/pet_surface_pipeline.py
+++ b/clinica/pipelines/pet_surface/pet_surface_pipeline.py
@@ -322,7 +322,7 @@ class PetSurface(cpe.Pipeline):
         if all(elem in os.environ.keys() for elem in ['SPMSTANDALONE_HOME', 'MCR_HOME']):
             if os.path.exists(os.path.expandvars('$SPMSTANDALONE_HOME')) and os.path.exists(os.path.expandvars('$MCR_HOME')):
                 print(Fore.GREEN + 'SPM standalone has been found and will be used in this pipeline' + Fore.RESET)
-                matlab_cmd = (os.path.join(os.environ['SPMSTANDALONE_HOME'], 'run_spm12.sh')
+                matlab_cmd = ('cd ' + os.path.expandvars('$SPMSTANDALONE_HOME') + ' && ' + './run_spm12.sh'
                               + ' ' + os.environ['MCR_HOME']
                               + ' script')
                 spm.SPMCommand.set_mlab_paths(matlab_cmd=matlab_cmd, use_mcr=True)
--- a/clinica/pipelines/pet_surface/pet_surface_pipeline.py
+++ b/clinica/pipelines/pet_surface/pet_surface_pipeline.py
@@ -276,7 +276,7 @@ class PetSurface(cpe.Pipeline):
                                                           'desikan_right',
                                                           'destrieux_left',
                                                           'destrieux_right',
-                                                          'use_SPM_standalone'],
+                                                          'use_spm_standalone'],
                                              output_names=[],
                                              function=utils.get_wf),
                                 name='full_pipeline_mapnode',
@@ -319,8 +319,8 @@ class PetSurface(cpe.Pipeline):
 
         full_pipe.inputs.matscript_folder_inverse_deformation = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
 
-        # This section of code determine wether to use SPM standalone or not
-        full_pipe.inputs.use_SPM_standalone = False
+        # This section of code determines wether to use SPM standalone or not
+        full_pipe.inputs.use_spm_standalone = False
         if all(elem in os.environ.keys() for elem in ['SPMSTANDALONE_HOME', 'MCR_HOME']):
             if os.path.exists(os.path.expandvars('$SPMSTANDALONE_HOME')) and os.path.exists(os.path.expandvars('$MCR_HOME')):
                 print(Fore.GREEN + 'SPM standalone has been found and will be used in this pipeline' + Fore.RESET)
@@ -333,7 +333,7 @@ class PetSurface(cpe.Pipeline):
                                   + ' ' + os.environ['MCR_HOME']
                                   + ' script')
                 spm.SPMCommand.set_mlab_paths(matlab_cmd=matlab_cmd, use_mcr=True)
-                full_pipe.inputs.use_SPM_standalone = True
+                full_pipe.inputs.use_spm_standalone = True
             else:
                 raise FileNotFoundError('$SPMSTANDALONE_HOME and $MCR_HOME are defined, but linked to non existent folder ')
 
--- a/clinica/pipelines/pet_surface/pet_surface_pipeline.py
+++ b/clinica/pipelines/pet_surface/pet_surface_pipeline.py
@@ -319,8 +319,8 @@ class PetSurface(cpe.Pipeline):
         full_pipe.inputs.matscript_folder_inverse_deformation = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
 
         full_pipe.inputs.use_SPM_standalone = False
-        if 'SPMSTANDALONE_HOME' and 'MCR_HOME' in os.environ.keys():
-            if os.path.exists(os.path.expandvars('SPMSTANDALONE_HOME')) and os.path.exists(os.path.expandvars('MCR_HOME')):
+        if all(elem in os.environ.keys() for elem in ['SPMSTANDALONE_HOME', 'MCR_HOME']):
+            if os.path.exists(os.path.expandvars('$SPMSTANDALONE_HOME')) and os.path.exists(os.path.expandvars('$MCR_HOME')):
                 print(Fore.GREEN + 'SPM standalone has been found and will be used in this pipeline' + Fore.RESET)
                 matlab_cmd = (os.path.join(os.environ['SPMSTANDALONE_HOME'], 'run_spm12.sh')
                               + ' ' + os.environ['MCR_HOME']
--- a/clinica/pipelines/pet_surface/pet_surface_pipeline.py
+++ b/clinica/pipelines/pet_surface/pet_surface_pipeline.py
@@ -255,6 +255,8 @@ class PetSurface(cpe.Pipeline):
         import nipype.pipeline.engine as npe
         import nipype.interfaces.utility as niu
         import clinica.pipelines.pet_surface.pet_surface_utils as utils
+        from colorama import Fore
+        from nipype.interfaces import spm
 
         full_pipe = npe.MapNode(niu.Function(input_names=['subject_id',
                                                           'session_id',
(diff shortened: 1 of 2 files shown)


clinica/pipelines/pet_surface/pet_surface_utils.py
```diff
@@ -872,7 +914,7 @@ def get_wf(subject_id,
                            name='vol2vol_mask')
 
     if 'SPMSTANDALONE_HOME' in os.environ:
-        tpmnii = os.path.join(str(os.getenv("SPM_HOME")), 'spm12_mcr/spm12/spm12/tpm/TPM.nii')
```
arnaudmarcoux: If `SPMSTANDALONE_HOME` exists, `SPM_HOME` does not necessarily exists. + The path with `spm12/spm12` was not correct

alexandreroutier:
Your fork is not synchronized with the main repo. Could you please update it?

In your case, I think you would need to do a cherry pick of  `10b6cad ` on `upstream/dev`. The commands would be something like this:
```
git fetch upstream
git rebase -i upstream/dev
# Chose your commit(s) and solve possible conflict(s)
git push --force
```

clinica/pipelines/pet_surface/pet_surface_utils.py
alexandreroutier: On the `dev` branch of Clinica, there is now a `get_tpm()` function (see [spm.py](https://github.com/aramis-lab/clinica/blob/dev/clinica/utils/spm.py)) to extract the tissue probability map. You should use/modify this to avoid duplicated lines.

Besides, I am currently refactoring some code while I'm finding a way to run ReconAll for regression tests. Anyway, I have noticed this:
GetSpmHome

I extracted the lines into a single `get_spm_home()` function but the latter assumes that `SPM_HOME` exists based on lines. 

This would mean that there is a possible bug in `atlas.py` when using SPM Standalone....

arnaudmarcoux: The path was right for Linux system indeed. See below.

For `get_atlas_labels()`: the lines

```python 
            if "SPMSTANDALONE_HOME" in os.environ:
                if platform.system() == 'Darwin':
                    SPM_HOME = os.environ['SPMSTANDALONE_HOME'] + "/spm12.app/Contents/MacOS/spm12_mcr"
``` 
make no sense to me. Here is the content of `$SPMSTANDALONE_HOME/spm12.app/Contents/MacOS/` : 
`applauncher  prelaunch    spm12.ctf    spm12_maci64`

clinica/pipelines/pet_surface/pet_surface_pipeline.py
```diff
@@ -272,7 +275,8 @@ def build_core_nodes(self):
                                                           'desikan_left',
                                                           'desikan_right',
                                                           'destrieux_left',
-                                                          'destrieux_right'],
+                                                          'destrieux_right',
+                                                          'use_SPM_standalone'],
```
alexandreroutier: This does not follow PEP8 convention, this should be `use_spm_standalone`.

clinica/pipelines/pet_surface/pet_surface_pipeline.py
```diff
@@ -315,6 +319,24 @@ def build_core_nodes(self):
 
         full_pipe.inputs.matscript_folder_inverse_deformation = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
 
+        # This section of code determine wether to use SPM standalone or not
```
alexandreroutier: Typos:
```suggestion
        # This section of code determines whether to use SPM standalone or not
```

clinica/pipelines/pet_surface/pet_surface_pipeline.py
```diff
@@ -315,6 +319,24 @@ def build_core_nodes(self):
...
+                                  + ' script')
+                elif platform.system() == 'Linux':
+                    matlab_cmd = (os.path.join(os.path.expandvars('$SPMSTANDALONE_HOME'), 'run_spm12.sh')
+                                  + ' ' + os.environ['MCR_HOME']
+                                  + ' script')
+                spm.SPMCommand.set_mlab_paths(matlab_cmd=matlab_cmd, use_mcr=True)
+                full_pipe.inputs.use_SPM_standalone = True
+            else:
+                raise FileNotFoundError('$SPMSTANDALONE_HOME and $MCR_HOME are defined, but linked to non existent folder ')
+
```
alexandreroutier: Naive question: I have the feeling that these lines will be reused at least in T1Volume* and PETVolume pipelines. Would it be possible to factorize these resources into a function?

clinica/pipelines/pet_surface/pet_surface_utils.py
```diff
@@ -200,10 +200,60 @@ def isclose(a, b, rel_tol=1e-9, abs_tol=0.0):
     return list_of_regions
 
 
+def runApplyInverseDeformationField_SPM_standalone(target,
```
alexandreroutier: Following PEP8 convention, this should be named `run_apply_inverse_deformation_field_spm_standalone` but I understand you did that to comply with `applyInverseDeformationField.m` Matlab file.

(For this one, I let you decide to modify or not)

clinica/pipelines/pet_surface/pet_surface_utils.py
```diff
@@ -886,12 +942,16 @@ def get_wf(subject_id,
                                       warping_regularization=[0, 0.001, 0.5, 0.05, 0.2]),
                           name='normalize_to_MNI')
 
+    if use_SPM_standalone is True:
+        fun_apply_inverse_deformation = utils.runApplyInverseDeformationField_SPM_standalone
```
alexandreroutier: What does the `fun_` prefix mean?

clinica/pipelines/pet_surface/pet_surface_utils.py
```diff
@@ -200,10 +200,60 @@ def isclose(a, b, rel_tol=1e-9, abs_tol=0.0):
...
+        raise SystemError('Only support Mac OS and Linux')
+    os.system(cmdline)
+
+    output_file = join(abspath('./'), prefix + basename(img))
+    if not exists(output_file):
+        raise IOError('Something went wrong while trying to run runApplyInverseDeformationField_SPM_standalone'
+                      + '. Output file not generated. Command launched :\n\t '
+                      + cmdline + '\n. We strongly recommand that you use the supported version of Matlab MCR '
+                      + ' recommanded by the creators of SPM.')
+    return output_file
```
alexandreroutier: I have no expertise for the `runApplyInverseDeformationField_SPM_standalone` function. If you want a feedack, I would advise you to ask @mdiazmel :-)

clinica/pipelines/pet_surface/pet_surface_utils.py
```diff
@@ -872,7 +923,12 @@ def get_wf(subject_id,
                            name='vol2vol_mask')
 
     if 'SPMSTANDALONE_HOME' in os.environ:
-        tpmnii = os.path.join(str(os.getenv("SPM_HOME")), 'spm12_mcr/spm12/spm12/tpm/TPM.nii')
+        # Path is different between SPM standalone MAC and Linux. First is MAC location
+        if platform.system() == 'Darwin':
+            tpmnii = os.path.join(str(os.getenv("SPMSTANDALONE_HOME")), 'spm12_mcr/spm12/tpm/TPM.nii')
+        elif platform.system() == 'Linux':
+            # Try Linux location
+            tpmnii = os.path.join(str(os.getenv("SPMSTANDALONE_HOME")), 'spm12_mcr/spm12/spm12/tpm/TPM.nii')
```
alexandreroutier: Why don't you use the `get_tpm()` function in `utils/spm.py` ?

clinica/pipelines/pet_surface/pet_surface_utils.py
alexandreroutier: In that case, I would advice to create an issue regarding this misleading point and fix this once for all.

Would that suit you?

alexandreroutier:
Hi,

I think I finished my feedback. Some remarks regarding PEP8 convention and code redundancy needs to be addressed.

Alexandre

clinica/pipelines/pet_surface/pet_surface_utils.py
arnaudmarcoux: This function does not seem to work if SPM standalone  only is installed (at least on a Mac machine). See screenshot below
Screen Shot 2019-09-24 at 15 01 59
All my environment variables are set properly though...
As I said, there is really some work to do to have spm standalone in order.(Maybe this is caused by the newest version of spm standalone, who knows !)

arnaudmarcoux:
Hi, 

Comments have been addressed.

clinica/pipelines/pet_surface/pet_surface_pipeline.py
alexandreroutier: I understand. I will add a card regarding this risky point.

In that case, could you create a `get_tpm()` function in `pet_surface_utils.py`? This will improve the readability of your code, decrease the complexity and will also help the person who will harmonize code regarding SPM-Standalone.

clinica/pipelines/pet_surface/pet_surface_utils.py
alexandreroutier: See my response [here](https://github.com/aramis-lab/clinica/pull/31#discussion_r327720728).

I close this discussion.

clinica/pipelines/pet_surface/pet_surface_pipeline.py
arnaudmarcoux: `get_tpm()` has been corrected in an other PR, (this commit https://github.com/aramis-lab/clinica/pull/35/commits/f7281fd042539bbae66ab8aeb64743b7fe6cc208)
I'll update the pipeline once the remove_grabbit PR is merged

PR #35, opened 2019-09-23
Remove grabbit library from Clinica

arnaudmarcoux:
This PR will replace grabbit library in Clinica by custom function that will eventually be more robust. We will then do not rely on an external library and remove a dependency, which is always good.

The 2 functions implemented are located in clinica/utils/inputs.py:

`clinica_file_reader` and `clinica_group_reader` that correspond to the 2 distincts usages we have when grabbing files.

clinica_file_reader returns a list of strings that matches the criterions, and clinica_group_reader a string.

**Integration in a pipeline :** 
If the pipeline must handle multiple kind of files, the code could go as follows in `build_input_node()` : 

```python 
all_errors = []
# First file to grab
try:
    read_parameters_node.inputs.pet = clinica_file_reader(self.subjects,
                                                             self.sessions,
                                                             self.bids_directory,
                                                             {'pattern': '*' + self.parameters['pet_type'] + '_pet.nii*',
                                                             'description': 'PET file})
except ClinicaException as e:
    all_errors.append(e)

# Grab template
try:
    final_template = clinica_group_reader(self.caps_directory, {'pattern': 'group-' + g_id + '_template.nii*', 'description': 'constructed template', 'needed_pipeline': 't1-volume or t1-volume-create-dartel')
except ClinicaException as e:
    all_errors.append(e)

# If any error happened, we display it at the end so that the user is aware of all 
# the problems at once
if len(all_errors) > 0:
    error_message = 'Clinica encountered errors while trying to read files in your BIDS or CAPS directories.\n'
    for msg in all_errors:
        error_message += str(msg)
    raise ClinicaError(error_message)
```

That produces the following error message  (outdated)
Screen Shot 2019-09-26 at 17 50 08

This new way of grabbing files have been applied to :  
- pet-surface
- t1-volume-existing-template
- t1-volume-tissue-segmentation
- machine-learning-spatial-svm
- t1-volume-create-dartel
- dwi-connectome
- dwi-dti
- dwi-preprocessing-using-phasediff-fieldmaps
- pet-volume (needs testing with pvc)
- t1-volume-parcellation
- t1-volume
- dwi-preprocessing-using-t1
- fmri-preprocessing (data for functional test must be changed because it exists a confusion with 2 pairs of files: `sub-01001TMM_ses-M00_task-rest_bold.(json/nii.gz)` and `sub-01001TMM_ses-M00_bold.(json/nii.gz)`
- t1-volume-existing-dartel
-t1-volume-dartel2mni


arnaudmarcoux:
How to solve this issue without over complicating the function ? 
Screen Shot 2019-09-23 at 16 47 20

alexandreroutier:
> All the code for grabbing files can now be replaced by one line : (here example of PET surface)
> Screen Shot 2019-09-23 at 17 04 37

I am aware that it is WIP but do you plan to notify the user that, if a CAPS file is missing, he must run the corresponding pipeline?

alexandreroutier:
When I rewrote `t1-freesurfer`, I paid attention to remove `fsaverage` &
cie symbolic links so it would not be present with newer runs of this
pipeline.

However, for old runs, I don't have a clear solution in mind. The ideal
solution would be to remove the symbolic link once and for all. What do you
think?



Le lun. 23 sept. 2019 à 16:47, Arnaud Marcoux <notifications@github.com> a
écrit :

arnaudmarcoux:
Well actually I found a workaround (illustrated in the screenshot at the top of this page).

For the other comment regarding warning properly the user, I'm working on it.

arnaudmarcoux:
@alexandreroutier I updated the first message of the PR, can you tell me if you find this kind of solution ok to you ?

alexandreroutier:
> Well actually I found a workaround (illustrated in the screenshot at the top of this page).
> 
> For the other comment regarding warning properly the user, I'm working on it.

I looked at your screenshot presenting exceptions. Some information are not useful for the users (e.g. "until level 10"). Besides, you do not explain for CAPS files from what pipeline it came from nor which pipeline he/she should run.

Previous work was done in `utils/io.py`. So you don't start from scratch regarding this point.

clinica/pipelines/dwi_connectome/dwi_connectome_pipeline.py
```diff
@@ -56,75 +56,98 @@ def build_input_node(self):
...
-        dwi_files = caps_layout.get(type='dwi', suffix='preproc', extensions=['.nii|.nii.gz'], return_type='file',
-                                    subject=participants_regex, session=sessions_regex)
-        error_message += check_input_caps_files(dwi_files, "DWI_PREPROC_NII", "dwi-preprocessing",
-                                                self.caps_directory, self.subjects, self.sessions)
+        dwi_files, err_msg = clinica_file_reader(self.subjects,
+                                                 self.sessions,
+                                                 self.caps_directory,
+                                                 {'pattern': '*_dwi_space-*_preproc.nii*',
+                                                  'description': 'preprocessed DWI',
+                                                  'needed_pipeline': 'dwi-preprocessing-using-t1 or dwi-preprocessing-using-fieldmap'})
```
alexandreroutier: What about using a dictionary instead?

When you will modify `dwi-dti`, you will copy/paste these informations and if one day, the CAPS, the description or the name of the pipeline is changing, you will have to change each line where the file is grabbed instead of modifying the dictionary once.

arnaudmarcoux: I'd rather duplicate a bit of code for the following reasons:
- The number of occurrences of grabbing the same file is quite low
- I find it very convenient to have all the required information to make the clinica_file_reader call without external call 
- For the same reason, it is then easier to maintain 
- CAPS changing names or pipeline changing name is (kind of) unlikely
- If this ever happens, it is really a matter of seconds to change it with Cmd + Shift + F in Pycharm (see below)
Screen Shot 2019-10-11 at 14 02 44

clinica/utils/inputs.py
```diff
@@ -0,0 +1,223 @@
...
+                                        {'pattern': 'rh.white',
+                                         'description': 'right hemisphere of outter cortical surface.',
+                                         'needed_pipeline': 't1-freesurfer'})
+                        the following error will arise:
+                        * More than 1 file found::
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/fsaverage/surf/rh.white
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/rh.EC_average/surf/rh.white
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/sub-ADNI011S4105_ses-M00/surf/rh.white
+                Correct usage (wanted for pet-surface): pattern string must be 'sub-*_ses-*/surf/rh.white')
+                    It then gives: ['/caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/sub-ADNI011S4105_ses-M00/surf/rh.white']
```
alexandreroutier: Actually, you still need to add `freesurfer_cross_sectional` in the regexp. Indeed, you will have multiple files with the upcoming `t1-freesurfer-longitdudinal`.

clinica/utils/inputs.py
```diff
@@ -0,0 +1,223 @@
...
+                raise RuntimeError('Could not determine if ' + input_dir + ' is a CAPS or BIDS directory')
+
+        else:
+            if len([sub for sub in listdir(input_dir) if (sub.startswith('sub-') and isdir(join(input_dir, sub)))]) > 0:
+                return True
+            else:
+                if isdir(join(input_dir, 'groups')):
+                    return False
+                else:
+                    raise RuntimeError('Could not determine if ' + input_dir + ' is a CAPS or BIDS directory')
```
alexandreroutier: This can be extracted from `clinica_file_reader` and be used for other functions/methods :)

alexandreroutier: BTW, you are not using Google Python Style for the docstrings.

clinica/utils/inputs.py
```diff
@@ -0,0 +1,223 @@
...
+    # Some check on the formatting on the data
+    assert pattern[0] != '/', 'pattern argument cannot start with char : / (does not work in os.path.join function). ' \
+                              + 'If you want to indicate the exact name of the file, use the format' \
+                              + ' directory_name/filename.extension or filename.extensionin the pattern argument'
+    assert len(subjects) == len(sessions), 'Subjects and sessions must have the same length'
+    if len(subjects) == 0:
+        raise RuntimeError('No subjects and sessions provided.')
+
+    # rez is the list containing the results
+    rez = []
```
alexandreroutier: Could be more readable:
```suggestion
    results = []
```

clinica/pipelines/dwi_connectome/dwi_connectome_pipeline.py
alexandreroutier: > * The number of occurrences of grabbing the same file is quite low

This is not a good argument - DRY (Don’t Repeat Yourself):
> If you have identical code in 3 or more places, it should be abstracted into a common method, class, etc. to be reusable and easily maintainable.

 :wink:


> * I find it very convenient to have all the required information to make the clinica_file_reader call without external call
> * For the same reason, it is then easier to maintain

You can have this behaviour using e.g. dictionaries, enum or classes mechanism.


> * CAPS changing names or pipeline changing name is (kind of) unlikely
> * If this ever happens, it is really a matter of seconds to change it with Cmd + Shift + F in Pycharm

It will eventually change when CAPS will be merged into BIDS-Derivatives. I agree that it is fast but it easier to modify the regexp once for all than to modify N times the regexp (e.g. 6 for the example with T1w files)

Before that, it is more likely that a new pipeline will be created and you will need to grab files. In this scenario, it will be very complicated to enforce that the developer uses the same description if the same files have been grabbed in another pipeline.

alexandreroutier: If my comment regarding grabbing files from `t1-freesurfer` pipeline is correct, you can see than you need to modify each pipeline to fix the regexp (i.e. `fmri-pipeline`, `pet-surface`, `dwi-connectome`).

In general, each possible error in a regexp  currently needs to be fixed into the involved pipeline(s).

alexandreroutier:
Hi @arnaudmarcoux,

I made a general review. My two main comments concerns 1) duplicated dictionaries when grabbing e.g. T1w files and 2) how error message is handled in `build_input_node()`.

Point 1) has been adressed in previous comments. For the second point, when looking at a `build_input_node()` method e.g. `dwi-preprocessing-using-t1`:

```python
        all_errors = []
        t1w_files, err_msg = clinica_file_reader(self.subjects,
                                                 self.sessions,
                                                 self.bids_directory,
                                                 {'pattern': '*_t1w.nii*',
                                                  'description': 'T1w MRI acquisition'})
        if err_msg:
            all_errors.append(err_msg)

        dwi_files, err_msg = clinica_file_reader(self.subjects,
                                                 self.sessions,
                                                 self.bids_directory,
                                                 {'pattern': 'dwi/*_*_*dwi.nii*',
                                                  'description': 'DWI NIfTI files'})
        if err_msg:
            all_errors.append(err_msg)

        bval_files, err_msg = clinica_file_reader(self.subjects,
                                                  self.sessions,
                                                  self.bids_directory,
                                                  {'pattern': 'dwi/*_dwi.bval',
                                                   'description': 'bval files'})
        if err_msg:
            all_errors.append(err_msg)

        bvec_files, err_msg = clinica_file_reader(self.subjects,
                                                  self.sessions,
                                                  self.bids_directory,
                                                  {'pattern': 'dwi/*_dwi.bvec',
                                                   'description': 'bvec files'})
        if err_msg:
            all_errors.append(err_msg)

        if len(all_errors) > 0:
            error_message = 'Clinica faced error(s) while trying to read files in your BIDS directory.\n'
            for msg in all_errors:
                error_message += msg
            raise ClinicaBIDSError(error_message)
```
We don't use pythonic mechanism of exceptions. My first proposition would be to modify `clinica_file_reader` such that it generates `Clinica{BIDS[CAPS}Error` then it would be caught e.g. 

```python
        exception_raised = False
        try:
            t1w_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': '*_t1w.nii*',
                                             'description': 'T1w MRI acquisition'})

            dwi_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': 'dwi/*_*_*dwi.nii*',
                                             'description': 'DWI NIfTI files'})
            # bval files
            bval_files = clinica_file_reader(self.subjects,
                                             self.sessions,
                                             self.bids_directory,
                                             {'pattern': 'dwi/*_dwi.bval',
                                              'description': 'bval files'})
            # bvec files
            bvec_files = clinica_file_reader(self.subjects,
                                             self.sessions,
                                             self.bids_directory,
                                             {'pattern': 'dwi/*_dwi.bvec',
                                              'description': 'bvec files'})

        except ClinicaBIDSError as e:
            if not exception_raised:
                cprint('Clinica faced error(s) while trying to read files in your BIDS directory.')
            exception_raised = True
            cprint(e)
        finally:
            if exception_raised:
                raise ClinicaBIDSError()
```
In this approach, it turns out that lines with try...except... are skipped as soon as an exception is raised. So this is equivalent to:

```python
        try:
            t1w_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': '*_t1w.nii*',
                                             'description': 'T1w MRI acquisition'})
            dwi_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': 'dwi/*_*_*dwi.nii*',
                                             'description': 'DWI NIfTI files'})
            # bval files
            bval_files= clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': 'dwi/*_dwi.bval',
                                             'description': 'bval files'})
            # bvec files
            bvec_files= clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': 'dwi/*_dwi.bvec',
                                             'description': 'bvec files'})

        except ClinicaBIDSError as e:
            cprint('Clinica faced error(s) while trying to read files in your BIDS directory.')
            raise e
        else:
            cprint("Clinica will be run on the following images...")
```
From that, I see two possibilities:
- Provide the instructions to generate a TSV file with the required files e.g.
```python
        [...]
        except ClinicaBIDSError as e:
            cprint("Clinica found missing image(s) and/or several images per participant/session.")
            cprint(e)
            cprint("To run this pipelines with required files, type:.")
            cprint("clinica iotools create-subjects-visits --modalities \"T1w DWI\" <BIDS> <OUTPUT_TSV> ")
            cprint("Then, re-run the command with -tsv <OUTPUT_TSV> option.")
            raise ClinicaBIDSError()
```

- Use try...except... to each "grab":
```python
        all_errors = []
        try:
            t1w_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': '*_t1w.nii*',
                                             'description': 'T1w MRI acquisition'})
        except ClinicaBIDSError as e:
            all_errors.append(e)

        try:
            dwi_files = clinica_file_reader(self.subjects,
                                            self.sessions,
                                            self.bids_directory,
                                            {'pattern': 'dwi/*_*_*dwi.nii*',
                                             'description': 'DWI NIfTI files'})
        except ClinicaBIDSError as e:
            all_errors.append(e)

        try:
            # bval files
            bval_files = clinica_file_reader(self.subjects,
                                             self.sessions,
                                             self.bids_directory,
                                             {'pattern': 'dwi/*_dwi.bval',
                                              'description': 'bval files'})
        except ClinicaBIDSError as e:
            all_errors.append(e)

        try:
            # bvec files
            bvec_files = clinica_file_reader(self.subjects,
                                             self.sessions,
                                             self.bids_directory,
                                             {'pattern': 'dwi/*_dwi.bvec',
                                              'description': 'bvec files'})
        except ClinicaBIDSError as e:
            all_errors.append(e)

        if len(all_errors) > 0:
            cprint('Clinica faced error(s) while trying to read files in your BIDS directory.')
            raise ClinicaBIDSError(all_errors)
```
The second proposition is very close to what we have currently except that it would be more pythonic. The first proposition would shorten each `build_input_node()`  method but would need to develop tools to create TSV based on specific BIDS/CAPS files.


What do you think?

arnaudmarcoux:
Regarding your last comment, we already have one usecase where `len(subject_list)` is expected to be different than the number of files we are grabbing. And we still need the file list.

In `t1_freesurfer_cli.py`
```python 
        # Extract subject IDs generated by the pipeline on CAPS folder
        caps_files, err_msg = clinica_file_reader(participant_ids,
                                                  session_ids,
                                                  self.absolute_path(args.caps_directory),
                                                  {'pattern': 'sub-*_ses-*/scripts/recon-all.log',
                                                   'description': 'recon all log'})

        # Extract missing subject IDs
        input_ids = [participant_id + '_' + session_id
                     for participant_id, session_id in zip(participant_ids, session_ids)]
        output_ids = extract_image_ids(caps_files)
        missing_ids = list(set(input_ids) - set(output_ids))
```

Raising an exception the way you showed it will not work in this specific case.
**Workaround** : add an optional argument in clinica_file_reader `raise_exception=True`. I have seen some python function use this trick like below :  

Screen Shot 2019-10-15 at 17 27 21

After that, it remains a more philosophical question. Should we raise an error in Clinica if a file is missing ? Considering that it is dependant on the user, this could be considered as normal behavior from clinica to say that the files were not found because they are not here. Actually we could only display the error message followed by `sys.exit(0)`, avoiding the big "clinica crash" in red, that contains a lot of useless information in this specific case (like the traceback). What do you think ?

Do you have any ressources that say it is more pythonic the way you presented it ? It's hard to get hands on this kind of material.

clinica/utils/inputs.py
arnaudmarcoux: Done

clinica/utils/inputs.py
arnaudmarcoux: Done

alexandreroutier:
I forgot this case where, for each pipeline, we will check images where it crashed. I would say that `raise_exception=True` is a great workaround: it would be the default behaviour since your function is mainly used for `build_input_node()` methods and we would use `raise_exception=False` for the case you describe. In the end, I would get rid of `err_msg ` as a return parameter of `clinica_file_reader`.


> After that, it remains a more philosophical question. Should we raise an error in Clinica if a file is missing ? Considering that it is dependant on the user, this could be considered as normal behavior from clinica to say that the files were not found because they are not here. Actually we could only display the error message followed by `sys.exit(0)`, avoiding the big "clinica crash" in red, that contains a lot of useless information in this specific case (like the traceback). What do you think ?

I would say (I can change my mind tomorrow ):
- If `participant_ids/session_ids` are not specified, it would grab as many files as possible without generating exceptions.
- If not, `raise_exception=True` would be the default behaviour.

Regarding the long Traceback you mentionned, it will not be displayed with any subclass of `ClinicaException` does not display it. (see `engine.py` :))



> Do you have any ressources that say it is more pythonic the way you presented it ? It's hard to get hands on this kind of material.

The single `try...except...` is standard in Python. For the several `try..except` case, examples are few (I found e.g. here: https://www.python.org/dev/peps/pep-0463/) but the `all_errors.append(e)` is new.

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+                 'description': 'Desikan-based segmentation (mri/aparc.a2009s+aseg.mgz)',
+                 'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESIKAN_SURF_L = {'pattern': 'sub-*_ses-*/*/lh.aparc.annot',
+                        'description': 'left hemisphere surface-based Desikan parcellation (lh.aparc.annot)',
+                        'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESIKAN_SURF_R = {'pattern': 'sub-*_ses-*/*/rh.aparc.annot',
+                        'description': 'right hemisphere surface-based Desikan parcellation (rh.aparc.annot)',
+                        'needed_pipeline': 't1-freesurfer'}
```
alexandreroutier: You need to add `freesurfer_cross_sectional` in the regexp. Otherwise, you will conflicts with `t1-freesurfer-longitudinal` pipeline.

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+               'description': 'FDG-PET data'}
+
+PET_FDG_JSON = {'pattern': 'pet/sub-*_ses-*_acq-fdg_pet.json',
+                'description': 'json file describing the point spread function (PSF) in FDG PET.'}
+
+PET_AV45_NII = {'pattern': 'pet/sub-*_ses-*_acq-av45_pet.nii*',
+                'description': 'AV45-PET data'}
+
+PET_AV45_JSON = {'pattern': 'pet/sub-*_ses-*_acq-av45_pet.json',
+                 'description': 'json file describing the point spread function (PSF) in AV45 PET.'}
```
alexandreroutier: json should be uppercase since it's an acronym:
```suggestion
                 'description': 'JSON file describing the point spread function (PSF) in AV45 PET.'}
```
(To do for all descriptions)

clinica/utils/inputs.py
```diff
@@ -0,0 +1,247 @@
...
+    pattern = information['pattern']
+    is_bids = determine_caps_or_bids(input_directory)
+
+    if is_bids:
+        check_bids_folder(input_directory)
+    else:
+        check_caps_folder(input_directory)
+
+    # Some check on the formatting on the data
+    assert pattern[0] != '/', 'pattern argument cannot start with char : / (does not work in os.path.join function). ' \
```
alexandreroutier:
```suggestion
    assert pattern[0] != '/', 'pattern argument cannot start with char: / (does not work in os.path.join function). ' \
```
or 
```suggestion
    assert pattern[0] != '/', 'pattern argument cannot start with char slash (/) (does not work in os.path.join function). ' \
```

clinica/utils/inputs.py
```diff
@@ -0,0 +1,247 @@
...
+                                        caps,
+                                        {'pattern': 'rh.white',
+                                         'description': 'right hemisphere of outter cortical surface.',
+                                         'needed_pipeline': 't1-freesurfer'})
+                        the following error will arise:
+                        * More than 1 file found::
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/fsaverage/surf/rh.white
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/rh.EC_average/surf/rh.white
+                            /caps/subjects/sub-ADNI011S4105/ses-M00/t1/freesurfer_cross_sectional/sub-ADNI011S4105_ses-M00/surf/rh.white
+                Correct usage (wanted for pet-surface): pattern string must be 'sub-*_ses-*/surf/rh.white')
```
alexandreroutier:
```suggestion
                Correct usage (wanted e.g. for pet-surface): pattern string must be 'freesurfer_cross_sectional/sub-*_ses-*/surf/rh.white')
```

clinica/utils/dwi.py
```diff
@@ -330,6 +330,7 @@ def hmc_split(in_file, in_bval, ref_num=0, lowbval=5.0):
 
 
 def check_dwi_volume(in_dwi, in_bvec, in_bval):
+    # TODO return str instead of raising an error, so that user can be informed of which file is causing problem
```
alexandreroutier: In upcoming new version `dwi-preprocessing`, it will be done at `init_node` pipeline. With this solution, compliant DWI files will run even though this is not the case for "non-compliant" DWI files.

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+
+T1_FS_WS_L = {'pattern': 'sub-*_ses-*/surf/lh.white',
+              'description': 'left hemisphere of outter cortical surface.',
+              'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESTRIEUX = {'pattern': 'sub-*_ses-*/mri/aparc.a2009s+aseg.mgz',
+                   'description': 'Destrieux-based segmentation (mri/aparc.a2009s+aseg.mgz)',
+                   'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESTRIEUX_SURF_L = {'pattern': 'sub-*_ses-*/*/lh.aparc.a2009s.annot',
```
alexandreroutier: To be coherent with files in `mri` folder:
```suggestion
T1_FS_DESTRIEUX_SURF_L = {'pattern': 'sub-*_ses-*/surf/lh.aparc.a2009s.annot',
```

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+                          'description': 'left hemisphere surface-based Destrieux parcellation (lh.aparc.a2009s.annot)',
+                          'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESTRIEUX_SURF_R = {'pattern': 'sub-*_ses-*/*/rh.aparc.a2009s.annot',
+                          'description': 'right hemisphere surface-based Destrieux parcellation (rh.aparc.a2009s.annot)',
```
alexandreroutier: To be coherent with files in `mri` folder:
```suggestion
                          'description': 'right hemisphere surface-based Destrieux parcellation (label/rh.aparc.a2009s.annot)',
```

clinica/utils/inputs.py
alexandreroutier: TODO: Update description with `freesurfer_cross_sectional`

clinica/utils/input_files.py
arnaudmarcoux: Done

clinica/utils/input_files.py
arnaudmarcoux: Done

clinica/utils/inputs.py
arnaudmarcoux: Done

clinica/utils/inputs.py
```diff
@@ -0,0 +1,247 @@
...
+
+    if is_bids:
+        check_bids_folder(input_directory)
+    else:
+        check_caps_folder(input_directory)
+
+    # Some check on the formatting on the data
+    assert pattern[0] != '/', 'pattern argument cannot start with char : / (does not work in os.path.join function). ' \
+                              + 'If you want to indicate the exact name of the file, use the format' \
+                              + ' directory_name/filename.extension or filename.extensionin the pattern argument'
```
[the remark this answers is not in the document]
arnaudmarcoux: Done

clinica/utils/inputs.py
arnaudmarcoux: Done

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+            'description': 'segmentation of white matter (mri/wm.seg.mgz)',
+            'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_BRAIN = {'pattern': 'sub-*_ses-*/mri/brain.mgz',
+               'description': ' extracted brain from T1w MRI',
+               'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_ORIG_NU = {'pattern': 'mri/orig_nu.mgz',
+                 'description': 'intensity normalized volume generated after correction for'
+                                ' non-uniformity in FreeSurfer (file .../mri/orig_nu.mgz) ',
```
alexandreroutier:
```suggestion
                                ' non-uniformity in FreeSurfer (file mri/orig_nu.mgz) ',
```

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+                 'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_WS_R = {'pattern': 'sub-*_ses-*/surf/rh.white',
```
alexandreroutier: I think `T1_FS_WS_R` should be coherent with `T1_FS_WM` defined above. What about something like:
- `T1_FS_R_WM_SURF`: this proposition would be very close to the [WIP BEP11](https://docs.google.com/document/d/1YG2g4UkEio4t_STIBOqYOwneLEs1emHIXbGKynx7V0Y/edit#)
- `T1_FS_WM_SURF_R`
- `T1_FS_SURF_WM_R`

clinica/utils/dwi.py
arnaudmarcoux: Ok

clinica/utils/input_files.py
alexandreroutier: `T1_FS_WS_R` should be coherent with `T1_FS_WM` defined above. What about something like:
- `T1_FS_R_WM_SURF` : this would be coherent with the [WIP BEP11](https://docs.google.com/document/d/1YG2g4UkEio4t_STIBOqYOwneLEs1emHIXbGKynx7V0Y/edit#) 
- `T1_FS_WM_SURF_R` 
- `T1_FS_SURF_WM_R` 
?

clinica/utils/input_files.py
arnaudmarcoux: Done

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+              'description': 'right hemisphere of outter cortical surface.',
```
alexandreroutier: Wrong ;). Proposition:
```suggestion
              'description': 'right white matter/gray matter border surface.',
```

clinica/utils/inputs.py
arnaudmarcoux: Done

clinica/utils/input_files.py
arnaudmarcoux: Done

clinica/utils/input_files.py
arnaudmarcoux: I coherence with `T1_FS_WM` I changed to`T1_FS_WM_SURF_R`

arnaudmarcoux: See above

clinica/utils/input_files.py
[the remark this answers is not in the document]
arnaudmarcoux: Done

clinica/utils/input_files.py
arnaudmarcoux: Done

arnaudmarcoux:
Comments have been addressed

clinica/utils/input_files.py
```diff
@@ -0,0 +1,128 @@
...
+
+T1_FS_WM_SURF_L = {'pattern': 't1/freesurfer_cross_sectional/sub-*_ses-*/surf/lh.white',
+                   'description': 'left white matter/gray matter border surface.',
+                   'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESTRIEUX = {'pattern': 't1/freesurfer_cross_sectional/sub-*_ses-*/mri/aparc.a2009s+aseg.mgz',
+                   'description': 'Destrieux-based segmentation (mri/aparc.a2009s+aseg.mgz)',
+                   'needed_pipeline': 't1-freesurfer'}
+
+T1_FS_DESTRIEUX_SURF_L = {'pattern': 't1/freesurfer_cross_sectional/sub-*_ses-*/label/lh.aparc.a2009s.annot',
```
alexandreroutier: These are discrete parcellations and not surfaces. From [BEP 11](https://docs.google.com/document/d/1YG2g4UkEio4t_STIBOqYOwneLEs1emHIXbGKynx7V0Y/edit#heading=h.y9xsm9g2uzp5), you can find use `DPARC` for discrete parcellation or simply `PARC`, it is up to you:
```suggestion
T1_FS_DESTRIEUX_PARC_L = {'pattern': 't1/freesurfer_cross_sectional/sub-*_ses-*/label/lh.aparc.a2009s.annot',
```
(Final update for Desikan needs to be done and we can merge this PR)

arnaudmarcoux: Done

arnaudmarcoux:
When CI is ok you can merge this !

PR #44, opened 2019-10-24
Change how to find NaN values in images

mdiazmel:
This PR changes the way how NaN values are found when loading an image.


alexandreroutier:
Hi,

I think the `data1 != data1` statement instead of `np.isnan(data1)` is also performed in other utils/pipelines.

@arnaudmarcoux: Could you confirm that this appears in other Clinica location(s)?

Alex

alexandreroutier:
I think I found one:
```
clinica/pipelines/pet_surface/pet_surface_utils.py:93:    nan_mask = data != data
```
I ran `git grep -n "!="` in Clinica and I did not see others :)

mdiazmel:
```
clinica/pipelines/pet_surface/pet_surface_utils.py:93:    nan_mask = data != data
```
Not sure that this data type has the `isnan` method (it's an image loaded by nibabel)... Need to be verified.

PR #45, opened 2019-10-24
Check relative offset of PET and T1w when coregistering with SPM

arnaudmarcoux:
Datasets were generated to see how SPM managed to coregister t1w and fdg pet.

3 dataset were created and tested on a simple pipeline that only coregister the 2 modalities with SPM.

**1st dataset**: PET is centered on the origin and t1w image is away 
results : when distance from origin is at more than 100 mm for t1w, coregistration fail (no errror message, but output is not what it should be) 
Screen Shot 2019-10-24 at 17 02 30

**2nd dataset**: t1w is centered on the origin and pet is away from the center.
results : when distance from origin is at more than 100 mm for pet, coregistration fail

**3rd dataset**: Same as first dataset, but all the volume are shifted at a fixed offset.
results: SPM is sensible to the relative distance between the 2 volumes, not to the fact that the 2 volumes must be centered

We implement a new function `check_relative_volume_location_in_world_coordinate_system` that checks, before coregistering, that 2 list of files are not too far apart (pairwise). If not, the following message is displayed:

Screen Shot 2019-10-24 at 16 52 25

Please note that a flag `--coreg_two_modalities` has been added, along with the new function `center_nifti_for_coreg`


clinica/iotools/utils/data_handling.py
```diff
@@ -658,6 +658,111 @@ def center_all_nifti(bids_dir, output_dir, modality, center_all_files=False):
...
+    return nifti_to_center
+
+
+def are_far_appart(file1, file2, threshold=80):
+    """
+    Tells if 2 files have a center located at more than a threshold distance
+    Args:
+        file1: (str) path to the first nifti file
+        file2: (str) path to the second nifti file
+        threshold: threshold to consider wether 2 files are too far appart
```
alexandreroutier:
```suggestion
        threshold: threshold to consider whether 2 files are too far appart
```

clinica/iotools/utils/data_handling.py
```diff
@@ -658,6 +658,111 @@ def center_all_nifti(bids_dir, output_dir, modality, center_all_files=False):
...
+
+    Returns:
+
```
alexandreroutier:
```suggestion
        True if distance between `file1` and `file2` are far from `threshold`, False otherwise.
```

alexandreroutier:
Hi @arnaudmarcoux ,

I have a few questions/comments for the  `--coreg_two_modalities` flag:
- Why this is not proposed for `pet-surface` pipeline? Some coregistration steps are done between FreeSurfer-T1w images and PET images. :thinking: 
- For `pet-volume`, the user has to run first the `t1-volume` pipeline. My concern concerns the facts that he will have uncentered-T1w-derived files in CAPS folder and centered-T1w/PET-derived files in CAPS folder as well. 

I don't know if my second point was clear but this could also happen in `pet-surface` pipeline. Do you have some insights to handle that?

clinica/iotools/utils/data_handling.py
arnaudmarcoux: Done

clinica/iotools/utils/data_handling.py
arnaudmarcoux: Done

arnaudmarcoux:
> Why this is not proposed for pet-surface pipeline? Some coregistration steps are done between FreeSurfer-T1w images and PET images. 🤔

I added 2 things:
-> The warning on uncentered t1w file appear when trying to launch t1-freesurfer, but specifying ```
f you are trying to launch the t1-freesurfer pipeline, you can ignore this message if you do not want to run the pet-surface pipeline afterward
```
-> I only check for the centering of PET data before running pet-surface. That way, only PET data is modified

> For pet-volume, the user has to run first the t1-volume pipeline. My concern concerns the facts that he will have uncentered-T1w-derived files in CAPS folder and centered-T1w/PET-derived files in CAPS folder as well.

Yes I see what you mean, segmentation in native space will not be able to be overlaid into newly centered t1w. I see one solution: We cannot touch t1w, so we center only pet file if they are off (but with a smaller threshold, like 30, or 40 mm). Because keep in mind that for t1-volume, the warning message appear, so we can consider t1w images to be centered. (at a maximum distance of 80 mm).

File: clinica/iotools/utils/data_handling.py [hunk 3]
```diff
@@ -682,7 +705,78 @@ def write_list_of_files(file_list, output_file):
     return output_file
 
 
-def check_volume_location_in_world_coordinate_system(nifti_list, bids_dir):
+def check_relative_volume_location_in_world_coordinate_system(label_1, nifti_list1,
+                                                              label_2, nifti_list2,
+                                                              bids_dir,
+                                                              modality):
+    import numpy as np
```
alexandreroutier: Explanations on label_1/2 and modality are needed.

PR #52, opened 2019-11-21
Remove duplicated MRtrix3 interfaces

alexandreroutier:
Based on the presentation of https://sonarqube.inria.fr/ this morning, I removed unused and duplicated MRtrix3 interfaces and centralize them into `lib/nipype/interfaces` 😇 

In a second time, time would be needed to check if this is now handled in newer versions of Nipype or PR otherwise.


mdiazmel:
Hi @alexandreroutier, thanks for doing this. 
I reviewed and everything seems alright. 
We can safely merge (I can do it of @arnaudmarcoux is too busy).

alexandreroutier:
Hi @mdiazmel,

I let you do the merge in that case :)

Alex

PR #58, opened 2020-01-08
Rewrite t1-freesurfer-longitudinal pipeline

alexandreroutier:
Hi,

This PR aims to rewrite t1-freesurfer-longitudinal pipeline while simplying its use at the same time. It also fix the case where`T1FreeSurferLongitudinalCorrection` needs to be connected to `T1FreeSurferTemplate` prior to running. Now, these two pipelines are independant.

On top of that:
- When running `t1-freesurfer` pipeline, the `regional_measures` folder may not be copied after the execution of the pipeline
- utils/longitudinal.py now centralize tools for longitudinal pipelines.
- Currently, Clinica Pipeline class and clinica/utils folder can not handle the case where we need to manipulate longitudinal IDs. When a new longitudinal pipeline will be developed into Clinica, refactoring will be needed.

Current status:
- `t1-freesurfer-template`
- [X] Rewrite pipeline
- [X] Handle overwrite option
- [X] Deal with case where SUBJECTS_DIR path is too long when using 1 time point

- `t1-freesurfer-longitudinal-correction`
- [X] Rewrite pipeline
- [x] Handle overwrite option

- Continuous integration:
- [x] CI Data
- [x] Update code


alexandreroutier:
Final testing is being performed but this PR can now be reviewed. Output console will look like:

t1-freesurfer

clinica/pipelines/t1_freesurfer_longitudinal/t1_freesurfer_template_pipeline.py
```diff
@@ -1,34 +1,38 @@
...
+        if os.path.isdir(caps_directory):
+            t1_freesurfer_files = clinica_file_reader(
+                list_participant_id, list_long_id,
+                caps_directory, T1_FS_T_DESTRIEUX, False
+            )
+            image_ids = [re.search(r'(sub-[a-zA-Z0-9]+)_(long-[a-zA-Z0-9]+)', file).group()
+                         for file in t1_freesurfer_files]
+        return image_ids
+
     def check_pipeline_parameters(self):
```
mdiazmel: Why we keep this (unnecessary) function in most of the pipelines? Is it supposed to be completed later?

clinica/pipelines/t1_freesurfer_longitudinal/t1_freesurfer_template_utils.py
```diff
@@ -1,1264 +1,214 @@
...
 
-def get_capsxsectional_path(caps_dir, subject, session):
-    """Get path to cross-sectional run in CAPS dir
+    This function will create folders and symbolic links in SUBJECTS_DIR env variable for upcoming run of recon-all.
 
-    Args:
-        caps_dir (string): location of CAPS folder
-        subject (string): participant ID
-        session_list (strings): session
+    Note(AG):
```
mdiazmel: Can you confirm that this info is also in the wiki of the pipeline?

alexandreroutier: Not yet but I plan to add a warning/info the wiki page explaining the creation of folder in $TMP written in the CLI ([cf screenshot](https://user-images.githubusercontent.com/24808663/74757893-e633e280-5276-11ea-82b1-ce6a50287e27.png))

clinica/pipelines/t1_freesurfer_longitudinal/t1_freesurfer_template_pipeline.py
alexandreroutier: This function is automatically called when the pipeline is built and, as a result, run.

It was designed for the checking of pipeline parameters (see e.g. [StatisticsSurface](https://github.com/aramis-lab/clinica/blob/dev/clinica/pipelines/statistics_surface/statistics_surface_pipeline.py)). This can be also used for setting default values of a pipeline when instanciating a new pipeline outside the CLI (e.g. with the continuous integration). See e.g. [T1VolumeTissueSegmentation](https://github.com/aramis-lab/clinica/blob/dev/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py). 

It was implemented in several pipeline but I don't remember if this is done for **all** pipelines.

Of course, if a pipeline has no parameter (e.g. DwiDti), this is useless :)

clinica/pipelines/t1_freesurfer_longitudinal/t1_freesurfer_template_utils.py
alexandreroutier: For this situation, the command line looks like:
T1FreeSurferTemplate

This is now explained on the Wiki:
Wiki

:)

alexandreroutier:
> It seems that this PR is almost done (excepting the CI/tests). I don't have enough experience with Freesurfer but I think that the pipeline is really doing the things in a clever way in order to manage longitudinal sessions.
> I found particularly nice the use of symlinks to simulate the folder structures.
> I added just a couple of comments to be addressed in the code.
> Nice job!

Thanks for your feedback! Credits for the use of symlinks are for @alexis-g-icm ;-)

Reference data were generated for `t1-freesurfer-template` and it should be generated today for  `t1-freesurfer-longitudinal-correction`.

Once the CI will pass on my computer, I will upload the data on the CI machines and this PR should be ready to merge.

alexandreroutier:
Data are correctly copied to CI machines and functional tests worked on linux workstation.

CI_FreeSurfer

PR #59, opened 2020-01-24
New version of DWI preprocessing pipelines

alexandreroutier:
This PR contains new versions of DWI-Preprocessing pipelines. They all rely on `eddy` tool [[Andersson et al., 2016](https://dx.doi.org/10.1016%2Fj.neuroimage.2015.10.019)] from FSL

- A very close version of `dwi-preprocessing-using-t1w` was evaluated in [[Wen et al., Submitted](https://arxiv.org/abs/1812.11183)]
- `dwi-preprocessing-using-phasediff-fmap` will be evaluted before merging


alexandreroutier:
Hi @arnaudmarcoux,

This PR is ready for review. I let this PR as a Draft PR since `dwi-preprocessing-using-phasediff-fmap` will be evaluated by Dario so small changes may appear for this pipeline.

Alexandre

PR #82, opened 2020-04-06
Update fetch_file function

mdiazmel:
Update the `fetch_file` function, verify downloaded file with SHA256.


mdiazmel:
Hi @alexandreroutier, can you do a quick revision of this PR ! 
Thanks !

mdiazmel:
@alexandreroutier ,

> Can you confirm that the line cert = ssl.get_server_certificate(("aramislab.paris.inria.fr", 443)) is not needed anymore?

Yes, not needed anymore, however I have to find a better way to handle the access to an https server. For the moment, current implementation works.

> Is hashlib a standard package (to make sure we do not need to add it to requirements)?

Yes, it's in the Python standard library. 

> Of note, we should at least put CAT12 atlases on the server. As a result, this dependency would not be needed anymore 🙃 Do we plan this for the 0.4 release?

Good idea. I will open an issue with this, and we could add it to the roadmap for 0.4.

I will fix the typo and merge.

PR #83, opened 2020-04-06
point to clinica logo in aramis-lab repo

katjaq:
Now that the image is in your repo, you can show it from there :)
And I had to fix the hidden link for display on BrainWeb.
Thank you again for being part of BrainWeb. :)


mdiazmel:
Ok !

PR #94, opened 2020-05-14
CI modifications

mdiazmel:
This PR includes two main changes:

* Automatically generation of testing reports in order to be displayed in the CI interface
* Jenkinsfile modification. It allows to recreate the Python environment  if the requirement file  changes.


mdiazmel:
Hi,

> One final question before merging : I think `junit-xml` needs to be added to `requirements-dev.txt`, right?

No, `junit-xml` reporting feature comes out-of-the-box with Pytest. It uses the current requirements file.

PR #99, opened 2020-06-06
Update T1 linear default behavior

mdiazmel:
From now, the pipeline crops the input image using the template. 

If the option `--uncropped_image` is added to the command line, the image is not cropped.

### DIFF
--- a/clinica/pipelines/t1_linear/t1_linear_cli.py
+++ b/clinica/pipelines/t1_linear/t1_linear_cli.py
@@ -33,8 +33,8 @@ class T1LinearCLI(ce.CmdParser):
         optional = self._args.add_argument_group(PIPELINE_CATEGORIES['OPTIONAL'])
 
         optional.add_argument("-cp", "--uncropped_image",
-                              help='''Do not crop the image with template 
-                              (cropped image are suggested for using with DL 
+                              help='''Do not crop the image with template
+                              (cropped image are suggested for using with DL
                               models)''',
                               action='store_true',
                               default=False)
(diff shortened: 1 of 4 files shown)


mdiazmel:
Hi @alexandreroutier ,
This PR changes the default options for `t1_linear` pipeline. Can you take a look, please?
We should edit the the pipeline documentation/wiki.

clinica/pipelines/t1_linear/t1_linear_cli.py
```diff
@@ -32,8 +32,10 @@ def define_options(self):
         # Clinica optional arguments
         optional = self._args.add_argument_group(PIPELINE_CATEGORIES['OPTIONAL'])
 
-        optional.add_argument("-cp", "--crop_image",
-                              help='Crop the image using a template (suggested for using with DL models)',
+        optional.add_argument("-cp", "--uncropped_image",
```
alexandreroutier, line 35:
```suggestion
        optional.add_argument("-ui", "--uncropped_image",
```

alexandreroutier:
Hi @mdiazmel,

I noticed a flag to update. BTW, should a:
```python
parameters = {
    'uncropped_image': False
}
```
be present in instanciation tests? 

I just finished to update the Wiki page of `t1-linear` pipeline. Once the modification done, you can safely merge :)

Alex

mdiazmel:
Hi @alexandreroutier ,

> I noticed a flag to update. BTW, should a:
> 
> ```python
> parameters = {
>     'uncropped_image': False
> }
> ```
> 
> be present in instanciation tests?

If we want to be coherent with the non-regression tests, the answer is yes. In practice, it's not necessary because the default value for `uncropped_image` is `False`.

I'll add it for the sake of clearness on the instantiation tests.

PR #127, opened 2020-09-18
Simplify adni converter

### LINKED ISSUE #126: Diagnosis conversion for sessions of ADNI3
Hello,

It seems that there is an issue in the conversion of the diagnoses of ADNI3 sessions as the columns `adni_diagnosis_change` and `diagnosis` are empty in the tsv files at the session level.
This could be due to the fact that the columns `DXCHANGE` and `DXCURREN` in `DXSUM_PDXCONV_ADNIALL.csv` are all empty for ADNI3, and that the new column `DIAGNOSIS` should be used instead.
I will ask the question at https://groups.google.com/g/adni-data to have the confirmation and then we should change the converter.

14thibea:
This PR closes #126 and simplify the output files by removing the field `adni_diagnosis_change` as all the relevant information is now in `diagnosis`.


alexandreroutier:
Hi @14thibea,

This PR looks good to me.

Can we merge this PR?

14thibea:
Yes!
If you agree removing the column `adni_diagnosis_change` which is redundant with `diagnosis` now.

alexandreroutier (commented):
The simpler the better

PR #139, opened 2020-10-30
Solve minor issues on ADNI to BIDS conversion

14thibea:
Closes #101 and closes #71 .

PR #140, opened 2020-11-02
Diagnosis change

### LINKED ISSUE #131: Changes in diagnosis in recent versions of ADNI
Hello,

We found out that some diagnoses of ADNI participants changed in the recent version of raw data of ADNI. This could lead to  reproducibility issues.
We are currently discussing with the ADNI data user group to better understand these changes.

14thibea:
Closes #131 .

Actually the field `DX_bl` does not give the diagnosis at baseline but the diagnosis at screening (why the participant was recruited). It may happen that at baseline (ses-M00) the diagnosis is not the same as there are more tests that are performed. Then the diagnosis at baseline is more reliable than `DX_bl`.

I suggest in this commit to rename the field associated to `DX_bl` `diagnosis_sc` (instead of the previous confusing name `diagnosis_bl`). Then we should warn the users as the fields are changing.


alexandreroutier:
How did you modify the TSV file?

I don't why but it has been "one-lined" (see https://github.com/aramis-lab/clinica/pull/140/files).

14thibea:
The only change is the last line that was added:
` Diagnosis at screening	diagnosis_sc	AD=Alzheimer's disease - bvFTD=behavioral variant Fronto Temporal Dementia	string						DX_bl	ADNIMERGE.csv										`

alexandreroutier:
Since the merge of PR #139, this PR cannot be merged due to conflicts to the following files:
- `clinica/iotools/converters/clinical_data_bids_correspondence/clinical_data_bids_correspondence_participant.tsv`
- `clinica/iotools/data/clinical_specifications_adni.xlsx`

I don't know for you what will be the faster between rebasing on `upstream/dev` or making a new PR 🤔

alexandreroutier (approved):

14thibea:
I now really understand why we should only use TSV files and not Excel...
I will try to rebase, if the result is awful I will make a new PR.

PR #142, opened 2020-11-03
Change diagnosis at baseline in diagnosis at screening

### LINKED ISSUE #131: Changes in diagnosis in recent versions of ADNI

14thibea:
Closes #131  .

Actually the field `DX_bl` does not give the diagnosis at baseline but the diagnosis at screening (why the participant was recruited). It may happen that at baseline (ses-M00) the diagnosis is not the same as there are more tests that are performed. Then the diagnosis at baseline is more reliable than `DX_bl`.

I suggest in this commit to rename the field associated to `DX_bl` `diagnosis_sc` (instead of the previous confusing name `diagnosis_bl`). Then we should warn the users as the fields are changing.


alexandreroutier (approved):
All good!

PR #160, opened 2021-01-25
Add known failures + version folder in conversion_info

14thibea:
Add new failures after a conversion of a new batch of images from ADNI3.
A version folder is added in `conversion_info` to avoid to erase previous `*_paths.tsv`.


mdiazmel (commented):
Hi @14thibea ,

Thanks for this PR.  I ran the nonregression test locally and everything seems ok.

I just have one question that came up during the review: whats the behavior of the converters when the raw databases are pointed by symlinks (and also hardlinks)? I mean, what's the path written in the tsv files inside the `conversion_info` folder (symlink or real path)?

14thibea:
Hi @mdiazmel 

I just tested and the paths written use the symlink and not the real path. Do you think it should be changed?

mdiazmel:
> I just tested and the paths written use the symlink and not the real path. Do you think it should be changed?

Non, it's ok. Maybe add a comment about it, somewhere in the documentation (for example at the "Step 2: Path extraction" section of the appendix in `docs/Converters/ADNI2BIDS.md`).

14thibea:
It's done ! :)

mdiazmel (approved):
I'll merge this PR!

PR #161, opened 2021-02-02
Update t1-linear doc

mdiazmel:
Add tip in documentation for T1-Linear to emphasize that `clinica run t1-linear --help` will display all the possible options for this pipeline.


14thibea (approved):
Sounds good to me!
Now we should add the same things on `t1-linear` documentation of ClinicaDL.

mdiazmel:
Update t1-linear documentation command to get all the possible options.

PR #162, opened 2021-02-03
Add longitudinal analysis

14thibea:
This PR intends to add a longitudinal analysis in `check-missing-modality`.
Maybe it should be done in another iotool, but the reason it was done in `check-missing-modality` is that it uses the outputs of this function.


mdiazmel (approved):
I'll merge this one!

PR #165, opened 2021-02-22
Simplify Volume pipelines (Container, Nipype interfaces)

alexandreroutier:
Hi @mdiazmel,

You can find below the content of the PR:
- Remove redundant functions to get container from filename (`{pet-|t1}-volume`, `dwi-dti`)
- Improve some docstrings and f-strings formatting
- Migrate DWI-DTI pipeline to Black
- Simplify DWI-DTI pipeline: the pipeline now uses Nipype-ANTs interface instead of `os.command` calls. As a result, content of `dwi_dti_workflows.py ` was merged into `dwi_dti_pipeline.py` leading to a shorter file with same functionalities.

Best,
Alex

PS : I will make another PR for Black configuration as discussed and another one for DWI-Preprocessing.


clinica/pipelines/dwi_dti/dwi_dti_cli.py
```diff
@@ -4,46 +4,53 @@
...
-        self._name = 'dwi-dti'
+        self._name = "dwi-dti"
 
     def define_description(self):
         """Define a description of this pipeline."""
-        self._description = ('DTI-based processing of DWI datasets:\n'
-                             'http://clinica.run/doc/DWI_DTI')
+        self._description = (
+            "DTI-based processing of DWI datasets:\n"
+            "http://www.clinica.run/doc/Pipelines/DWI_DTI"
```
mdiazmel, line 15: Now that the documentation is stored in an alternative server, the url will be: https://aramislab.paris.inria.fr/clinica/docs/public/latest/Pipelines/DWI_DTI/
This change will be done also in another pipelines. Maybe we have to think about redirecting all the links automatically...

mdiazmel (commented):
Hi Alex,
Seem ok, just we have to figure out how we manage the urls pointing to the new documentation.

clinica/pipelines/dwi_dti/dwi_dti_cli.py
alexandreroutier: I have wrote an Issue about it (#170) which is independent of this PR (the URL was wrong since the beginning 😅 ).

mdiazmel:
I'll merge this PR.

Thanks @alexandreroutier !

PR #171, opened 2021-02-26
Roi extraction

14thibea:
This PR intends to add ROI extraction to a `deeplearning-prepare-data`.
The corresponding file must be transformed to use nipype.


14thibea:
Replaced by #214

PR #175, opened 2021-03-05
Fix minor bug with pvc

14thibea:


14thibea:
the doc must be updated to be more clear on psf tsv file.

14thibea:
Hi @mdiazmel ,
This PR is ready to be merged!
It solves a bug in `pet-volume` when performing pvc.

alexandreroutier:
Hi @14thibea,

I understand your change in 37b541f because of ADNI. Instead of 'lowercasing' tracer without explanation, maybe we can add a BIDS Layout e.g.:

```text
bids
bids
└── sub-001
    └── ses-M00
            └── sub-001_ses-M00_task-rest_acq-fdg_pet.nii.gz
            └── sub-001_ses-M00_task-rest_acq-av45_pet.nii.gz
```
and say that the corresponding TSV file for PSF information should match the value for `acq` key?

Poking @nburgos for extra feedback.

14thibea:
Hi @alexandreroutier 

I tried to add your suggestion, is it better like this?

alexandreroutier:
Looks good to me :) I added an issue #176 so that this error would not occur anymore in the future.

I'll let @mdiazmel do the merge.

Alex

PR #177, opened 2021-03-11
Change suffix

14thibea:
Hi @mdiazmel ,

The user I sent data to told me that the suffix by used was a bit misleading. This PR intends to change it.


mdiazmel:
Ok, I'll merge this one!

PR #190, opened 2021-04-07
Bugfix when using a nipype node with the container_from_file() function

mdiazmel:
Bugfix detected in non-regression test for `PETVolume` and  `T1VolumeTissueSegmentation`: wrong input name for nipype nodes when using the `container_from_file()` function.


omar-rifai:
Looks good, thanks for the changes @mdiazmel.

PR #193, opened 2021-04-12
Handle case of CAPS without groups folder

14thibea:


clinica/iotools/utils/data_handling.py
```diff
@@ -201,9 +201,10 @@ def create_merge_file(
         if pipelines is None:
             for pipeline_name, pipeline_fn in pipeline_options.items():
                 merged_df, summary_df = pipeline_fn(caps_dir, merged_df, **kwargs)
-                merged_summary_df = pd.concat([merged_summary_df, summary_df])
+                if summary_df is not None:
+                    merged_summary_df = pd.concat([merged_summary_df, summary_df])
 
-                if len(summary_df) == 0:
+                if len(summary_df) == 0 or summary_df is not None:
```
omar-rifai, line 207: Hello @14thibea, this is problematic when summary_df is Nonetype(). Did you mean:

summary_df _is_ None or len(summary_df) == 0?

omar-rifai: Also, the dataframe indexing at line 223 seems to be problematic when the summary_df is None

14thibea: I proposed commits for both comments, tell me if it is OK !

PR #197, opened 2021-04-13
Fix CLI for merge-tsv iotool

mdiazmel:
Fix CLI for merge-tsv tool

### DIFF
--- a/clinica/iotools/utils/data_handling_cli.py
+++ b/clinica/iotools/utils/data_handling_cli.py
@@ -142,7 +142,7 @@ class CmdParserMergeTsv(ce.CmdParser):
             "--ignore_session_scan_files",
             default=False,
             action="store_true",
-            help="If given, the session ans scan files will not be read and added to "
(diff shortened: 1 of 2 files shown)


clinica/iotools/utils/data_handling_cli.py
```diff
@@ -135,7 +135,14 @@ def define_options(self):
...
             action="store_true",
-            help="If given, the session and scan files will not be read and added to "
+            help="If given, the scan files will not be read and added to "
+            "the final output. This may accelerate the procedure.",
+        )
+        iotools_options.add_argument(
+            "--ignore_session_scan_files",
+            default=False,
+            action="store_true",
+            help="If given, the session ans scan files will not be read and added to "
```
omar-rifai, line 145: There is a small typo in the help message, I think it should read: 
"If given, the session files will not be read and added to"

mdiazmel: Done in 74d55a9

Issue #167, opened 2021-02-25
Remove FSL dependency from OASIS2BIDS converter


mdiazmel:
**Describe the bug**
FSL dependency was introduced in commit 8dda1394b
See `git blame clinica/iotools/converters/oasis_to_bids/oasis_to_bids.py`

**To Reproduce**
Run non-regression test for OASIS converter in current dev branch.
BIDS validator detects an error in the header of nifty files form OASIS3

**Additional context**
See here https://neurostars.org/t/error-when-converting-t1w-to-bids-t1w-file-with-too-many-dimensions/2568/6
One possible solution could be to use nibabel to rewrite the header of the files.

PR #144, opened 2020-11-06
Bids validator

14thibea:
Fix or ignore errors found by BIDS validator.
All pipelines were fixed except NIFD.

PR #201, opened 2021-04-14
Perform OASIS conversion even if FSL is not installed

14thibea:


omar-rifai (approved):
Thanks @14thibea, I will merge these changes.

alexandreroutier:
Hi @14thibea,

I find it weird than the resulting conversion will depend on if FSL is present or not.

If you knwo what if the issue on header, would it safer to use <https://nipy.org/nibabel/nifti_images.html#the-nifti-header> to fix the dimension information instead? It would would FSL from pre-requisites while having a reproducible converter at the same time.

Best,
Alex

omar-rifai:
Hi @alexandreroutier, 

Is it really just an information problem? Doesn't fslroi actually convert the image? In which case I think only modifying the header information could lead to some issues as it would no longer match with the actual data of the image? 

Omar

mdiazmel:
I'm agree with @alexandreroutier .  

The right approach to solve this problem is to bypass FSL in the OASIS2BIDS convertor. Nibabel could be an alternative. See issue #167 for more references.

omar-rifai:
All right @14thibea, in case we need to actually transform the data and not just the header:

I found a [nibabel function](https://nipy.org/nibabel/reference/nibabel.funcs.html#nibabel.funcs.four_to_three) that transforms a 4D image into a list of 3D images. I think you can use that and take the first (?) element of the list so that we are left with an image with dim0 = 3 as required.

If it's only the header that is corrupted, I guess we can change the header manually.

Omar

alexandreroutier:
I think the first attempt of @14thibea in https://github.com/aramis-lab/clinica/pull/144/commits/788d830ac994c36133a6b4595cc0b9c9f4e06270 from PR #144 was close to the solution.

From what I understood in the forum conversation in Issue #167, OASIS headers has the following info in their header:
```
>>> n2_header = n2_img.header
>>> print(n2_header)                     
    <class 'nibabel.nifti2.Nifti2Header'> object, endian='<'
[...]
    dim             : [ 4 <dim1> <dim2> <dim3> <dim4> <dim5> <dim6> <dim7> ]
[...]
```
I would try to use `get_dim` to check this assumption and if that's the case, I would use replace `4` par `3` and update the header with `set_dim`. Of course, we need to make sure that `fsl_roi` did not modify anything else in the header of the image.

Alex

14thibea:
Following the suggestion of @omar-rifai I replaced FSL by `nibabel.funcs.four_to_three`.
I tested on the first participant of OASIS and ran bids-validator and it seems OK.

14thibea:
The tests crashed because there is no space left on the CI machine, but I would also like to test on the whole dataset before merging, I'll ping you when it's ok!

14thibea:
Ok for me on the whole dataset @omar-rifai

omar-rifai:
Great thanks, seems fine with me. If OK with @mdiazmel  and @alexandreroutier, I'll merge as is ? I think this approach has the advantage of seeming less "hacky" then just modifying the header while being equivalent to what is currently done using FSL.

alexandreroutier:
Hi @omar-rifai,

Since it worked on the whole dataset; it is good for me!

Alex

PR #210, opened 2021-04-21
Add option to skip Clinica question

### LINKED ISSUE #207: Cannot run t1-freesurfer on cluster when nifti file is not centered
**Describe the bug**
Some nifti files may not be centered, and SPM may fail on these files. This is not the case for freesurfer.
However, clinica asks for an answer (yes or no) to resume or not running the pipeline.
When running the pipeline on a cluster we cannot give an answer and the pipeline crashes.

**To Reproduce**
run t1-freesurfer on sub-AIBL1499_ses-M00_T1w.nii.gz on a cluster.

**Expected behavior**
An option should be provided to force running and avoid asking for an input.

**Screenshots**
Capture d’écran 2021-04-19 à 10 28 30

14thibea:
Closes #207 .
Could not test yet as the cluster is down.

### DIFF
--- a/clinica/engine/cmdparser.py
+++ b/clinica/engine/cmdparser.py
@@ -100,6 +100,7 @@ class CmdParser:
         add_wd_flag=True,
         add_nprocs_flag=True,
         add_overwrite_flag=False,
+        add_yes_flag=False,
     ):
         clinica_standard_options = self._args.add_argument_group(
             PIPELINE_CATEGORIES["CLINICA_OPTIONAL"]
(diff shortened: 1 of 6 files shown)


14thibea:
It's working for me!

alexandreroutier:
Hi @14thibea,

Why `skip_question` parameter should be only in `t1-freesurfer` pipeline?

It may be more convenient to extend in `cmdparser.py` this function:
```python
    def add_clinica_standard_arguments(
        self,
        add_tsv_flag=True,
        add_wd_flag=True,
        add_nprocs_flag=True,
        add_overwrite_flag=False,
    ):
```
🤔 

Best,
Alex

PR #213, opened 2021-04-28
Add new metadata to AIBL converter

14thibea:
JSON files associated to Nifti files will be created to give more information on the images.


mdiazmel (approved):
HI @14thibea ,

Looks good to me!

Don't forget to update reference data for the tests! (see with @omar-rifai or with me).

PR #224, opened 2021-05-06
Restrict nipype updates to the current major version

ghisvail:
I know the nipype team is actively working on the next major version of nipype. We might want to avoid accidental updates to it considering it's a critical dependency.


mdiazmel (approved):
Nice! Thanks @ghisvail !

LGTM, I'll merge!

PR #227, opened 2021-05-07
Update NR test for deeplearning-prepare-data with pet-linear modality

omar-rifai:
Add the pet-linear modality to the non regression test of deeplearning-prepare-data pipeline


mdiazmel (approved):
LGTM, I'll merge this one!

PR #234, opened 2021-05-10
docs: Install dcm2niix with Homebrew

ghisvail:


ghisvail:
I have updated the instructions but running `dcm2nii` on my machine gives me the following error `bad CPU type in executable: dcm2nii`.

Not unexpected for such an old release though.

ghisvail:
I wonder whether it would not be best to translate all our calls to `dcm2nii` to their equivalent in `dcm2niix` anyway. I believe it really is beyond deprecated at this point.

omar-rifai:
Yes, definitely better. If you want to have a look I think it'd be a welcome update.

ghisvail:
I'll advise to merge the updated docs as is.

I'll deal with the migration to `dcm2niix` in a separate PR which addresses #235.

PR #236, opened 2021-05-11
docs: Update instructions for MRtrix3

ghisvail:
Installing MRtrix3 with Homebrew is quite a pain on Big Sur.

It requires XCode and Qt5, which are quite significant in size, and users may get this unwelcoming error message if XCode is not setup properly:

`xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance`

I have updated our docs to point towards the official channels recommended upstream.


mdiazmel (approved):
Thanks @ghisvail!

How about the dependencies of the Conda package of MRtrix3? Are they lighter than the dependencies proposed by Homebrew? 

Anyway, it's a good idea to propose to the users multiple sources for installing the "third-party" software, besides the official documentation!

PR #245, opened 2021-05-19
Install dcm2niix through conda-forge

ghisvail:
Probably easier to deploy `dcm2niix` alongside `clinica` using conda instead of following additional instructions to build your own version or fetch a binary + exposing it somewhere in `$PATH`.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,8 +1,10 @@
 name: clinica_env
 channels:
   - defaults
+  - conda-forge
 dependencies:
   - python=3.7
   - pip
   - pip:
       - -r requirements.txt
+  - dcm2niix


mdiazmel (commented):
Hi @ghisvail ,

The file `environment.yml` is (currently) only used to create a Python environment into the CI machines (where `dcm2niix` is installed in the classic way). We don't advice users to use a specific Conda environment (probably someone want to use `virtualenvs`), so this file is transparent for users. If we want to provide a `yml` file for the users, we should try to have most of the packages installed by Conda (and not by `pip`). We can consider to do this choice once we have verified that the dependency tree for Clinica is well solved by Conda (it wasn't the case some time ago). 

Could you please add a word to the documentation, to propose the users to install `dcm2niix` via Conda, if they wish it ? (This file: `clinica/docs/Third-party.md`).

ghisvail:
> only used to create a Python environment into the CI machines

Oh, I see. I have used it to setup my development environment by mistake. It's worked fine so far.

> We don't advice users to use a specific Conda environment

I believe it would be a nicer experience if developers could just run `conda env create -f environment.yaml` and have most third-party utilities available this way, instead of following specific instructions for each. And it could be used for CI and Docker packaging purposes as additional benefits. Too much of a stretch, perhaps.

Feel free to close it.

mdiazmel:
> I believe it would be a nicer experience if developers could just run conda env create -f environment.yaml and have most third-party utilities available this way, instead of following specific instructions for each. And it could be used for CI and Docker packaging purposes as additional benefits. Too much of a stretch, perhaps.

Yes, I'm agree. Indeed that was what we used to do some time ago (at least for the "pure Python" dependencies) but frequently we encountered inter-dependency problems when some packages where updated. I known that Conda has recently changed the way how to solve the dependencies tree, maybe it's worth to try again... and of course, having the third-party installed as Conda packages is a big plus!  The downside is that we have to maintain three files `environment.yml`, `requirements.txt` and `setup.py`.

For this PR, do you want to modify the documentation to propose to the users to install dcmniix as a conda package?

ghisvail:
> For this PR, do you want to modify the documentation to propose to the users to install dcmniix as a conda package?

I don't think that's useful if `dcm2nixx` is installed standalone in a conda env separate from Clinica. Since conda deployments are not something we are ready to support yet, maybe we should close this and revisit later once I get more tools accepted to `conda-forge`.

PR #246, opened 2021-05-20
Remove reference to --force_new_extraction in AIBL converter CLI

ghisvail:
There is no such option in the AIBL converter CLI. I guess it came from a copy-pasta of the ADNI one, where this option is available.


omar-rifai:
Good catch. Thanks

PR #251, opened 2021-05-21
Major refactoring of adni2bids clincal data extraction

omar-rifai:
-Change xlsx reference tables by tsv tables 
-Clean the create_sessions_dict function to use built-in pandas function


ghisvail (approved):

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -564,237 +510,194 @@ def remove_fields_duplicated(bids_fields):
...
+        examdate_bl = datetime.strptime(row["EXAMDATE_bl"], "%Y-%m-%d")
+        delta = examdate - examdate_bl
+
+        updated_age = round(
+            float(row["AGE"]) + (delta.days / 365.25),
+            1,
+        )
+    else:
+        updated_age = row["AGE"]
+    return updated_age
```
ghisvail, line 547: If row corresponds to a row in some upstream `DataFrame`, perhaps this could be done more efficiently with Pandas directly?

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -564,237 +510,194 @@ def remove_fields_duplicated(bids_fields):
...
+
+
+def pad_id(ref_id):
+    """Add leading zeros to the RID to keep à length of 4"""
+    rid = str(ref_id)
+    # Fill the rid with the needed number of zero
+    if 4 - len(rid) > 0:
+        zeros_to_add = 4 - len(rid)
+        rid = "0" * zeros_to_add + rid
+    return rid
```
ghisvail, line 557: Same remark: if this is part of a higher processing loop on a Pandas dataframe, using Pandas string processing primitives would likely be more efficient.

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -893,7 +796,7 @@ def create_adni_scans_files(conversion_path, bids_subjs_paths):
                             f"{Fore.RED}No information found for file {file_name}.{Fore.RESET}"
                         )
 
-                    scans_df = scans_df.fillna("n/a")
+                    scans_df = scans_df.fillna("")
```
ghisvail, line 799: Why the empty string instead of "n/a" previously? May be worth adding a comment.

clinica/iotools/utils/data_handling.py
```diff
@@ -142,6 +142,9 @@ def create_merge_file(
                                 for key, value in json_dict.items():
                                     new_col_name = f"{modality}_{key}"
                                     scans_dict.update({new_col_name: value})
+                    scans_dict = {
+                        str(key): str(value) for key, value in scans_dict.items()
+                    }
```
ghisvail, line 147: Why the explicit conversion?  Instead, I think `pd.DataFrame(..., dtype="string")` should work.

ghisvail (commented):
LGTM with a few comments

omar-rifai:
Thanks for the review!

 The "n/a" is indeed an issue, I will fix this. 
 
 For the code optimization part, it's a fair point and I'll try to look for alternatives if time allows but right now the performance is already pretty decent (<1hr) on the full ADNI database compared to several hours with the past version.

clinica/iotools/utils/data_handling.py
omar-rifai: This is kind of a 'hot-fix' because the merge was breaking from the modifications in [PR 213](https://github.com/aramis-lab/clinica/pull/213). 

This part of the code needs another major refactoring.

PR #258, opened 2021-05-28
Add missing else branch to scan file writer

ghisvail:
Missing else branch may lead to `f_type`key being undefined.

Add a `continue` statement to keep iterating nonetheless.


omar-rifai:
Good find. Thanks!

PR #269, opened 2021-06-07
[ADNI/AIBL]  image centering as optional arguments

omar-rifai:
Put the origin of the coordinate system at the center of the image using an optional command line argument for ADNI and AIBL instead of default.


omar-rifai:
After further discussion, we will move the image centering option to the different pipelines instead of the converter

PR #286, opened 2021-06-14
Add compat module for errno

ghisvail:
Required to maintain compatibility with Python versions older than 3.7.

Having compatibility code in `<package>/compat.py` is pretty standard.

### DIFF
--- a/clinica/compat.py
+++ b/clinica/compat.py
@@ -1,0 +1,4 @@
+try:
+    import errno
+except ImportError:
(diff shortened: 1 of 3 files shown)


ghisvail:
The other option is to raise our Python version compatibility to 3.7+ and just use `import errno` everywhere.

mdiazmel (approved):
LGTM,
I'll merge this PR!

PR #288, opened 2021-06-14
Remove unused smooth option for t1-v-r-d pipeline

ghisvail:
This option is not used in the run command of the pipeline. Probably left over from copy-pasting.


mdiazmel:
I close this PR because it has became deprecated after the CLI refactor (PR#275).

PR #289, opened 2021-06-16
Add oasis3 features

ecamenen:
Add converter from OASIS3 data https://www.oasis-brains.org/#data into bids format


clinica/iotools/bids_utils.py
```diff
@@ -114,6 +115,8 @@ def create_participants_df(
     for i in range(0, len(participant_df)):
         if study_name == "OASIS":
             value = (participant_df["alternative_id_1"][i].split("_"))[1]
+        elif study_name == "OASIS3":
+            value = re.sub("OAS3", "", participant_df["alternative_id_1"][i])
```
ghisvail, line 119: You should be able to use the `replace`method from string instead of a regexp substitution.

clinica/iotools/bids_utils.py
```diff
@@ -222,23 +229,23 @@ def create_sessions_dict(
...
+
+                    subj_dir = path.join(
+                        path.dirname(path.dirname(clinical_data_dir)),
+                        "out",
+                        "bids",
+                        subj_bids
+                    )
+                    session_names = get_bids_subjs_list(subj_dir)
+                    for s in session_names:
+                        s_name = re.sub("ses-", '', s)
```
ghisvail, line 241: Same comment, use the `replace` method.

clinica/iotools/bids_utils.py
```diff
@@ -450,12 +457,13 @@ def write_sessions_tsv(bids_dir, sessions_dict):
...
-                    "i",
-                ],
-            )
+            columns = list(list(sessions_dict.values())[0].values())[0].keys()
+            session_df = pd.DataFrame(columns=columns)
+            it = 0
+            for k in sessions_dict[bids_id].keys():
+                session_dict = sessions_dict[bids_id][k]
+                session_df_temp = pd.DataFrame(session_dict, index=[k, ])
+                session_df = pd.concat([session_df, session_df_temp])
```
ghisvail, line 466: Why do you need to build a DataFrame by successive concatenation? Can't you do it in one go, which is faster and more idiomatic?

clinica/iotools/bids_utils.py
```diff
@@ -485,25 +493,23 @@ def write_scans_tsv(bids_dir, bids_ids, scans_dict):
...
 
     for bids_id in bids_ids:
-        # Create the file
         sessions_paths = glob(path.join(bids_dir, bids_id, "ses-*"))
         for session_path in sessions_paths:
-            scans_df = pd.DataFrame()
             session_name = session_path.split(os.sep)[-1]
-            tsv_name = bids_id + "_" + session_name + "_scans.tsv"
-
+            scans_df = pd.DataFrame()
```
ghisvail, line 502: More of a nitpick, but please avoid moving code around without touching it 🙏 

The lighter the diff of the PR, the easier it gets for people to review your work

clinica/iotools/bids_utils.py
```diff
@@ -485,25 +493,23 @@ def write_scans_tsv(bids_dir, bids_ids, scans_dict):
...
+            # Create the file
+            tsv_name = "_".join([bids_id, session_name, "scans.tsv"])
```
ghisvail, line 504: Similar comment. You just rewrote the same line in a more idiomatic fashion, but it adds nothing of value to this PR.

ghisvail: Not like it's a bad thing, but again the lighter the diff the better.

clinica/iotools/bids_utils.py
```diff
@@ -485,25 +493,23 @@ def write_scans_tsv(bids_dir, bids_ids, scans_dict):
...
             # If the file already exists, remove it
             if os.path.exists(path.join(bids_dir, bids_id, session_name, tsv_name)):
                 os.remove(path.join(bids_dir, bids_id, session_name, tsv_name))
 
             mod_available = glob(path.join(bids_dir, bids_id, session_name, "*"))
             for mod in mod_available:
                 mod_name = os.path.basename(mod)
-                files = glob(path.join(mod, "*"))
+                files = glob(path.join(mod, "*.nii.gz"))
```
ghisvail, line 512: Why is this required?

clinica/iotools/bids_utils.py
```diff
@@ -514,9 +520,12 @@ def write_scans_tsv(bids_dir, bids_ids, scans_dict):
                             for carac in file_name.split("_")
                             if "-" in carac
                         }
-                        f_type = description_dict["acq"].upper()
+                        f_type = description_dict["trc"].upper()
```
ghisvail, line 523: Why acq -> trc?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+        # --Create participants.tsv--
+        participants_df = bids.create_participants_df(
+            "OASIS3", clinic_specs_path, clinical_data_dir, bids_ids
+        )
+
+        # Replace the values of the diagnosis_bl column
+        participants_df["diagnosis_bl"].replace([0.0, np.nan], "CN", inplace=True)
+        participants_df["diagnosis_bl"].replace(
+            [0.5, 1.0, 1.5, 2.0], "AD", inplace=True
+        )
```
ghisvail, line 41: Are we sure `diagnosis_bl` will always take this specific set of values? Is the general rule that a strictly positive value of `diagnosis_bl` corresponds to "AD" otherwise "CN"?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+        # Following line has no sense
+        # participants_df['diagnosis_bl'].replace(participants_df['diagnosis_bl']>0.0, 'AD', inplace=True)
+        participants_df = participants_df.fillna("n/a")
+        participants_df.to_csv(
+            path.join(bids_dir, "participants.tsv"),
+            sep="\t",
+            index=False,
+            encoding="utf-8",
+        )
```
ghisvail, line 50: Use the `na_rep` argument of `pandas.to_csv` instead.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+            import re
+            import shutil
+
+            def copy_file(source, target):
+                try:
+                    new_filename = re.sub("OAS3", "OASIS3", path.basename(source))
+                    new_filename = re.sub("_echo-", "_run-", new_filename)
+                    if re.search(".*_pet.(nii.gz|json)", new_filename) is not None:
+                        new_filename = re.sub("task-rest_", "", new_filename)
+                        new_filename = re.sub("acq", "trc", new_filename)
```
ghisvail, line 100: You don't need `re.sub` for all this. `str.replace` is enough.

ghisvail: Also, I am not quite sure what you are doing here. Pehaps a comment would help.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+            session_name = old_sess_fold[2]
+            session_folder = path.join(bids_subj_folder, "ses-" + session_name)
+            if not os.path.isdir(session_folder):
+                os.mkdir(path.join(session_folder))
+
+            # In order do convert the Analyze format to Nifti the path to the .img file is required
+            nifti_folders = glob(
+                path.join(subj_folder, "**/*.nii.gz"), recursive=True
+            )
+            #TODO: if not nifit ( len(img_file_path) = 0), then convert
```
ghisvail, line 132: Todo?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+
+            session_name = old_sess_fold[2]
+            session_folder = path.join(bids_subj_folder, "ses-" + session_name)
+            if not os.path.isdir(session_folder):
+                os.mkdir(path.join(session_folder))
+
+            # In order do convert the Analyze format to Nifti the path to the .img file is required
+            nifti_folders = glob(
+                path.join(subj_folder, "**/*.nii.gz"), recursive=True
+            )
```
ghisvail, line 131: I don't understand. Your comment says you are looking for folders containing an `.img`, yet you're doing a glob search on files ending with `.nii.gz`. This is confusing. Besides `nifti_folders` is not a good name since it will contain file paths, not folders.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
```diff
@@ -0,0 +1,165 @@
...
+
+                # Looking for an existing func or anat folder
+                dir_name = path.dirname(nifti_folder)
+                regex = re.compile("(?P<folder>anat|func)[1-9]")
+                if regex.search(dir_name) is not None:
+                    output_folder = regex.search(dir_name).group("folder")
+                elif regex.search(path.abspath(dir_name)) is not None:
+                    output_folder = regex.search(
+                        path.abspath(dir_name)
+                    ).group("folder")
```
ghisvail, line 150: Not sure I understand what's happening here. Can't you just filter with `if "anat" in path or "func" in path` according to your comment?

test/nonregression/test_run_iotools.py
```diff
@@ -0,0 +1,241 @@
+# coding: utf8
```
ghisvail, line 1: If I understand correctly, these set of tests did not exist before this PR, correct?

If so, it would be worth submitting a different PR for those not related to OASIS3. This PR should be specifically about adding OASIS3 conversion.

ghisvail: It's a great initiative nonetheless, and it deserves its own separate review process imo.

ghisvail (changes requested):
Thanks @ecamenen for your work on OASIS3. It looks good overall and there are some corners of the code which requires clarification or polish. I have left a few comments to guide you.

clinica/iotools/bids_utils.py
ecamenen: My bad. Automatic merge problem: I was coding this feature and it was coded elsewhere.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: Should I replace in the same time all the "fillna" in clinica?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: I based my work on the OASIS1 code: clinica/iotools/converters/oasis_to_bids/oasis_to_bids.py

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: Task metadata, notably parsed by some code GUI managers (e.g., IntelliJ). I forgot to remove it.

test/nonregression/test_run_iotools.py
ecamenen: Probably a quick merge problem. An old file that I reworked. It probably has been converted to Jenkins since then. I ll remove it.

clinica/iotools/bids_utils.py
ecamenen: Discussions with Elina (and Ninon, Mauricio & cie) where the "acq" suffix in the PET files should become "trc".

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: That does not anwer my question. I have a feeling that any value over 0 should be "AD", otherwise "CN" and the code could be simplified to reflect this rule.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: Older comments for OASIS which the OASIS3 code was largely inspired by. I didn't dare touch these comments too much without a reviewing. Indeed, it's on nifti paths. I could have done it for this one since it was obsolete.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: No, let's keep the focus on your PR.

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: (After discussions with @14thibea , it was requested to )
"# according to the new BIDS convention for PET, to use the keyword trc instead of acq. 
# in the same vein to remove the optional task keyword which is useless because it is always rest
# that the minIP files look like the list given on page 10 of this document https://www.oasis-brains.org/files/OASIS-3_Imaging_Data_Dictionary_v1.8.pdf --> so we would put the swi and minIP files in a swi/ folder, and we would make the bids-validator ignore it since it is not in the standard."

Maybe a proofreading from her as well ... to see if I did something wrong ;)

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: Then let's remove the comment and rename the variable to `nifti_files`

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: Thanks for clarifying. Then please add a comment as a preamble to this code explaining all this to make the necessary context explicit.

clinica/iotools/bids_utils.py
ecamenen: Unlike for example OASIS1 where the data extraction procedure results in folders and files with very standardized formats, the procedure under OASIS3 is not standardized at all with a messy tree folders. It was therefore necessary to perform a regex to retrieve only the nifti.

ghisvail: I see, thanks for clarifying. Are we sure this change will not impact other converters which use this utiliy function?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: I don't think so or I don't understand either :) I have to parse the path of the json or nifti, but the format of the OASIS3 files and folders is quite variable. For example, I can have : **/(anat|func)?/.*\.(json|nii\.gz). Either, the json or the nifti can be or not be in a folder called anat or func and be in a variable number of parent folders (1 or 2).

clinica/iotools/bids_utils.py
ecamenen: I'm not really a Python programmer: can you make me a code proposal to understand?

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ecamenen: @14thibea, do you know the origin of that piece of code?

clinica/iotools/bids_utils.py
ecamenen: Normally, each directory should eventually contain a nifiti. I have done tests for OASIS1, I don't have the files (data-test) for the other converters. I thought the Jenkins automated tests checked that based for exemple on those in test/nonregression/test_run_iotools.py. If needed, we can ask @mdiazmel for the data-test files for other converters that I can run by hand.

ghisvail: Agreed

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: Ok then, please add a comment prior to this code with your reasoning 👍

clinica/iotools/converters/oasis3_to_bids/oasis3_to_bids.py
ghisvail: It's based on the CDR score, so it makes sense that anything strictly above 0 is coded `AD` otherwise `CN`. So it would be best to replace this code with something like `paticipants_df["diagnosis_bl"] = paticipants_df["diagnosis_bl"].fillna(0).apply(lambda x: "AD" if x > 0 else "CN")`. That should handle all cases including missing values.

clinica/iotools/bids_utils.py
ghisvail: There are cases where we have got a sidecar JSON file (containing additional metadata as prescribed by the BIDS spec) alongside the Nifti that we would want to copy.

ghisvail: Are there any files in OASIS-3 that we definitely don't want to copy?

ecamenen: OK, are there JSON metadata files in other DBs? It seemed to me that OASIS1 was fairly standard and did not contain any, but maybe it was just the data-test I was provided. In my case, these json files are copied together with the nifti in converters/oasis3_to_bids (convert_images; line 138).

OASIS3 does contain a number of files that should not be copied (e.g., images of post-process analysis results, etc). To get an idea, a test dataset I used can be downloaded [here](https://owncloud.icm-institute.org/index.php/s/FvfVmXPwYJO0JNx). Some examples of input image files are in/unorganized (and reference outputs in ref/). (If needed, tests are done with the following command: pytest --verbose nonregression/test_run_iotools.py -k "test_run_Oasis3ToBids")

ghisvail: I was about to ask whether you could share a link to some example data. I have requested access to OASIS-3 but it's still pending approval.

> OK, are there JSON metadata files as inputs in other DBs?

That may happen in newer datasets (the Harvard one comes to mind).

Another thing to consider is that other modalities may not produce nifti files, but some other native research format  (like bvec / bval files for DTI). Those would not be captured with your change.

I think what needs to be done is ensuring that you only copy the right files from OASIS-3 in a staging BIDS directory, and only then call this utility to write the corresponding scan files.

ecamenen: Oops! my bad: I rerun the code line by line and at this step ... this bids procedure is called in the convert_clinical_data function of the converter ... which takes place after the convert_images step, the file ordering procedure!

So ... when one run write_scans_tsv from bids_util called by the converter ... the files are already ordered according to the bids standards. And there was indeed no need to parse the nifti paths.

That's why I removed my regex in the last commit. OFC, the OASIS1 and OASIS3 diff tests (expected vs observed) work.

clinica/iotools/bids_utils.py
```diff
@@ -459,11 +459,8 @@ def write_sessions_tsv(bids_dir, sessions_dict):
         if bids_id in sessions_dict:
             columns = list(list(sessions_dict.values())[0].values())[0].keys()
             session_df = pd.DataFrame(columns=columns)
-            it = 0
             for k in sessions_dict[bids_id].keys():
-                session_dict = sessions_dict[bids_id][k]
-                session_df_temp = pd.DataFrame(session_dict, index=[k, ])
-                session_df = pd.concat([session_df, session_df_temp])
+                session_df.loc[k] = sessions_dict[bids_id][k]
```
ghisvail: you could even simplify this with:

```python
for k, v in sessions_dict[bids_id].items():
    session_df.loc[k] = v
```

which to me sounds like you could do it once with something like:

```python
pd.DataFrame.from_dict(sessions_dict)
```

and avoid iterating through the dictionary?

ecamenen: Your simplified loop gives the expected result:
image

Your one-line code gives an array that should be transposed: 
image

ghisvail: > Your one-line code gives an array that should be transposed:

What happens if you add this option: `(..., orient="index"`)

ecamenen: This modification below is OK for you?
image

ghisvail: I think you should be able to build `session_df` in one go without the need to transpose (which is not idiomatic). Also aliasing DataFrame to DT is not great either. People may expect aliases for package names (like `pandas` -> `pd`) but usually not for datatypes.

ecamenen: I had not seen your previous comment ... nicel, it works! I ll commit.

ghisvail: What's the format of the `session_dict`and the expected DataFrame you'd want out of it?

ghisvail: > I had not seen your previous comment ... nice, it works! I ll commit.

Great. Thanks for your patience :-)

ecamenen: image

ghisvail (approved):

mdiazmel (approved):
Thanks @ecamenen , great job! 

LGTM. From what was discussed internally, tests and documentation will be added in a next PR.

I'll merge this PR.

PR #295, opened 2021-06-24
Solve incompatibility issue for CreateMergeFile test

mdiazmel:
There is an issue when running `test_run_CreateMergeFile`. The output from different OS is different (probably due to rounding problems of numpy & co). To avoid this problem, numeric values from dataframes are rounded to 12 decimals when the tsv files are written by the function `create_merge_file`.

### DIFF
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -164,8 +164,8 @@ def create_merge_file(
     merged_df = merged_df[col_list]
 
     tmp = merged_df.select_dtypes(include=[np.number])
-    # Round numeric values in dataframe to 12 floating point values
-    merged_df.loc[:, tmp.columns] = np.round(tmp, 12)
+    # Round numeric values in dataframe to 6 decimal values
+    merged_df.loc[:, tmp.columns] = np.round(tmp, 6)
     merged_df.to_csv(out_path, sep="\t", index=False)
     cprint("End of BIDS information merge.", lvl="debug")
 
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -85,7 +85,10 @@ def create_merge_file(
         ]
         row_participant_df.reset_index(inplace=True, drop=True)
         if len(row_participant_df) == 0:
-            cprint(msg=f"Participant {subject} does not exist in participants.tsv", lvl="warning")
+            cprint(
+                msg=f"Participant {subject} does not exist in participants.tsv",
+                lvl="warning",
+            )
             row_participant_df = pd.DataFrame([[subject]], columns=["participant_id"])
 
         if ignore_sessions_files:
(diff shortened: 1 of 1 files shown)


clinica/iotools/utils/data_handling.py
```diff
@@ -160,6 +163,9 @@ def create_merge_file(
     col_list.insert(1, col_list.pop(col_list.index("session_id")))
     merged_df = merged_df[col_list]
 
+    tmp = merged_df.select_dtypes(include=[np.number])
+    # Round numeric values in dataframe to 12 floating point values
+    merged_df.loc[:, tmp.columns] = np.round(tmp, 12)
```
omar-rifai, line 168: It seems that we are the one that normalize these scores to between 0-1, I think we can round this to something more reasonable

omar-rifai (commented):
Thanks for this fix. The ADNI fields that caused the issues are in the data file ADASSCORE.csv. In the file they are not normalized. I think we can round to something less precise and merge!

clinica/iotools/utils/data_handling.py
mdiazmel: What it would be more reasonable ? 6 decimals, for example ?

ghisvail: Depends on the range and order of accuracy of the source values.

mdiazmel: I suppose that different clinical data (coming from different datasets) has different range of values. Can we deduce an order of accuracy that is common to all the dataset processed by Clinica ?

omar-rifai: I'm currently investigating the issue again by looking into the code of create_tsv. Do you remember if the rounding issue was with adas_Qx scores or something else?

omar-rifai: It seems (part of the problem) might come from the files [atlas statistics](https://aramislab.paris.inria.fr/clinica/docs/public/latest/Pipelines/T1_Volume/#t1-volume-volume-based-processing-of-t1-weighted-mr-images-with-spm) tsv files generated by the `t1-volume`pipeline. It might then indeed be relevant to keep the precision that high for these cases as it is generated by a software.

mdiazmel: Yes, I confirmed that these values come from the atlases used in `t1-volume` and `pet-volume` (column names: `t1-volume_group-ADNIbl_atlas*` and `pet-volume_group-ADNIbl_atlas-*`). Do we keep all these decimal values?

omar-rifai: After discussion with @nburgos,  6 decimal digits seems to be largely sufficient for these values which are averages. If no objection, we can change to that and merge.

mdiazmel: Done! 
I'll merge when the CI pass.

PR #297, opened 2021-06-24
Apply session filter to the correct column in ADNIMERGE

omar-rifai:
Fix a bug in ADNI converter that prevented the calculation of the correct age in the session files.

### DIFF
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -7,10 +7,10 @@ and gives to the user some other utils to work with the pipelines.
 """
 import click
 
-from .engine.template import cli as generate_cli
-from .iotools.converters.cli import cli as convert_cli
-from .iotools.utils.cli import cli as iotools_cli
-from .pipelines.cli import cli as run_cli
+from clinica.engine.template import cli as generate_cli
+from clinica.iotools.converters.cli import cli as convert_cli
+from clinica.iotools.utils.cli import cli as iotools_cli
+from clinica.pipelines.cli import cli as run_cli
 
 CONTEXT_SETTINGS = dict(
     # Extend content width to avoid shortening of pipeline help.
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -650,7 +650,7 @@ def create_adni_sessions_dict(
             cprint(f"\tReading clinical data file: {location}")
 
             df_file = pd.read_csv(file_to_read_path, dtype=str)
-            df_filtered = filter_subj_bids(df_file, location, bids_ids)
+            df_filtered = filter_subj_bids(df_file, location, bids_ids).copy()
 
             if not df_filtered.empty:
                 df_filtered["session_id"] = df_filtered.apply(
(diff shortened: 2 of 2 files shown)


clinica/cmdline.py
```diff
@@ -7,10 +7,10 @@
...
 import click
 
-from .engine.template import cli as generate_cli
-from .iotools.converters.cli import cli as convert_cli
-from .iotools.utils.cli import cli as iotools_cli
-from .pipelines.cli import cli as run_cli
+from clinica.engine.template import cli as generate_cli
+from clinica.iotools.converters.cli import cli as convert_cli
+from clinica.iotools.utils.cli import cli as iotools_cli
+from clinica.pipelines.cli import cli as run_cli
```
ghisvail, line 13: That should not be necessary. You should fix your `pylint` setup instead.

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -650,17 +650,16 @@ def create_adni_sessions_dict(
             cprint(f"\tReading clinical data file: {location}")
 
             df_file = pd.read_csv(file_to_read_path, dtype=str)
-            df_filtered = filter_subj_bids(df_file, location, bids_ids)
+            df_filtered = filter_subj_bids(df_file, location, bids_ids).copy()
```
ghisvail, line 653: Why is an explicit copy required?

ghisvail: Could it be fixed in `filter_subj_bids` instead by returning `df.loc[condition, :]` instead of `df[conditon]`?

ghisvail (changes requested):

clinica/cmdline.py
omar-rifai: the issue is not with the pylint but with the python debugger which can not find this relative path. It seems to be an issue only with this file (as it is top level). 
Is there a strong reason against being explicit with the path here?

ghisvail: If it's only this file then ok (though debugging works on PyCharm so there might be something we are missing).

clinica/iotools/converters/adni_to_bids/adni_utils.py
omar-rifai: I can look into it, but as the pattern to create a new column is the same as for the code below I suspect I'd need to explicitly copy at some point.

ghisvail: I think it's best to have it fixed properly in what's returned by the function.

> the pattern to create a new column is the same as for the code below I suspect I'd need to explicitly copy at some point.

You could avoid it with [pandas.DataFrame.assign](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.assign.html).

ghisvail (approved):

ghisvail (approved):

clinica/iotools/converters/adni_to_bids/adni_utils.py
omar-rifai: Good call. I fixed it in `filter_subj_bids` but it seems it doesn't solve the `SettingWithCopyWarning` because of later functions which would ideally require a different architecture (file reader for each clinical file as discussed).  I suggest to leave it in its current state for this PR and use a different PR to change architecture if needed.

PR #298, opened 2021-06-24
Add an init file to the non regression folder

MatthieuJoulot:
Add an __init__.py file making the non regression folder a package.


omar-rifai:
Great, thank you.

PR #300, opened 2021-06-26
Fix unresolved reference in SVM pipeline

ghisvail:
Bug introduced by an incomplete renaming of an argument (cc @mdiazmel).


mdiazmel (approved):
LGTM
I'll merge this one!

PR #311, opened 2021-07-01
Use makedirs(folder, exist_ok=True) function

alexandreroutier:
Found a few ones :)


mdiazmel:
Thanks @alexandreroutier ,

LGTM. 
At some moment in the future, we can maybe think on using the [Pathlib module](https://docs.python.org/3/library/pathlib.html#) that provides more methods to perform this kind of validations.

I'll merge this one!

PR #314, opened 2021-07-05
Enh tau amyloid status session tsv

MatthieuJoulot:
On ADNI, add the column tau_stat (tau status) and a_stat (amyloid status) to the sessions.tsv, computed with the adni_ptau for the tau_stat, and computed with adni_av45, adni_pib and adni_abeta for the a_stat.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -472,6 +472,34 @@ def write_adni_sessions_tsv(df_subj_sessions, bids_subjs_paths):
     import os
     from os import path
 
+    def compute_amyloid_status(row, tau_status=False):
+        stat = ""
+        if (
+            pd.isnull(row["adni_av45"])
+            and pd.isnull(row["adni_pib"])
+            and isinstance(row["adni_abeta"], float)
+        ):
+            stat = "Au"
+        elif row["adni_av45"] > 1.1 or row["adni_pib"] > 1.5 or row["adni_abeta"] < 192:
+            stat = "A+"
+        elif (
+            (pd.isnull(row["adni_av45"]) or row["adni_av45"] < 1.1)
+            and (pd.isnull(row["adni_pib"]) or row["adni_pib"] < 1.5)
+            and (isinstance(row["adni_abeta"], float) or row["adni_abeta"] > 192)
+        ):
+            stat = "A-"
+        return stat
+
+    def compute_ptau_status(row):
+        stat = ""
+        if pd.isnull(row["adni_ptau"]):
+            stat = "Tu"
+        elif row["adni_ptau"] > 23:
+            stat = "T+"
+        else:
+            stat = "T-"
+        return stat
+
     df_subj_sessions["adas_memory"] = (
         df_subj_sessions["adas_Q1"]
         + df_subj_sessions["adas_Q4"]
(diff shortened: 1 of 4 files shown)


clinica/utils/inputs.py
```diff
@@ -306,15 +309,13 @@ def clinica_file_reader(
...
-                f"\t* ({sub} | {ses}): No file found\n"
-            )
+            error_encountered.append(f"\t* ({sub} | {ses}): No file found\n")
         # Otherwise the file found is added to the result
         else:
             results.append(current_glob_found[0])
 
     # We do not raise an error, so that the developper can gather all the problems before Clinica crashes
-    if len(error_encountered) > 0 and raise_exception is True:
+    if len(error_encountered) > 0:
```
omar-rifai, line 318: This should not be in this PR.

clinica/utils/inputs.py
```diff
@@ -327,10 +328,22 @@ def clinica_file_reader(
...
         for msg in error_encountered:
             error_message += msg
-        if is_bids:
-            raise ClinicaBIDSError(error_message)
-        else:
-            raise ClinicaCAPSError(error_message)
+
+        # If the input directory is not complete, as in files are missing, but not all of them,
+        # print an error message
+        if len(results) > 0:
```
omar-rifai, line 334: Same as above (not in this PR)

omar-rifai (changes requested):
Hi @MatthieuJoulot , thanks for the PR. But please create a separate one for each task.

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -491,6 +521,19 @@ def write_adni_sessions_tsv(df_subj_sessions, bids_subjs_paths):
 
     df_subj_sessions = df_subj_sessions.fillna("n/a")
 
+    # compute the amyloid and ptau status
+    df_subj_sessions[
+        ["adni_av45", "adni_pib", "adni_abeta", "adni_ptau"]
+    ] = df_subj_sessions[["adni_av45", "adni_pib", "adni_abeta", "adni_ptau"]].apply(
+        pd.to_numeric, errors="coerce"
```
omar-rifai, line 528: Looking at the [doc](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html), I see that `coerce` will force non-numeric data to NaN.

However, in the [BIDS specs](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#tabular-files), missing values are encoded as "n/a". 

As a result, I think we should transform to numeric "on the fly" when calling the `compute_amyloid_status` and `computer_ptau_status` instead.

MatthieuJoulot: Thanks, I have corrected this issue, the data frames columns used as arguments now remain with n/a instead of being changed to NaN

ghisvail: I think it is a bit premature to translate n/a values here. Encoding missing values should happen at write time and `DataFrame.to_csv` has a specific argument (`na_rep`) to enable this.

omar-rifai: @ghisvail  are you referring to l.522? I agree it could be migrated to write time. I created an [issue](https://github.com/aramis-lab/clinica/issues/326) to clean up the code a bit. We could solve it in a new PR.

omar-rifai:
LGTM, will merge.

omar-rifai (approved):

Issue #302, opened 2021-06-28
Improve the pipeline help docs


ghisvail:
From our recent discussion on the topic, our CLI help docs for running pipelines could benefit for a little more polish and consistency. For instance, it is assumed the user is familiar with the content of our [getting started](https://aramislab.paris.inria.fr/clinica/docs/public/latest/InteractingWithClinica/) page but it's not mentioned in the help docs as prerequisites.

Suggestions for improvement:

- [ ] Explictly list the prerequisites for running the pipeline.
- [ ] Add a representative example for running the pipeline. 
- [ ] Rename and add a description for each pipeline option group. The current proposal is:
    * pipeline_options -> pipeline_specific_options (options specific to the pipeline being run)
    * standard_options -> common_pipeline_options (options common to all Clinica pipelines)
    * advanced_options -> advanced_pipeline_options (for experts only)

PR #316, opened 2021-07-05
Enhance description of CLI for pipelines

omar-rifai:
This PR addresses [Issue 302](https://github.com/aramis-lab/clinica/issues/302) by enhancing the documentation in the CLI help strings


clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_cli.py
```diff
@@ -89,7 +92,9 @@
         "`segm-whitematter_probability.nii.gz`"
     ),
 )
-@cli_param.option_group.standard_options
+@optgroup.group(
+    "Common pipeline options", help="Options common to all Clinica pipelines"
+)
```
alexandreroutier, line 97: What about defining this group in [option_group.py](https://github.com/aramis-lab/clinica/blob/dev/clinica/pipelines/cli_param/option_group.py)? This would avoid copy/paste in other `*.cli.py` files

omar-rifai: Indeed much better ! thanks.

mdiazmel:
Hi @omar-rifai ,

I did a pass for all over the "long descriptions" of the CLI objects. 

The convention that I used to uniform all messages is the following:

```python
def cli(
...
) -> None:
"""Short description of the pipeline
   
   ARGUMENT N description of the argument if difficult to deduce from its name.

   ARGUMENT N+1 

Prerequisite: You have to perform the XXXXX pipeline before running this pipeline

See https://aramislab.paris.inria.fr/clinica/docs/public/latest/Pipelines/XXXXX/ 

```

I let you the hand if you consider that it can be merged.

PR #324, opened 2021-07-12
Remove check for ANTSPATH

### LINKED ISSUE #255: Do not rely on ANTSPATH for checking ANTs CLI availability
As long as the `ants` cli is registered somewhere in `$PATH`, there is no need to proxy check for `$ANTSPATH`. For instance, if `ants` is installed through Homebrew or Anaconda (maybe later) then `$ANTSPATH` is completely superfluous.

ghisvail:
ANTs binaries should already be in PATH. No need for this extra check.

Closes #255


omar-rifai:
Looks, good. Thanks.

PR #325, opened 2021-07-12
Add doc for OASIS3

ecamenen:


docs/Converters/OASIS3TOBIDS.md
```diff
@@ -0,0 +1,166 @@
...
+
+!!! note
+    You do not have to modify the original folder name before using the converter.
+
+!!! warning
+    We do not currently support the conversion of OASIS-2.
+
+## Modalities supported
+
+[REMOVE?] Currently, the converter converts only the T1-weighted MRI images and clinical data.
```
ghisvail, line 86: `[REMOVE?]`

what's that?

ghisvail (commented):
The documentation is great. Just wondering what's the purpose of the `[REMOVE?]` tags.

docs/Converters/OASIS3TOBIDS.md
ecamenen: Not being an imaging expert at all, I wanted to make sure I could remove those sentences ;)

ghisvail: Well for a non imaging expert, your docs are pretty well written :-)

Do you mind if we perform the last edits ourselves before merging?

ecamenen: Aramis team complete the documentation? No worries: feel free to do it. Or don't hesitate to tell me what is missing.

MatthieuJoulot:
Hi. 
I feel like the documentation is missing a part explaining what imaging data to download and how to download it, since the Xnat downloading interface is confusing regarding the "scan format": NIFTI is the image, and BIDS is the data that comes with it so you need both. Moreover, since the converter doesn't convert PUP nor Freesurfer, it would be good to precise to not download it, since it makes the image downloading part longer. 
Same goes for the scan types: it would be nice to only download what the converter can handle, from what I understand T1w right now.
If it is okay with you I can edit and make the changes ?

ghisvail (approved):

PR #328, opened 2021-07-13
Fix the filter_subj_bids function

MatthieuJoulot:
Fix the filter_subj_bids function, which did not apply the changes to the dataframe and caused the converters to crash.


omar-rifai:
Good catch, thanks!

PR #334, opened 2021-07-16
Edit 'Interacting with Clinica' section to be coherent with recent changes in CLI

mdiazmel:
Change arguments typography and add `Common options` subsection to "Interacting with Clinica"


omar-rifai:
Thanks.

PR #342, opened 2021-07-23
Raise exception on empty dataframe

omar-rifai:
Raise explicit exception when creating clinical data and the image folders are not found.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -715,6 +715,8 @@ def create_adni_sessions_dict(
                 ]
                 df_subj_session = pd.concat([df_subj_session, df_filtered], axis=1)
 
+    if df_subj_session.empty:
+        raise Exception("Empty dataset detected. Clinical data can not be extracted")
     df_subj_session.drop(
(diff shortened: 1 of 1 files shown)


clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -715,6 +715,8 @@ def create_adni_sessions_dict(
                 ]
                 df_subj_session = pd.concat([df_subj_session, df_filtered], axis=1)
 
+    if df_subj_session.empty:
+        raise Exception("Empty dataset detected. Clinical data can not be extracted")
```
ghisvail, line 719: I'd suggest to use a `ValueError` instead of the generic `Exception` type. The latter is meant to be subclassed rather than used directly.

ghisvail (changes requested):

clinica/iotools/converters/adni_to_bids/adni_utils.py
ghisvail:
```suggestion
        raise ValueError("Empty dataset detected. Clinical data can not be extracted")
```

omar-rifai: Good call. Changed.

ghisvail (approved):

PR #347, opened 2021-07-27
Raise minimum Python version to 3.8

ghisvail:
Python 3.8 was released end of 2019 and is the default version the [current Ubuntu LTS](https://packages.ubuntu.com/focal/python3). It brought some significant [improvement](https://docs.python.org/3/whatsnew/3.8.html) compared to Python 3.7 some of which would strongly benefit current and future development of Clinica, including:

- Significant performance increase on file tree manipulation through the use of native kernel calls
- Better ergonomics for `shutil.copytree` which is used by our converters for merging BIDS datasets
- Stabilization of `asyncio.run` (which the backend for the future webapp will use extensively)
- Significant perfomance increase of the runtime overall

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ environment and activate it before installing clinica (you can also use
 `virtualenv`):
 
 ```sh
-conda create --name clinicaEnv python=3.7
+conda create --name clinicaEnv python=3.8
 conda activate clinicaEnv
 ```
 
--- a/clinica/__init__.py
+++ b/clinica/__init__.py
@@ -12,9 +12,9 @@ version = __version__
 # version = pkg_resources.require("Clinica")[0].version
 # __version__ = version
 
-# python 3.6 minimum version is required
-if sys.version_info < (3, 6):
-    print(f"Clinica {__version__} requires Python 3.6")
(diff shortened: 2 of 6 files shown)


ghisvail:
I have allowed the use of Python 3.9 as well since the next stable release of Debian (scheduled this fall) will have version 3.9 as default and Anaconda provides it as well.

ghisvail:
CI fails to deploy the conda env because the MacOS version used by the runners is very old and its build tools are quite outdated (xgboost tries to build from source instead of using its wheel package and fails because CMake is too old).

It should run fine on current MacOS releases.

Besides, we are mixing conda and pip packages in the deployment process which is usually not recommended. But that's a different issue.

ghisvail:
> CI fails to deploy the conda env because the MacOS version used by the runners is very old and its build tools are quite outdated (xgboost tries to build from source instead of using its wheel package and fails because CMake is too old).

I managed to bypass the issue using the appopriate version restriction, which I have documented in the commit message. This restriction should be removed once our CI setup for macos is upgraded appropriately.

ghisvail:
Closing, to be reconsidered later, maybe.

PR #352, opened 2021-07-28
 Fix DLPD checks after default values have changed by click introduction

omar-rifai:
The default values of some options have changed after the migration to Click. Consequently the checks in the code of the form option == None no longer worked. Updated the pipeline with consistent checks.

### DIFF
--- a/clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
+++ b/clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
@@ -145,9 +145,7 @@ class DeepLearningPrepareData(cpe.Pipeline):
                     self.masks_location,
                     self.roi_list,
                     self.mask_pattern,
-                    None
-                    if not self.parameters.get("use_uncropped_image")
-                    else not self.parameters.get("use_uncropped_image"),
+                    not self.parameters.get("use_uncropped_image"),
                 )
         else:
             self.masks_location = ""
@@ -351,10 +349,8 @@ class DeepLearningPrepareData(cpe.Pipeline):
         )
         extract_roi_node.inputs.masks_location = self.masks_location
         extract_roi_node.inputs.mask_pattern = self.mask_pattern
-        extract_roi_node.inputs.cropped_input = (
-            None
-            if not self.parameters.get("use_uncropped_image")
-            else not self.parameters.get("use_uncropped_image")
+        extract_roi_node.inputs.cropped_input = not self.parameters.get(
+            "use_uncropped_image"
         )
         extract_roi_node.inputs.roi_list = self.roi_list
         extract_roi_node.inputs.uncrop_output = self.parameters.get("roi_uncrop_output")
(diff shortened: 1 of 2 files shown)


omar-rifai:
@ghisvail if you test and it works, please merge directly.

clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
```diff
@@ -138,15 +138,15 @@ def build_input_node(self):
...
                 raise ValueError("A list of regions must be given.")
             else:
                 check_mask_list(
                     self.masks_location,
                     self.roi_list,
                     self.mask_pattern,
                     None
-                    if self.parameters.get("use_uncropped_image") is None
+                    if not self.parameters.get("use_uncropped_image")
                     else not self.parameters.get("use_uncropped_image"),
```
ghisvail, line 150: This part of the code gets me very confused for the following reasons:

- Use of double negation: we are doing a boolean negation of a boolean flag named `use_uncropped_image`.
- The value computed is in fact not boolean but ternary since it also accepts `None`. This looks intended since further in the code path, in the `cropping` parameter of the `find_mask_path` function, there is a 3-path branching being done.

Besides the questionable sanity of the code, this won't work with Click. `use_uncropped_image` will always receive a value since it is declared as a flag. If we really need to distinguish between 3 cases, then we should be using an enum.

clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
```diff
@@ -353,7 +353,7 @@ def build_core_nodes(self):
         extract_roi_node.inputs.mask_pattern = self.mask_pattern
         extract_roi_node.inputs.cropped_input = (
             None
-            if self.parameters.get("use_uncropped_image") is None
+            if not self.parameters.get("use_uncropped_image")
             else not self.parameters.get("use_uncropped_image")
```
ghisvail, line 357: Same comment as above.

ghisvail (changes requested):

clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
omar-rifai: Good catch on the `find_mask_path' three path branching. I updated the code to reflect the use of a boolean variable. I agree that the double negation can be confusing but I opted for minimal-changes-that-make-this-work for now as:
-The function is meant to migrate soon
- Input from @ravih18 would be valuable before refactoring

clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
omar-rifai: see reply above.

clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
ghisvail: Which means now this line can be changed to just `cropping=not self.parameters.get("use_uncropped_image")`

omar-rifai: Fixed

ghisvail (approved):

PR #353, opened 2021-07-29
Fix default value of modulate flag

### LINKED ISSUE #351: T1-volume --modulate is inconsistent
The option --modulate in t1-volume seems to be set by default to False in a subset of the subpipelines. A coherent behavior across sub-pipelines needs to be specified

ghisvail:
The switch to Click changed the behavior of the `--modulate` flag which used to be set to true by default. This PR brings back the old behavior. Use `--no-modulate` to explicitly disable modulation.

Here is the corresponding output of the help string:

```sh
clinica run t1-volume -h
[...]
    --modulate / --no-modulate    Modulate output images, no modulation preserves concentrations.  [default: modulate]
```

Closes #351

### DIFF
--- a/clinica/pipelines/t1_volume_dartel2mni/t1_volume_dartel2mni_cli.py
+++ b/clinica/pipelines/t1_volume_dartel2mni/t1_volume_dartel2mni_cli.py
@@ -27,7 +27,7 @@ def cli(
     group_label: str,
     smooth: List[int] = (8,),
     tissues: List[int] = (1, 2, 3),
-    modulate: bool = False,
+    modulate: bool = True,
     voxel_size: Tuple[float, float, float] = (1.5, 1.5, 1.5),
     subjects_sessions_tsv: Optional[str] = None,
     working_directory: Optional[str] = None,
--- a/clinica/pipelines/cli_param/option.py
+++ b/clinica/pipelines/cli_param/option.py
@@ -37,9 +37,9 @@ low_bval = option(
 )
 
 modulate = option(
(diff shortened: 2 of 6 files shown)


ghisvail:
After inspection of the code, I confirm the value of the `modulate` flag is passed verbatim to the nipype interface for Dartel.

clinica/pipelines/t1_volume_dartel2mni/t1_volume_dartel2mni_cli.py
```diff
@@ -27,7 +27,7 @@ def cli(
     group_label: str,
     smooth: List[int] = (8,),
     tissues: List[int] = (1, 2, 3),
-    modulate: bool = False,
+    modulate: bool = True,
```
omar-rifai, line 30: Can't we indicate somehow that there are both modulate and no modulate flags in here?

ghisvail: `modulate` here is the variable receiving the boolean value. The flag definitions are in the corresponding `Click.option` which eventually provides its value. Not much to be done further.

ghisvail: I added an explicit `"modulate"` target variable name for what it's worth.

omar-rifai:
Can you please also update the doc to reflect the changes?

ghisvail:
Done

PR #354, opened 2021-07-29
Add explicit disable flags where appropriate

ghisvail:
Following the [issue](#351) with the modulate flag, add explicit disable flags to other boolean inputs where it makes sense.

One-way flags like overwrite_x or force_y are left as is, since it's obvious they are modifiers for the normal execution.


ghisvail:
@omar-rifai would be nice to get your input on this, i.e. whether we want to apply what we did for `modulate` to other pipeline flags or not.

clinica/pipelines/cli_param/option.py
```diff
@@ -24,8 +24,10 @@
 
 dont_save_warped_unmodulated = option(
     "-dswu",
-    "--dont_save_warped_unmodulated",
-    is_flag=True,
+    "--dont_save_warped_unmodulated/--save_warped_unmodulated",
+    "dont_save_warped_unmodulated",
+    default=False,
```
omar-rifai, line 29: does `default=False` here means: `save_warped_unmodulated`?  Regardless, I think that for consistency we ought to have the non-negated action first (i.e save_warped_unmodulated/dont_save_warped_unmodulated). What do you think?

ghisvail: > does default=False here means: save_warped_unmodulated?

Yes

> I think that for consistency we ought to have the non-negated action first. What do you think?

What matters is the logic of the target variable. Here the code expects a boolean with the behaviour of `dont_save_warped_unmodulated`, not its complement. If we swap the values as you propose (so essentially turning `dont_save_warped_unmodulated` into `save_warped_unmodulated`), then we'd also need to complement the actual value everywhere this boolean is used in code.

In the end, it's nitpicking and I don't particularly care either or tbh. Your call.

omar-rifai: All right, fair point. I guess we can leave it as is for now.

omar-rifai: Actually, on second thought and after seeing in the code that we end up doing `"save_warped_unmodulated": not dont_save_warped_unmodulated` maybe it is better to have be consistent with the non-negated part first. 
What do you think @mdiazmel  ?

ghisvail: Assuming it's the only location where this boolean is used, then we'd be better off avoiding the double negation indeed.

mdiazmel: > Actually, on second thought and after seeing in the code that we end up doing "save_warped_unmodulated": not dont_save_warped_unmodulated maybe it is better to have be consistent with the non-negated part first.
What do you think @mdiazmel ?

I think that the best option is to keep the original behavior everywhere is possible.

mdiazmel (approved):
LGTM!

ghisvail:
Closing as discussed earlier in the group meeting

PR #359, opened 2021-08-02
fix indices create-merge-tsv

14thibea:
Values in the columns `first_column_index` and `last_column_index` in the summary file were false.
This commit fixes this bug.


omar-rifai:
Thanks. I changed the description to indicate that the bug was in the `index`columns rather than the names. I'll merge.

omar-rifai (approved):

PR #360, opened 2021-08-02
Update create-merge-tsv with output from freesurfer-longitudinal

omar-rifai:
Add function to include the tsv outputs from freesurfer-longitudinal to the merge-tsv file


omar-rifai:
Following a discussion with @DarioSARACINO, the an additional enhancement to the pipeline has been suggested:


- Allow the users to choose between Freesurfer's "Volume" "Surface" or "Thickness" outputs for integration in the final merge-tsv file
- Add an optional command line argument to choose between these three (with default to Volume i.e current behavior)

omar-rifai:
@ghisvail @mdiazmel  this PR is ready for review :) The CI failure is due to the `dataset_description.json` to Oasis3 which is unrelated

ghisvail:
I have rebased onto `dev` to fix CI

clinica/iotools/utils/data_handling.py
```diff
@@ -58,11 +59,14 @@ def create_merge_file(
             raise IOError("The path to the CAPS directory is wrong")
 
     if not os.path.isfile(path.join(bids_dir, "participants.tsv")):
-        raise IOError("participants.tsv not found in the specified BIDS directory")
-    participants_df = pd.read_csv(path.join(bids_dir, "participants.tsv"), sep="\t")
+        participants_df = create_participants_df(bids_dir)
+        # raise IOError("participants.tsv not found in the specified BIDS directory")
```
ghisvail, line 63: Since the BIDS specification does not make `participants.tsv` obligatory, I think this part is a welcome change. However, I think it is redundant with the call to `get_subject_session_list` below it, which will also traverse the BIDS / CAPS tree to list all participants (and respective sessions).

What do you think?

clinica/iotools/utils/data_handling.py
```diff
@@ -235,6 +241,24 @@ def create_merge_file(
         cprint("End of CAPS information merge.", lvl="debug")
 
 
+def create_participants_df(bids_dir: str) -> pandas.DataFrame:
+    """
+    Function to create participant.tsv "on the fly" in case it is not found in the BIDS directory
+    """
+    import pandas as pd
```
ghisvail, line 248: You should probably just `from pandas import DataFrame` for the type signature and use it inside the function without requiring another qualified import here.

clinica/iotools/utils/data_handling.py
```diff
@@ -235,6 +241,24 @@ def create_merge_file(
...
+    import os
+
+    subjects_ids = [
+        name
+        for name in os.listdir(bids_dir)
+        if os.path.isdir(os.path.join(bids_dir, name)) and "sub" in name
+    ]
```
ghisvail, line 255: It might be possible to simplify this part with `Path.glob("sub-*")`?

clinica/iotools/utils/data_handling.py
```diff
@@ -696,7 +720,7 @@ def create_subs_sess_list(
             session_df = pd.read_csv(
                 path.join(sub_path, subj_id + "_sessions.tsv"), sep="\t"
             )
-            session_list = list(session_df["session_id"].to_numpy())
+            session_list = list(session_df["session_id"].dropna().to_numpy())
```
ghisvail, line 723: How come that we can get n/a values for the `session_id` here?

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,10 +60,114 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+
     from clinica.iotools.converters.adni_to_bids.adni_utils import (
         replace_sequence_chars,
     )
     from clinica.utils.stream import cprint
+    from pathlib import Path
+
+    # Ensures that df is correctly indexed
+    if "participant_id" in df.columns.values:
+        df.set_index(["participant_id", "session_id"], inplace=True, drop=True)
```
ghisvail, line 72:
```suggestion
        df.set_index(["participant_id", "session_id"], inplace=True, verify_integrity=True)
```

ghisvail: And what to do if the dataframe is actually not sound (ValueError due to failed integrity check)?

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,10 +60,114 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+            long_id = next(mod_path.glob("long*")).name
+        except StopIteration:
+            cprint(
+                f"Could not find a longitudinal dataset for participant {participant_id} {session_id}",
+                lvl="warning",
+            )
+            flag_skip = True
+
+        if flag_skip:
+            continue
```
ghisvail, line 94:
```suggestion
        try:
            long_id = next(mod_path.glob("long*")).name
        except StopIteration:
            cprint(
                f"Could not find a longitudinal dataset for participant {participant_id} {session_id}",
                lvl="warning",
            )
            continue
```

ghisvail: This should work?

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,10 +60,114 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+
+        if mod_path.exists():
+            # Looking for atlases
+            atlas_paths = sorted(
+                (mod_path).glob(f"{participant_id}_{session_id}_*volume.tsv")
+            )
+            for atlas_path in atlas_paths:
+                if "-wm_volume" not in str(atlas_path) and "-ba_volume" not in str(
+                    atlas_path
+                ):
```
ghisvail, line 111:
```suggestion
                if not any(
                    s in atlas_path.name 
                    for s in ["-wm_volume", "-ba_volume"]
                ):
```

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,10 +60,114 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+                        ses_df[label_list] = atlas_df["label_value"].to_numpy()
+
+            # Always retrieve subcortical volumes
+            atlas_path = next(
+                sorted(
+                    (mod_path).glob(
+                        f"{participant_id}_{session_id}_*segmentationVolumes.tsv"
+                    )
+                )
+            )
```
ghisvail, line 136: What happens if we get a `StopIteration` exception here?

docs/IO.md
```diff
@@ -161,6 +161,8 @@ If an input list of subjects and sessions is given, the merged file will only ga
     ...
     ```
 
+A complete list of  optional arguments can be obtained with the command line `clinica merge-tsv --help`
```
ghisvail, line 164:
```suggestion
A complete list of optional arguments can be obtained with the command line `clinica merge-tsv --help`
```

ghisvail (changes requested):

clinica/iotools/utils/data_handling.py
omar-rifai: Removed. Wasn't used in data_handling

clinica/iotools/utils/data_handling.py
omar-rifai: not relevant anymore as I removed the function

clinica/iotools/utils/pipeline_handling.py
omar-rifai: maybe more elegant but not more readable

omar-rifai: no?

clinica/iotools/utils/data_handling.py
omar-rifai: You are right, I added this function because the code was already using  `participants.tsv` for extracting the subject list. Updated.

clinica/iotools/utils/pipeline_handling.py
ghisvail: Your call. Tomorrow if you need to add another case (say "-foo_volume"), the functional form with `any` is simpler to update (just add the new case to the list) than the imperative form. You can also use DeMorgan's theorem to change the `any` form to an `all` form if it turns out to be clearer to read (it might be).

clinica/iotools/utils/data_handling.py
omar-rifai: It should not be the case Indeed.

clinica/iotools/utils/pipeline_handling.py
omar-rifai: Updated to cprint an error message and exit

clinica/iotools/utils/pipeline_handling.py
omar-rifai: All right will keep it for now. Noted for cases with more options.

clinica/iotools/utils/pipeline_handling.py
omar-rifai: Added try block to print warning and skip if applicable.

clinica/iotools/utils/data_handling.py
```diff
@@ -57,18 +57,27 @@ def create_merge_file(
...
+    try:
+        sub_ses_df.set_index(
+            ["participant_id", "session_id"], inplace=True, verify_integrity=True
+        )
+    except ValueError:
+        cprint(
+            "Found duplicate subject session pair. Please check subjects_sessions list for any misentry.",
+            lvl="error",
+        )
+        return
```
ghisvail, line 83: I wonder why not simply run `pd.DataFrame(...).drop_duplicates().set_index(...)`, or do `pd.DataFrame(...)`, detect duplicates with `.duplicated()`, show a warning and then drop duplicates and set index, instead of a logging an error string followed by an early return?

From a UX perspective, it's probably less frustrating I believe.

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -47,8 +47,67 @@ def pet_volume_pipeline(
...
+    caps_dir: str, df: pd.DataFrame, dti_atlas_selection=None, **kwargs
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+
+    from pathlib import Path
+
+    # Ensures that df is correctly indexed
+    if "participant_id" in df.columns.values:
+        df.set_index(
+            ["participant_id", "session_id"], inplace=True, verify_integrity=True
+        )
```
ghisvail, line 60: I wonder why is `set_index` conditional here, given we are iterating on index values a couple of lines below? We should probably run it unconditionnally and `try / catch` if the input dataframe is not in the expected shape.

clinica/iotools/utils/pipeline_handling.py
ghisvail: Same comment as above, we probably want to unconditionnally run set_index with the expected column names (no if case) to reject unsound input dataframes.

clinica/iotools/utils/pipeline_handling.py
ghisvail: It's also more explicit to test `atlas_path.name` (or even `atlas_path.steam` maybe?) so that we know where the expected pattern is in the full path.

```suggestion
                if "-wm_volume" not in atlas_path.name and "-ba_volume" not in atlas_path.name:
```

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,10 +117,115 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+                atlas_df = pd.read_csv(atlas_path, sep="\t")
+                label_list = [
+                    "t1-freesurfer-segmentationVolumes_" + x
+                    for x in atlas_df.label_name.values
+                ]
+
+                ses_df[label_list] = atlas_df["label_value"].to_numpy()
+
+            except StopIteration:
+                cprint("Segmentation volume file not found. Skipping.", lvl="warning")
```
ghisvail, line 201: Better catch early always.

```suggestion
            try:
                atlas_path = next(
                    sorted(
                        (mod_path).glob(
                            f"{participant_id}_{session_id}_*segmentationVolumes.tsv"
                        )
                    )
                )
            except StopIteration:
                cprint("Segmentation volume file not found. Skipping.", lvl="warning")

            atlas_df = pd.read_csv(atlas_path, sep="\t")
            label_list = [
                "t1-freesurfer-segmentationVolumes_" + x
                for x in atlas_df.label_name.values
            ]
            ses_df[label_list] = atlas_df["label_value"].to_numpy()
```

ghisvail (changes requested):

clinica/iotools/utils/data_handling.py
omar-rifai: Well I suppose we could have it as only a warning. But I suppose if we got to this point then the dateset got corrupted somehow and it might not be the only issue. That was my reasoning

clinica/iotools/utils/pipeline_handling.py
omar-rifai: The lines you removed from the try do not make sense if the file is not found.

clinica/iotools/utils/data_handling.py
ghisvail: It depends on the origin of the `sub_ses` dataframe I guess. If it comes from a path discovery mechanism, then it should be unique. However, human mistakes (like a duplicate line) are possible when coming from a custom file.

I am just thinking that, for the latter case, users might find it frustrating that Clinica fails and ask them to clean their mess first, instead of discarding duplicate pairs and eventually warning them if need be.

I am fine with either. Your call 👍

clinica/iotools/utils/pipeline_handling.py
ghisvail: I see, so it's a completely optional code path then.

clinica/iotools/utils/pipeline_handling.py
omar-rifai: Added stem.

clinica/iotools/utils/pipeline_handling.py
omar-rifai: Correct. That was a mere copy paste from previous code. Granted it could be optimized.

clinica/iotools/utils/data_handling.py
omar-rifai: I mean the "verify integrity" part was your suggestion (!) If you  think `.drop_duplicates().set_index(...)` is better, I can switch it to that.

clinica/iotools/utils/pipeline_handling.py
omar-rifai: Fixed

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -47,8 +47,68 @@ def pet_volume_pipeline(
...
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    from pathlib import Path
+
+    # Ensures that df is correctly indexed
+    try:
+        df.set_index(
+            ["participant_id", "session_id"], inplace=True, verify_integrity=True
+        )
+    except Exception:
+        raise KeyError("Fields `participant_id` and `session_id` are required.")
```
ghisvail, line 61:
```suggestion
    except KeyError:
        raise KeyError("Fields `participant_id` and `session_id` are required.")
```

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,14 +118,125 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
 
     # Ensures that df is correctly indexed
-    if "participant_id" in df.columns.values:
-        df.set_index(["participant_id", "session_id"], inplace=True, drop=True)
+    try:
+        df.set_index(
+            ["participant_id", "session_id"], inplace=True, verify_integrity=True
+        )
+    except Exception:
+        raise KeyError("Fields `participant_id` and `session_id` are required.")
```
ghisvail, line 134:
```suggestion
    except KeyError:
        raise KeyError("Fields `participant_id` and `session_id` are required.")
```

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -58,14 +118,125 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
...
+        replace_sequence_chars,
+    )
+
+    # Ensures that df is correctly indexed
+    try:
+        df.set_index(
+            ["participant_id", "session_id"], inplace=True, verify_integrity=True
+        )
+    except Exception:
+        raise KeyError("Fields `participant_id` and `session_id` are required.")
```
ghisvail, line 239:
```suggestion
    except KeyError:
        raise KeyError("Fields `participant_id` and `session_id` are required.")
```

ghisvail (changes requested):
Last changes from me.

clinica/iotools/utils/data_handling.py
ghisvail: Let's keep it that way for now. If users were to complain, we know we can relax that a little with this alternative.

ghisvail (approved):

PR #370, opened 2021-08-04
Enable OASIS-3 converter

ghisvail:
- Expose the converter to the CLI
- Include OASIS-3 docs to Clinica's
- Add ref to OASIS-3 converter in README


ghisvail:
Would be nice to have it seat in trunk before the release, and test it and the docs one more time.

MatthieuJoulot:
I have tested the converter following the instructions from the documentation, and it doesn't work. I get the following error:

Screenshot 2021-08-06 at 10 12 08

ghisvail:
This PR is placed on hold as it requires a closer look at the clinical, MR and PET data to define a strategy for matching them into a coherent BIDS session.

PR #371, opened 2021-08-04
Fix default parameters for nonregression tests dlpd

omar-rifai:
Fix default parameters for deeplearning prepare data and the non regression tests.

### DIFF
--- a/test/nonregression/pipelines/test_run_pipelines_dl.py
+++ b/test/nonregression/pipelines/test_run_pipelines_dl.py
@@ -44,25 +44,32 @@ def test_run_DLPrepareData(cmdopt):
     roi_params = {
         "extract_method": "roi",
         "roi_list": ["rightHippocampusBox", "leftHippocampusBox"],
-        "roi_uncrop_output": False,
+        # "roi_uncrop_output": False,
+        "use_uncropped_image": True,
     }
 
     data = [image_params, slice_params, patch_params, roi_params]
 
     for parameters in data:
+
         for modality in modalities:
+
             parameters["modality"] = modality
+
             if modality == "pet-linear":
                 parameters["acq_label"] = "av45"
                 parameters["suvr_reference_region"] = "pons2"
                 parameters["use_uncropped_image"] = False
                 DLPrepareData_Generic(root, working_dir, parameters)
+
             elif modality == "custom":
+                parameters["use_uncropped_image"] = True
                 parameters["custom_template"] = "Ixi549Space"
                 parameters[
                     "custom_suffix"
                 ] = "graymatter_space-Ixi549Space_modulated-off_probability.nii.gz"
                 DLPrepareData_Generic(root, working_dir, parameters)
+
             elif modality == "t1-linear":
                 for flag in uncropped_image:
                     parameters["use_uncropped_image"] = flag
--- a/test/nonregression/pipelines/test_run_pipelines_dl.py
+++ b/test/nonregression/pipelines/test_run_pipelines_dl.py
@@ -44,7 +44,6 @@ def test_run_DLPrepareData(cmdopt):
     roi_params = {
         "extract_method": "roi",
         "roi_list": ["rightHippocampusBox", "leftHippocampusBox"],
-        # "roi_uncrop_output": False,
         "use_uncropped_image": True,
     }
 
--- a/clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
+++ b/clinica/pipelines/deeplearning_prepare_data/deeplearning_prepare_data_pipeline.py
@@ -73,7 +73,7 @@ class DeepLearningPrepareData(cpe.Pipeline):
                 FILE_TYPE = T1W_LINEAR_CROPPED
         if self.parameters.get("modality") == "t1-extensive":
             FILE_TYPE = T1W_EXTENSIVE
-            self.parameters["use_uncropped_image"] = None
+            self.parameters["use_uncropped_image"] = True
         if self.parameters.get("modality") == "pet-linear":
             FILE_TYPE = pet_linear_nii(
                 self.parameters.get("acq_label"),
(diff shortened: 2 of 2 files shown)


test/nonregression/pipelines/test_run_pipelines_dl.py
```diff
@@ -44,25 +44,32 @@ def test_run_DLPrepareData(cmdopt):
...
+                parameters["use_uncropped_image"] = True
                 parameters["custom_template"] = "Ixi549Space"
                 parameters[
                     "custom_suffix"
                 ] = "graymatter_space-Ixi549Space_modulated-off_probability.nii.gz"
                 DLPrepareData_Generic(root, working_dir, parameters)
+
             elif modality == "t1-linear":
                 for flag in uncropped_image:
                     parameters["use_uncropped_image"] = flag
```
ghisvail, line 75: Why the added line breaks? Looks unnecessary in the diff.

test/nonregression/pipelines/test_run_pipelines_dl.py
```diff
@@ -44,25 +44,32 @@ def test_run_DLPrepareData(cmdopt):
     roi_params = {
         "extract_method": "roi",
         "roi_list": ["rightHippocampusBox", "leftHippocampusBox"],
-        "roi_uncrop_output": False,
+        # "roi_uncrop_output": False,
+        "use_uncropped_image": True,
```
ghisvail, line 48: I wonder what `roi_uncrop_output` is used for then?

test/nonregression/pipelines/test_run_pipelines_dl.py
mdiazmel: > Why the added line breaks? Looks unnecessary in the diff.

Agree with @ghisvail about the line breaks. Black seems to be flexible with empty lines inside functions but to ease the diff  is better to avoid them.

omar-rifai: For readability ! The function was already really cramped as we are testing several parameters and options but if you think we need to go back, np.

test/nonregression/pipelines/test_run_pipelines_dl.py
omar-rifai: --use_uncropped_image is for _using_ a cropped image as input
--roi_uncrop_output is for _generating_ a cropped tensor as output

test/nonregression/pipelines/test_run_pipelines_dl.py
ghisvail: I was wondering whether it was Black doing this somehow. I think for new code it's a good practice, but for PRs on existing code it can add unwanted noise. But, ymmv.

omar-rifai: Granted that for reviewing the PR it's not great. Apologies for that but I think it is worth keeping them in this case for future maintenance of this function.

test/nonregression/pipelines/test_run_pipelines_dl.py
```diff
@@ -44,25 +44,32 @@ def test_run_DLPrepareData(cmdopt):
     roi_params = {
         "extract_method": "roi",
         "roi_list": ["rightHippocampusBox", "leftHippocampusBox"],
-        "roi_uncrop_output": False,
+        # "roi_uncrop_output": False,
```
mdiazmel, line 47: These two parameter are not exchangeable: `roi_uncrop_output` allows to have cropped versions of the outputs (around the roi) while `use_uncropped_image` indicates if the input images are cropped or not.

mdiazmel (changes requested):
It seems that there is a confusion around the parameters `use_uncropped_image` and `roi_uncrop_output` that are not exchangeable. 

Beyond this, I figured out that our way to propose the tests is not very pythonic. We should use decorators to pass the parameters of the pipelines (as proposed in the Pytest tutorials and in ClinicaDL tests). Maybe it's worth to note this for a future PR.

test/nonregression/pipelines/test_run_pipelines_dl.py
ghisvail: I am glad I asked

omar-rifai: I know they are not exchangeable and I did aim at using the later as per the data for the non-regression. Is there still an issue?

mdiazmel: I had the impression that you were exchanging these parameters... anyway you need to verify if the test reference data uses the cropped version or not. If yes, this line must be uncommented.

omar-rifai: No sorry for the confusion, these were two different changes.
`roi_uncrop_output` was commented so that the output is coherent with the data in the ref folder
`use_uncropped_image` was added because of the masks we have as input are uncropped.

I tested with the parameters locally on the reference data we had and the tests passed on my machine. Unless the data in the CI machines have changed, it should be ok.

omar-rifai:
> Beyond this, I figured out that our way to propose the tests is not very pythonic. We should use decorators to pass the parameters of the pipelines (as proposed in the Pytest tutorials and in ClinicaDL tests). Maybe it's worth to note this for a future PR.
+1

omar-rifai:
@ghisvail, j'ai testé le changement et je pense qu'il va falloir juste enlever ces deux lignes.

 je ne sais pas pourquoi `use_uncropped_image`  avait été fixé à `None` avant et pas à True mais le default = False ne fonctionne pas car il va s'attendre à un mask avec la chaine de caractère `desc_Cropped` alors qu'elle n'existe pas

ghisvail (approved):

omar-rifai:
@mdiazmel, is it OK for you ? If so I think we can merge.

mdiazmel:
> @mdiazmel, is it OK for you ? If so I think we can merge.

It's OK!

PR #373, opened 2021-08-05
Replace is None checks with more idiomatic statements

omar-rifai:


clinica/pipelines/engine.py
```diff
@@ -112,21 +112,18 @@ def __init__(
...
             self._base_dir_was_specified = True
-
-        if name:
-            self._name = name
         else:
-            self._name = self.__class__.__name__
+            self.base_dir = mkdtemp()
+            self._base_dir_was_specified = False
+
+        self._name = name or self.__class__.__name__
```
mdiazmel, line 122: I'm not familiar with this syntax. What's the purpose of this `or` between two instructions?

mdiazmel (commented):
Seems ok, can be merged quickly.

clinica/pipelines/engine.py
omar-rifai: I guess it's a shortcut to assign the variable whose value is not `None`. In cases both are not `None`, the first variable will be assigned.

ghisvail: It's equivalent to a coalesce

ghisvail: i.e. takes the first argument which is not None

clinica/iotools/utils/data_handling.py
```diff
@@ -191,10 +191,10 @@ def create_merge_file(
         if not pipelines:
             for pipeline_name, pipeline_fn in pipeline_options.items():
                 merged_df, summary_df = pipeline_fn(caps_dir, merged_df, **kwargs)
-                if summary_df is not None:
+                if summary_df:
                     merged_summary_df = pd.concat([merged_summary_df, summary_df])
 
-                if summary_df is None or len(summary_df) == 0:
+                if not summary_df or len(summary_df) == 0:
```
ghisvail, line 197: I wonder whether there is a `len(summary_df) > 0` missing in the first test. Otherwise I don't see why the extra check for `len`is necessary 🤔

clinica/iotools/utils/pipeline_handling.py
```diff
@@ -93,7 +93,7 @@ def t1_freesurfer_pipeline(caps_dir, df, freesurfer_atlas_selection=None, **kwar
             for atlas_path in atlas_paths:
                 atlas_name = atlas_path.split("_parcellation-")[1].split("_")[0]
                 if path.exists(atlas_path) and (
-                    freesurfer_atlas_selection is None
+                    not freesurfer_atlas_selection
                     or (
                         freesurfer_atlas_selection is not None
```
ghisvail, line 98:
```suggestion
                        freesurfer_atlas_selection
```

clinica/iotools/utils/data_handling.py
omar-rifai: yes seems like an oversight. Maybe we can add it?

PR #375, opened 2021-08-05
Apply black formatting to DWI-Connectome

alexandreroutier:
Also used f-string formatting as well as calls to some function in `utils/*.py`.

### DIFF
--- a/clinica/pipelines/dwi_connectome/dwi_connectome_pipeline.py
+++ b/clinica/pipelines/dwi_connectome/dwi_connectome_pipeline.py
@@ -1,11 +1,11 @@
 # coding: utf8
 
+import clinica.pipelines.engine as cpe
+
 # Use hash instead of parameters for iterables folder names
 from nipype import config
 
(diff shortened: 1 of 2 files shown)


ghisvail:
I am not convinced of the value provided by this PR close to a release. It's a large reformatting patch and it's time-consuming to review. Better to postpone this reformatting early in the next development cycle and maybe apply it to the whole code base as opposed to just `dwi-connectome`.

ghisvail:
Closing in favor of a future run of `black` and `isort` on the whole code base.

alexandreroutier:
Hi @ghisvail,

The pipeline is the last one that did not have `black` formatting with special attention to node connections (`# fmt: on/off` was used after manual formatting).

Best,
Alex

mdiazmel:
Let me reopen this PR, in fact as @alexandreroutier mentions, `black` and `isort` was already applied to the whole code base (see 1c4ee411) with only a few exceptions. This pipeline is part of these exceptions.

PR #376, opened 2021-08-05
Use Python dict setdefault()

alexandreroutier:
Finish harmonization of `check_pipeline_parameters` method in all pipelines.


ghisvail (approved):

ghisvail:
@alexandreroutier for the record you might be interested in #382

PR #377, opened 2021-08-06
Enh bye freesurfer conv

MatthieuJoulot:
Remove the usage of FreeSurfer as a fallback for dcm2niix conversion in the `aibl-to-bids` converter.

To do:

- [x]  test the aibl database to see whether FreeSurfer is used and is useful to convert. 3 conversion are done using FreeSurfer. 
- [x] Download them again to be sure they were not updated and now work.
- [ ] If they still do not work-> Indicate these cases to dcm2niix so they can fix them.

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -774,6 +774,7 @@ def run_dcm2niix(command):
         cprint(
             msg=(
                 "DICOM to BIDS conversion with dcm2niix failed:\n"
+                f"command: {command}\n"
                 f"{output_dcm2niix.stdout.decode('utf-8')}"
             ),
             lvl="warning",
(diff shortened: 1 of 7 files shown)


ghisvail:
Could you verify there is no longer any ref to freesurfer in the docs for the converters?

MatthieuJoulot:
@ghisvail done !

ghisvail:
> @ghisvail done !

Here: https://github.com/aramis-lab/clinica/blob/dev/docs/Third-party.md#third-party-software

MatthieuJoulot:
@ghisvail good catch, thanks !

ghisvail (approved):

MatthieuJoulot:
- [x] The four T1 cases that did not work with dcm2niix and needed FreeSurfer, work when downloaded again.

They will need to be updated on the Lustre.

PR #378, opened 2021-08-06
Drop minimum version for xgboost

ghisvail:
- Avoid strict version pinning, which is not recommended
- A newer version of CMake is required for xgboost 1.x, which is not available in our MacOS CI

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ numpy >= 1.17
 scikit-learn == 0.23
 nilearn >= 0.6.0
 colorlog >= 5
-xgboost < 1
+xgboost
 xlrd >= 1.2.0, < 2.0.0
 scipy == 1.2.3
 matplotlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ numpy >= 1.17
 scikit-learn == 0.23
 nilearn >= 0.6.0
 colorlog >= 5
-xgboost == 0.80
+xgboost < 1
 xlrd >= 1.2.0, < 2.0.0
 scipy == 1.2.3
 matplotlib


mdiazmel:
>     * Avoid strict version pinning, which is not recommended

Indeed, an idea that we discussed sometimes ago was to keep strict version pinning in the `requirements.txt` for all the dependencies (in order to provide a stable test (CI) environment) and to left minimal/maximal versions in the packaging file (`setup.py`).  This can be done later, in a different PR (it needs some exhaustive testing).

>     * A newer version of CMake is required for xgboost 1.x, which is not available in our MacOS CI

I can also update CMake in the macOS agents, and avoid the version restriction on xgboost. What's the minimal required version to compile xgboost 1.x?

ghisvail:
> This can be done later, in a different PR (it needs some exhaustive testing).

Agreed.

> I can also update CMake in the macOS agents, and avoid the version restriction on xgboost. What's the minimal required version to compile xgboost 1.x?

The minimum CMake version has recently been [raised to 3.14](https://github.com/dmlc/xgboost/commit/b2d300e727b36143a437a328034e1f020dd53e64#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a). So we'd need at least that.

Ideally, it'd be best to upgrade macOS to a recent enough version which allows deployment of the wheel packages available on PyPI instead of compiling from source. That would save time and simplify the CI setup requirements. We'd need at least macOS 10.14 Mojave for that based on the [most recent release to date](https://pypi.org/project/xgboost/1.4.2/#files).

mdiazmel:
> The minimum CMake version has recently been [raised to 3.14](https://github.com/dmlc/xgboost/commit/b2d300e727b36143a437a328034e1f020dd53e64#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a). So we'd need at least that.

Done. Cmake version in macOS agents was upgraded to version 3.21.1.
After upgrading Cmake, I successfully installed `xgboost-1.4.2` in a clean conda environment (wheel package was successfully built). Other dependencies installed with this version of `xgboost` are `numpy-1.21.1` and `scipy-1.7.1`.

> Ideally, it'd be best to upgrade macOS to a recent enough version which allows deployment of the wheel packages available on PyPI instead of compiling from source. That would save time and simplify the CI setup requirements. We'd need at least macOS 10.14 Mojave for that based on the [most recent release to date](https://pypi.org/project/xgboost/1.4.2/#files).

Concerning upgrading macOS, we are currently limited to working with this version. The guys in charge of the infrastructure are working on getting access to recent versions of macOS, but I don't know when this will be possible.

ghisvail:
> Done. Cmake version in macOS agents was upgraded to version 3.21.1.

Great, so I can lift the versioned dependency as a result.

mdiazmel:
Perfect. I merge this one and I'll delete the branch from the repo.

PR #381, opened 2021-08-06
Fix selected default value handling

ghisvail:
As a follow-up for #373 where a few edge-cases were found afterwards.

### DIFF
--- a/clinica/pipelines/machine_learning_spatial_svm/spatial_svm_utils.py
+++ b/clinica/pipelines/machine_learning_spatial_svm/spatial_svm_utils.py
@@ -850,7 +850,7 @@ def heat_finite_elt_2D_tensor2(x0, t_final, t_step, h, g):
 
 
 def heat_solver_tensor_3D_P1_grad_conj(
-    f, g, t_final, h, t_step, CL_value, epsilon: float = 0.1
+    f, g, t_final, h, t_step, CL_value=None, epsilon: float = 0.1
 ):
     """
     It solves the poisson's equation in 1D on the regular mesh (with mesh of size h)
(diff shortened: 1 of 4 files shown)


omar-rifai:
Is there a code-breaking bug or are these all just enhancement? In the later case, I'd prefer to leave the review for after the release.

ghisvail:
> Is there a code-breaking bug or are these all just enhancement?

I can't tell you for sure those changes fix an actual bug if it has not happened yet. They do make the code safer to weird upstream calls like what happened with the `mask_pattern` in the dlpd pipeline.

You asked me to make a second pass on the default values covered in #373, that's what I did. If you feel unneasy about it, then feel free to discard it.

omar-rifai:
No of course, the point was not to discard this PR only postpone the merge it if it didn't fix code-breaking behavior like for dlpd.
The reason I'd rather postpone to after the release is because it also introduces new patterns that I don't feel safe to check  in a hurry.

ghisvail:
> if it didn't fix code-breaking behavior like for dlpd.

Imo, it will prevent similar breakage as dlpd (i.e. by passing None explicitly) that we have not hit yet.

clinica/pipelines/machine_learning_spatial_svm/spatial_svm_utils.py
```diff
@@ -1082,9 +1081,7 @@ def heat_solver_equation(input_image, g, FWHM, t_step, dartel_input):
     input_image_read = nib.load(input_image)
     input_image_data = input_image_read.get_fdata(dtype="float32")
 
-    u = utils.heat_solver_tensor_3D_P1_grad_conj(
-        input_image_data, g, beta, h, t_step, CL_value=None, epsilon=None
-    )
+    u = utils.heat_solver_tensor_3D_P1_grad_conj(input_image_data, g, beta, h, t_step)
```
omar-rifai, line 1084: Why are we removing the default values here?

ghisvail: Because this call will not trigger the default values for `CL_value` and `epsilon`. If we are not going to pass proper values to them, we should not specify them.

PR #389, opened 2021-08-20
Compress preprocessed NIfTI files during install

ghisvail:
I just found out that preprocessed NIfTI files from NIFD are provided uncompressed. This PR ensures they end up compressed at install time.


mdiazmel (commented):
Thanks @ghisvail,

Seems ok! I'll merge.

PR #397, opened 2021-08-27
Promote GitHub discussions for new support requests

ghisvail:
As discussed in our last group meeting, the plan is to freeze our Google group and transition to using GitHub discussions for handling support requests.

GitHub discussions is now stable, provides a better experience for writing support requests, and is not under threat of joining the evergrowing Google [graveyard](https://killedbygoogle.com/).

This PR invites users to first check past answers from Google group, but start new requests on GitHub discussions. This way both channels can still co-exist until we decide on a freeze date for the Google group.


ghisvail:
`/mnt/data/builds/workspace: Input/output error` 🤦

mdiazmel (approved):
Cool! Thanks @ghisvail !

We should remember to update the "Support" link on the [website](http://www.clinica.run/) (and also  the release announce there). 

I'll merge!

PR #401, opened 2021-09-01
Fix bash instruction to init conda

mdiazmel:
Fix Conda initialization in CI scripts.


mdiazmel:
I'll merge this quickly to avoid CI problems in future PRs!

PR #403, opened 2021-09-03
Fix NIFD to BIDS conversion for MRI-only collections

ghisvail:
Also adds more accurate type signatures and prefer `None`over `pd.NA`for BIDS labels to simplify the code.

Closes #402

PR #408, opened 2021-09-06
Fix imaging data finder for alternative NIFD layout

### LINKED ISSUE #406: NIFD converter fails with participants with alternative imaging data layout
The current implementation of the NIFD converter expects imaging data to follow this flat layout:

`/path/to/collection/<imaging_data_folder>/*.dcm`

There is however a different layout for some participants which is nested:

`/path/to/collection/<subject>/<description>/<date>/<identifier>/*.dcm`

So far, the NIFD converter has been tested with participants organized with the flat layout, for instance `1_S_0006` and fails with those organized with the nested structure, for instance `1_S_0001`.

ghisvail:
For both the flat and nested NIFD layouts, the image data ID is encoded in the filenames. This PR updates the image data finder to parse these IDs from the files before computing the image data ID to source directory mapping.

Closes #406

### DIFF
--- a/clinica/iotools/converters/nifd_to_bids/nifd_utils.py
+++ b/clinica/iotools/converters/nifd_to_bids/nifd_utils.py
@@ -95,15 +95,22 @@ def find_imaging_data(imaging_data_directory: PathLike) -> Iterable[Tuple[str, s
     import re
     from pathlib import Path
 
-    pattern = re.compile(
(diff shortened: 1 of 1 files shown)


ghisvail:
Depends on #403 for testing

mdiazmel:
Seems ok!

Let's merge this one in order to test #403!

PR #412, opened 2021-09-08
Update fields in ADNI sessions file

omar-rifai:
Corrects two issues with adni2bids converter:
-Update diagnosis field entries from Dementia to AD 
-Remove rows with Empty or "Nv" value for the sessions

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -718,11 +718,11 @@ def create_adni_sessions_dict(
                 ]
                 df_subj_session = pd.concat([df_subj_session, df_filtered], axis=1)
 
+    df_subj_session = df_subj_session[
+        (~df_subj_session.session_id.isin(["ses-Nv", "sc", False]))
+    ]
     if df_subj_session.empty:
         raise ValueError("Empty dataset detected. Clinical data cannot be extracted")
-    df_subj_session.drop(
-        df_subj_session[df_subj_session.session_id == "sc"].index, inplace=True
-    )
     write_adni_sessions_tsv(df_subj_session, bids_subjs_paths)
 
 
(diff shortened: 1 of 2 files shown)


clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -718,11 +718,11 @@ def create_adni_sessions_dict(
                 ]
                 df_subj_session = pd.concat([df_subj_session, df_filtered], axis=1)
 
+    df_subj_session = df_subj_session[
+        (~df_subj_session.session_id.isin(["ses-Nv", "sc", False]))
+    ]
```
ghisvail, line 723: Do you know where these anomalous session IDs are coming from?

ghisvail: `False` sounds particularly wrong to me in this context, and might be a sign there is another issue higher up the processing flow. What do you think?

omar-rifai: False is just a flag I used to signal there isn't data available for the sessions. For the others, "Nv" is the same flag but when ADNI detects it and "sc" refers to the screening session. Which was converted to "bl" in the previous code but from what I understood they are not equivalent.

ghisvail: > False is just a flag I used to signal there isn't data available for the sessions.

Oh I see. Maybe substituting `False` for `None` would make more sense to signal the absence of data? On first read it was not very obvious tbh.

> For the others, "Nv" is the same flag but when ADNI detects it and "sc" refers to the screening session. Which was converted to "bl" in the previous code but from what I understood they are not equivalent.

Ok now it makes sense 👍 

In the interest of future maintenance, I think it'd be worth adding a comment to record this piece of context which helps understand why this filtering step is here.

omar-rifai: done

ghisvail (approved):

PR #414, opened 2021-09-13
Fix unexpected argument in non-regression tests

ghisvail:
`shared_folder_name` no longer exists


mdiazmel (approved):
Thx! 

The arguments of the  function `compare_folders` changed in a previous PR. There is a big refactoring waiting for the non-regression tests in the pipelines. 

I'll merge this PR.

PR #417, opened 2021-09-14
NIFD: Fix error reporting when no imaging data found

ghisvail:
DataFrame.from_records fails with a TypeError if the source iterator
yields no value instead of creating an empty dataframe. This change
replaces the useless empty dataframe check with an appropiate
try/catch block.


MatthieuJoulot:
Works as it is supposed to

mdiazmel (approved):
Seems ok to me,

I'll merge!

PR #429, opened 2021-09-20
Activate oasis3 ci

MatthieuJoulot:
Activate OASIS3ToBids converter in the non-regression test.
Change a few variable names for clarity.

### DIFF
--- a/clinica/iotools/converters/oasis3_to_bids/oasis3_utils.py
+++ b/clinica/iotools/converters/oasis3_to_bids/oasis3_utils.py
@@ -68,11 +68,29 @@ def read_clinical_data(
 def read_imaging_data(imaging_data_directory: PathLike) -> DataFrame:
     import pandas as pd
 
-    s = pd.Series(find_imaging_data(imaging_data_directory), name="source_path")
+    source_path_series = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
 
-    b = s.str.rsplit(pat="/", n=1, expand=True).drop(columns=1)[0].rename("source_dir")
-    a = s.str.split(pat="/", n=1, expand=True).drop(columns=1)[0].rename("path")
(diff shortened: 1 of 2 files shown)


clinica/iotools/converters/oasis3_to_bids/oasis3_utils.py
```diff
@@ -68,11 +68,29 @@ def read_clinical_data(
...
 
-    s = pd.Series(find_imaging_data(imaging_data_directory), name="source_path")
+    source_path_series = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
 
-    b = s.str.rsplit(pat="/", n=1, expand=True).drop(columns=1)[0].rename("source_dir")
-    a = s.str.split(pat="/", n=1, expand=True).drop(columns=1)[0].rename("path")
-    df_source = pd.concat([s, a, b, a.str.split("_", expand=True)], axis=1)
+    source_dir_series = (
```
omar-rifai, line 75: is that a directory (dir) or a file?

clinica/iotools/converters/oasis3_to_bids/oasis3_utils.py
```diff
@@ -68,11 +68,29 @@ def read_clinical_data(
...
+        source_path_series.str.rsplit(pat="/", n=1, expand=True)
```
omar-rifai, line 76: Can you look into using `PurePath`from pathlib for cross-platform compatibility?

omar-rifai (changes requested):

clinica/iotools/converters/oasis3_to_bids/oasis3_utils.py
ghisvail: Or use `Path.parent` / `Path.name` instead of splitting. If you do need splitting, then it's probably best to use `os.sep` in the interest of cross-platform agnosticity indeed.

omar-rifai (approved):

PR #392, opened 2021-08-23
Add converter for HABS

ghisvail:
This converter is in an experimental stage, i.e. it performs conversion properly on a limited number of modalities (anatomical MRI and PET currently) but its implementation may still be subject to change.

PR #430, opened 2021-09-22
Write dataset description for all converters

ghisvail:
Dataset description files are a requirement for BIDS compliance.

- Define a proper model for the BIDS dataset description
- Ensure the model is used in bids utility functions
- Add missing dataset description for some converters

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -27,7 +27,7 @@ class BidsDatasetDescription(BaseModel):
     dataset_type: DatasetType = DatasetType.raw
 
     def write(self, to: TextIOBase) -> None:
-        to.write(self.json(by_alias=True))
+        to.write(self.json(by_alias=True, indent=4))
 
 
 # -- Methods for the clinical data --
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -1,6 +1,33 @@
 """Methods used by BIDS converters."""
-
+from enum import Enum
+from io import TextIOBase
 from typing import List
+
+from pydantic import BaseModel
+
+
+class BidsDatasetDescription(BaseModel):
+    """Model class for a BIDS description."""
+
+    class Config:
+        fields = {"bids_version": "BIDSVersion"}
+        allow_population_by_field_name = True
+
+        @classmethod
+        def alias_generator(cls, string: str) -> str:
+            """Generate an alias as PascalCase."""
+            return "".join(word.capitalize() for word in string.split("_"))
+
+    class DatasetType(Enum):
+        raw = "raw"
+        derivative = "derivative"
+
+    name: str
+    bids_version: str
+    dataset_type: DatasetType = DatasetType.raw
+
+    def write(self, to: TextIOBase) -> None:
+        to.write(self.json(by_alias=True))
 
 
 # -- Methods for the clinical data --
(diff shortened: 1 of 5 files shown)


ghisvail:
Note that the current definition of the BIDS dataset description model is the most nominal one based on the current specs. But it's implemented in a way that future extension only requires adding a properly typed attribute to the class and a field-to-alias mapping in `Config.fields` if the field is not serialized to pascal case.

clinica/iotools/bids_utils.py
```diff
@@ -1,9 +1,36 @@
...
+
+class BidsDatasetDescription(BaseModel):
+    """Model class for a BIDS description."""
+
+    class Config:
+        fields = {"bids_version": "BIDSVersion"}
+        allow_population_by_field_name = True
+
+        @classmethod
+        def alias_generator(cls, string: str) -> str:
```
omar-rifai, line 19: Please replace string with a more context-relevant name

clinica/iotools/bids_utils.py
```diff
@@ -1,9 +1,36 @@
...
+
+
+class BidsDatasetDescription(BaseModel):
+    """Model class for a BIDS description."""
+
+    class Config:
+        fields = {"bids_version": "BIDSVersion"}
+        allow_population_by_field_name = True
+
+        @classmethod
```
omar-rifai, line 18: Shouldn't this be a static method?

requirements.txt
```diff
@@ -24,3 +24,4 @@ click >= 7
 click-option-group >= 0.5, < 0.6
 openpyxl
 fsspec
+pydantic>=1.4
```
omar-rifai, line 27: In the effort of moving towards typing and making the code more robust, the use of pydantic seems to be relevant.  But please concert before introducing new dependencies to the requirements.

clinica/iotools/bids_utils.py
```diff
@@ -1,9 +1,36 @@
...
 from typing import List
 
+from pydantic import BaseModel
+
+
+class BidsDatasetDescription(BaseModel):
+    """Model class for a BIDS description."""
+
+    class Config:
+        fields = {"bids_version": "BIDSVersion"}
```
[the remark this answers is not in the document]
ghisvail: Fixed in #392

ghisvail:
Blocked by #392, which implements a dataset description model for HABS that could be used for the other converters targeted in this PR. Depends which one we want to merge first between #392 and this one.

ghisvail:
We still need to update the CI data to include the missing `dataset_description.json` file for some converters. To be done after the current CI refactoring and testing efforts.

ghisvail:
Before we consider merging this, I'd like to test the same approach with `attrs` (an alternative to `pydantic`), which is used by `pydra` extensively. If we were to consider switching to `pydra` at some point, we might benefit from using the same dependencies.

ghisvail:
I believe the alternative solution in #535 using `attrs` is better from a conceptual and maintenance perspective. No need to learn the intricacies of `pydantic` (especially the `Config` inner class) and closer to the composition over inheritence principle where Python has been going lately. The `attrs` solution is basically a `dataclass` (record type) composed with a converter registry provided by `cattrs`. Simple.

omar-rifai:
Should we close this PR?

ghisvail:
It will be closed automatically once #535 is merged

PR #431, opened 2021-09-23
[WIP] Create adni json from xml metadata 

omar-rifai:
Module to allow integration of xml metadata to the adni-to-bids clinical data converter


omar-rifai:
Hello @emaheux. Please let me know if this would satisfy the requirements for leaspy.
The few modifications I made over your code is:
- Running the xml metadata extractor when the argument --xml_path is given
- Removing the use of the multiprocessing library (in the objective of switching to a multithreading/asyncio library).

omar-rifai:
I have kept the resulting file with all the data extracted separate from the BIDs datastructure but could have the `write` function add the data next to the images if necessary.

emaheux:
Thank you Omar :)

> I have kept the resulting file with all the data extracted separate from the BIDs datastructure but could have the `write` function add the data next to the images if necessary.

It would have been nice to have this metadata as part of the BIDS structure (metadata of scans tsv? - or at least in merged tsv BIDS.tsv?) but if it is not possible / too complicated it will be fair enough to load this extra json file!

omar-rifai:
> Thank you Omar :)
> 
> > I have kept the resulting file with all the data extracted separate from the BIDs datastructure but could have the `write` function add the data next to the images if necessary.
> 
> It would have been nice to have this metadata as part of the BIDS structure (metadata of scans tsv? - or at least in merged tsv BIDS.tsv?) but if it is not possible / too complicated it will be fair enough to load this extra json file!

No it should be doable. I was just wondering what would be best. I'll keep you posted. 
(for merge tsv it should appear automatically if added to the scan tsv file).

omar-rifai:
Closed as no consensus found currently as to where this data should ideally be integrated. `Merge-tsv` should also see a refactoring soon (c.f [issue 559](https://github.com/aramis-lab/clinica/issues/559)) which goes against integrating the data there.

omar-rifai:
Hi @emaheux, since Nicolas is working on similar topics, we were wondering if this is still relevant. I did not have the time to pursue the task but it can be an opportunity to pick up the work. Let us know.

emaheux:
Indeed @omar-rifai, this is still relevant (though obviously not blocking / top-priority  since we circumvented the issue long time ago by adding ADNI imaging metadata downstream by ourselves with the codebase I gave you). 

Yet if a _Leaspyer_ needs to conduct experiments with a newer version of ADNI, I think it would be really nice / more consistent to have those imaging metadata directly accessible in Clinica outputs!

PR #440, opened 2021-09-24
Remove strict versioning in dependencies

ghisvail:


mdiazmel (commented):
Cool!
I'll try to make a run on the non-regression test to see if there are any problems with this update before approving/merging.

ghisvail:
Let's focus on #441 since we'll get this for free.

PR #441, opened 2021-09-24
Use poetry for dependency management

ghisvail:
Work in progress.


ghisvail:
We could commit the lock file at a later stage as explained in this [section of the documentation](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control).

mdiazmel:
I quickly saw the `UnicodeEncodeError` in the ["Build and publish" stage at the CI](https://ci.inria.fr/clinica-aramis/blue/organizations/jenkins/clinica/detail/PR-441/18/pipeline). It's really strange, it happens in different VM every time (it's not related to a single machine). I would say that it can be related to the dependencies used to build the docs (and how poetry grab them). To investigate...

ghisvail:
> I would say that it can be related to the dependencies used to build the docs

It works fine on my Macbook and my Linux machine at home.

Could you confirm `LC_ALL=C.UTF-8` (or any UTF-8 locale) is set for these VMs?

mdiazmel:
> Could you confirm `LC_ALL=C.UTF-8` (or any UTF-8 locale) is set for these VMs?

Yes, that's the case! This is the output of `locale` for all of them:

```
LANG=C.UTF-8
LANGUAGE=en_US:
LC_CTYPE="C.UTF-8"
LC_NUMERIC="C.UTF-8"
LC_TIME="C.UTF-8"
LC_COLLATE="C.UTF-8"
LC_MONETARY="C.UTF-8"
LC_MESSAGES="C.UTF-8"
LC_PAPER="C.UTF-8"
LC_NAME="C.UTF-8"
LC_ADDRESS="C.UTF-8"
LC_TELEPHONE="C.UTF-8"
LC_MEASUREMENT="C.UTF-8"
LC_IDENTIFICATION="C.UTF-8"
LC_ALL=C.UTF-8
```

mdiazmel:
Since the "build" and the "instantiation tests"  stages has been removed in the mac VMs, I propose to uncomment (temporary) the lines corresponding to these stages in this PR (cf. the Jenkinsfile), in order to test how `poetry` deals with the dependencies in Mac.

ghisvail:
> in order to test how poetry deals with the dependencies in Mac.

Capture d’écran 2021-09-28 à 17 09 38

ghisvail:
Victory:

Capture d’écran 2021-09-29 à 11 49 26

ghisvail:
Let's try with the lock file now

ghisvail:
With the lock file:

Capture d’écran 2021-09-29 à 11 49 26

pyproject.toml
```diff
@@ -1,11 +1,83 @@
...
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-ensure_newline_before_comments = true
-line_length = 88
-skip_gitignore = true
+[tool.poetry]
+name = "clinica"
+version = "0.5.1"
```
mdiazmel, line 3: If version will be set here, it seems that there is no reason to keep `clinica/VERSION` (to verify).

pyproject.toml
```diff
@@ -1,11 +1,83 @@
...
+matplotlib = "*"
+niflow-nipype1-workflows = "*"
+scikit-image = "^0.16.2"
+torch = "^1.0"
+torchvision = "*"
+pydicom = "*"
+networkx = "*"
+click = "^7"
+click-option-group = "^0.5"
+xlrd = ">= 1.2"
```
mdiazmel, line 52: It shouldn't be "safer" to use the caret requirement here (like for most of the other requirements) 
```suggestion
xlrd = "^1.2"
```

pyproject.toml
```diff
@@ -1,11 +1,83 @@
...
+openpyxl = ">= 2.6"
```
mdiazmel, line 53: Same that before.
```suggestion
openpyxl = "^2.6"
```

docs/Installation.md
```diff
@@ -132,26 +132,19 @@ conda deactivate
...
-If you plan to contribute to Clinica or want to test the current development version:
-
-- Clone the Git repository
-
-    ```sh
-    git clone https://github.com/aramis-lab/clinica.git
-    cd clinica
-    git checkout dev
-    ```
+This section is intended for users who plan to contribute to Clinica or test the current development version.
```
mdiazmel, line 135: I would add here the instructions to create a Conda environment by using the `environment.yml` file and then install Clinica "dev" with poetry inside. Btw, what's the behavior of `poetry shell` triggered from inside a conda env?

mdiazmel (commented):
Hi @ghisvail ,

Thanks for this nice contribution. Globally seems ok, there are some comments inside the code. 

Need to test the package creation and how it performs when running the "execution" tests for all the pipelines (but this should be done certainly in another PR).

Jenkinsfile
```diff
@@ -24,18 +27,20 @@ pipeline {
                  eval "$(conda shell.bash hook)"
                  conda env create --force --file environment.yml -n clinica_env_${BRANCH_NAME}
                  conda activate clinica_env_$BRANCH_NAME
-                 pip install -r requirements-dev.txt
+                 LC_ALL=C.UTF-8 poetry install --no-root
```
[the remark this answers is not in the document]
ghisvail: Agreed, I'll try to remove them again and check.

pyproject.toml
```diff
@@ -1,11 +1,83 @@
...
+fsspec = "*"
+# docs
```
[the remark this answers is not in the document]
ghisvail: if you do `poetry install` the docs dependency are not installed

if you do `poetry install --extras docs`, all dependencies will be installed including the docs

they the equivalent of `extras_require` in setuptools

poetry 1.2 will introduce *dependency groups* which will be a bit more clever, but for now that's all we've got

pyproject.toml
ghisvail: indeed, topic for a follow-up PR

pyproject.toml
ghisvail: Actually the answer is neither. They should be left as `*` and let `pandas` set the versioned requirements transitively. Also I have read reports of `xlrd` being subject to security vulnerabilites (because XML parsing, classic) so keeping it a v1.x is not advisable.

pyproject.toml
ghisvail: Same rationales (security), to be set transitively by Pandas.

docs/Installation.md
ghisvail: > I would add here the instructions to create a Conda environment by using the environment.yml file and then install Clinica "dev" with poetry inside.

Good point. Although technically, any Python 3.7 provider (system, pyenv, homebrew) would work too with these instructions.

> `Btw, what's the behavior of poetry shell triggered from inside a conda env?`

Passthrough.

ghisvail: ^ precisely, you get:

```console
$ poetry shell
Virtual environment already activated: /path/to/env
```

Jenkinsfile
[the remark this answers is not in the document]
ghisvail: It works.

pyproject.toml
omar-rifai: How is this version number updated?

clinica/iotools/converters/nifd_to_bids/nifd_utils.py
```diff
@@ -26,7 +26,7 @@ def find_clinical_data(clinical_data_directory: PathLike) -> Optional[DataFrame]
 
 
 def read_clinical_data(clinical_data_directory: PathLike) -> DataFrame:
-    from pandas import NA, CategoricalDtype
+    import pandas as pd
```
omar-rifai, line 29: Why are these changes in this PR?

omar-rifai (approved):
Looks great, only some code that needs to be in a separate PR.

clinica/iotools/converters/nifd_to_bids/nifd_utils.py
ghisvail: Now that we are using pandas v1.2, it could also be simplified with `dtype="Int64"` and `dtype="Float64"` (notice the capitalization to distinguish between Pandas and Numpy).

ghisvail:
Another reason to get this merged quickly: https://github.com/aramis-lab/clinica/issues/458

Python 3.6 has to go :-)

PR #443, opened 2021-09-28
Open TSV files in text mode in NIFD and OASIS-3

ghisvail:
Binary write mode was only introduced in Pandas version 1.2. Use text mode instead for compatibility with Python 3.6 (Pandas version 1.1.x).


ghisvail:
Another find from `poetry`

PR #448, opened 2021-09-30
Extarct working_dir parameter from cmdopt

omar-rifai:
Bugfix to adapt pytest to code update in PR[404](https://github.com/aramis-lab/clinica/pull/404)


mdiazmel:
Hi @omar-rifai ,

I'm working in a refactor of the non-regression tests for all the pipelines (as done for the converters/utils). This modification will be included there.

PR #461, opened 2021-10-11
Fix missing non-interactive mode in pipelines

ghisvail:
It was discovered internally that some pipelines did not have an option to turn interactivity off, resulting in undesired prompts when processing on a cluster environment. This PR adds the missing `yes` option to these pipelines and propagate it to `skip_question`.


omar-rifai:
Thanks looks good. For future referece, might be worthwhile leaving a small comment explaining the purpose of the PR.

PR #469, opened 2021-10-14
Fix dwi-preprocessing CLI (initrand and use_cuda should be bool)

alexandreroutier:


ghisvail:
Indeed, thanks for the fix.

Issue #478, opened 2021-10-16
Error with !clinica iotools merge-tsv d1bitOut  d1capout/participants.tsv


SonaLily:
Hello! I'm very grateful for all the help everyone has given me so far. I thought it would be good to open a new issue for this bug, as the old thread has been closed.
I have been able to successfully run : 
!clinica run t1-linear d1BIDSOut  d1CAPSout --n_procs  4
Then I continued to run:
!clinica iotools merge-tsv d1bitOut  d1capout/participants.tsv

And got following error message:
Traceback (most recent call last):
  File "/usr/local/bin/clinica", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/dist-packages/clinica/cmdline.py", line 78, in main
    cli()
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/clinica/iotools/utils/cli.py", line 255, in merge_tsv
    tracers_selection=pet_tracers_selection,
  File "/usr/local/lib/python3.7/dist-packages/clinica/iotools/utils/data_handling.py", line 160, in create_merge_file
    merged_df = merged_df.append(row_df)
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/frame.py", line 7751, in append
    sort=sort,
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/reshape/concat.py", line 287, in concat
    return op.get_result()
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/reshape/concat.py", line 503, in get_result
[...]

PR #486, opened 2021-10-21
Remove duplicate columns if found

omar-rifai:
Following [Issue 478](https://github.com/aramis-lab/clinica/issues/478), `merge-tsv` is not currently robust to duplicate columns in the tsv files. 

A quick fix is to drop duplicates if found. Open to other suggestions.

### DIFF
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -157,7 +157,7 @@ def create_merge_file(
                 )
 
                 # remove duplicated columns
-                row_df = row_df.loc[:, ~row_df.columns.duplicated()]
+                row_df = row_df.loc[:, ~row_df.columns.duplicated(keep="last")]
 
                 merged_df = merged_df.append(row_df)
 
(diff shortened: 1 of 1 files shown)


ghisvail:
According to the BIDS spec, you may have duplicate attributes following the inheritance principle.

For instance, `age` in `participants.tsv` may qualify the age of the participant at recruitment for the study, whilst `age` in `sessions.tsv` will qualify his age at the time of the imaging session.

Therefore, in case of attribute duplication, I believe we should favor the one that has been overridden in the sessions file.

clinica/iotools/utils/data_handling.py
```diff
@@ -155,6 +155,10 @@ def create_merge_file(
                 row_df = pd.concat(
                     [row_participant_df, row_session_df, row_scans_df], axis=1
                 )
+
+                # remove duplicated columns
+                row_df = row_df.loc[:, ~row_df.columns.duplicated()]
```
ghisvail, line 160:
```suggestion
                row_df = row_df.loc[:, ~row_df.columns.duplicated(keep="last")]
```

PR #501, opened 2021-11-04
Change the tested directory for StatisticsVolumeCorrection

MatthieuJoulot:
Change the tested directory for StatisticsVolumeCorrection, to test only the outputs and not the subjects part copied from the caps. 

The subjects directory in the outputs is a copy of the the subjects directory in the inputs. Hence, it should not be tested as it is not "generated" by the pipeline.


pyproject.toml
```diff
@@ -40,7 +40,7 @@ nilearn = "^0.6"
 colorlog = "^5"
 xgboost = "*"
 scipy = "^1.2.3"
-matplotlib = "*"
+matplotlib = "3.3.*"
```
ghisvail, line 43: Why does the matplotlib version need to be constrained to `3.3.x`?

MatthieuJoulot: Because it is apparently causing the StatisticsVolumeCorrection pipeline to crash. 
It is suggested in this link: https://stackoverflow.com/questions/67969047/error-trying-to-re-register-the-builtin-cmap-cubehelix-when-trying-to-import
I have also confirmed that the pipeline works with a version inferior to 3.4.0 (3.3.4 to be specific) !

ghisvail: Are we doing this colormap registration ourselves or is it done transitively by one of our dependencies (like `yt` in the link)?

MatthieuJoulot: It is done by one of our dependencies

ghisvail: Could you tell me which one please? Or give me a link to a stack trace with the error 🙏

MatthieuJoulot: I just sent you the relevant stack traces !

ghisvail: Ok, it's `nilearn`

ghisvail: Fixed since [nilearn 0.7.0](https://github.com/nilearn/nilearn/commit/0324d59a134b2fbc7c1372b2c20c117e9214ba08).

So the right solution would be to raise the minimum version of `nilearn` to 0.7.0 and leave `matplotlib` unconstrained.

ghisvail (approved):

mdiazmel (commented):
Looks good !

Just, I don't understand with the lock file has changed ? The update of nilearn is not related to this PR, non ?

omar-rifai:
@MatthieuJoulot , can you please handle @mdiazmel's comment?

MatthieuJoulot:
> Just, I don't understand with the lock file has changed ? The update of nilearn is not related to this PR, non ?

The pipeline StatisticVolumeCorrection crashed, because of a call to matplotlib done by nilearn. Since nilearn fixed this bug in the version 0.7.0, we have updated it.

omar-rifai (approved):

mdiazmel:
LGTM,

I'll merge this one!

PR #504, opened 2021-11-18
Fix typo in StatisticsSurface cli

omar-rifai:
the command line argument `covariate` was spelled without an `s` but used as `covariates` in the call to cli()


omar-rifai:
Also fix the use of the variable as a list instead of a string in the non-regression tests

mdiazmel (commented):
Seems ok,

It can be merged quickly.

Issue #491, opened 2021-10-25
Issue with t1-volume / t1-volume-existing-template CLI


alexandreroutier:
Hi!

I wanted to reproduce the bug described in the forum thread: [t1-volume-existing-template issue](https://groups.google.com/g/clinica-user/c/ySXFipWXYMI).

Using data CI `T1VolumeExistingDartel`, I had a different issue on my side:
```
❯ clinica run t1-volume-existing-template bids caps TestGroup -np 2
Traceback (most recent call last):
  File "<Miniconda>/envs/ClinicaDev/bin/clinica", line 5, in <module>
    main()
  File "/Users/alexandre.routier/Projects/Git/GitHub/alexandreroutier/clinica/clinica/cmdline.py", line 76, in main
    cli()
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 1257, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 700, in make_context
    self.parse_args(ctx, args)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 1048, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click_option_group/_core.py", line 52, in handle_parse_result
    return super().handle_parse_result(ctx, opts, args)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 1623, in handle_parse_result
    value = self.full_process_value(ctx, value)
  File "<Miniconda>/envs/ClinicaDev/lib/python3.7/site-packages/click/core.py", line 1965, in full_process_value
    return Parameter.full_process_value(self, ctx, value)
[...]

PR #506, opened 2021-11-22
Update changelog and version for release 0.5.3

omar-rifai:
Update `changelog.md` and release version in `pyproject.toml` in preparation for release v0.5.3


ghisvail:
I believe we are missing a version bump in the `clinica/VERSION` file.

Not sure whether it still matters though 🤔

mdiazmel:
> I believe we are missing a version bump in the clinica/VERSION file.

> Not sure whether it still matters though 🤔

Currently it doesn't matters because after the switch to `click`, `clinica --version` reads the version information from the `pyproject.toml` file. It should be nice if this field can be recovered from the `clinica/VERSION` file. I think that this will solve the problem that I discussed in the MM chat last week (dev installation with poetry showing the same version after switching the commit in the repository).

alexandreroutier:
Hi all,

I don't know what the 0.5.3 version will contain but it could be worth to merge PR #505 before its release. Hopefully, this would be the last PR that will fix the CLI. :)

Btw, you should mention that  the CLI for `dwi-connectome` (issue due to `--n_track` #466) and  `t1-volume*` / `pet-volume` pipelines (issue due to `--smooth` #491) were fixed.

Best,
Alex

ghisvail:
> clinica --version reads the version information from the pyproject.toml file

Technically, it uses the new [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#distribution-versions) module which is now the [recommended way](https://www.python.org/dev/peps/pep-0396/#toc-entry-2) to fetch version information from now on.

mdiazmel:
> Technically, it uses the new importlib.metadata module which is the recommended way to fetch version information from now on.

That's the reason why I propose to retrieve the version information in the `pyproject.toml` file (the package metadata) from the `clinica/VERSION` file. I would prefer not to touch this file (`pyproject.toml`) systematically at each release. If somebody knows another way to do it, go ahead. But don't bother with this, it's only a detail, I don't want to block the PR.

ghisvail:
> > Technically, it uses the new importlib.metadata module which is the recommended way to fetch version information from now on.
> 
> That's the reason why I propose to retrieve the version information in the `pyproject.toml` file (the package metadata) from the `clinica/VERSION` file. I would prefer not to touch this file (`pyproject.toml`) systematically at each release. If somebody knows another way to do it, go ahead. But don't bother with this, it's only a detail, I don't want to block the PR.

Moved to https://github.com/aramis-lab/clinica/pull/508

PR #507, opened 2021-11-22
Split Clinica and Nipype logging

### LINKED ISSUE #363: Default cprint messages are not displayed
**Describe the bug**
Default `cprint` callls are not displayed unless the `--verbose` is used:

t1-freesurfer-no-verbose

t1-freesurfer-verbose

This has now changed. I would think that `INFO` level should be the displayed

**Expected behavior**
See for instance screenshots in https://github.com/aramis-lab/clinica/pull/58

ghisvail:
Nipype's logging can become quite verbose and is only relevant under
abnormal circumstances. Moving its output to a separate file clears the
Clinica console output from unwanted nipype information regardless of
the desired verbosity, whilst providing the necessary debug information
for support requests.

This commit isolates Nipype's logging from Clinica's. Nipype now works
in debug mode and logs its output to a file instead of the console
previously. The verbosity option for Clinica has also been simplified to
two values (INFO and DEBUG) with INFO kept as default.

Closes #363

### DIFF
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -32,6 +32,7 @@ def setup_logging(verbose: bool = False) -> None:
     """
     import logging
     import sys
+    from logging.handlers import RotatingFileHandler as RFHandler
 
     import nipype
     from colorlog import ColoredFormatter, StreamHandler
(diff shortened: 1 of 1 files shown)


ghisvail:
c.f. #457 for additional context

omar-rifai (approved):

PR #516, opened 2021-11-25
Stop on first fail during regression testing

ghisvail:
Considering the duration of the regression tests, perhaps it would be worth failing on first error?

For instance [here] the PET pipelines failed on MacOS early, while the Linux are still going a few hours later.

It's just a proposal, perhaps the current behavior is what we want.


ghisvail:
After discussion, we do want the tests to keep running even if they fail on one architecture.

PR #524, opened 2021-12-09
Enable FreeSurfer to run with missing files

MatthieuJoulot:
Clinica's file reader current version doesn't allow to visualize missing files (*T1w.nii.gz) and to run the pipeline at the same time.
To enable this and give more flexibility, we return the error message containing the missing files and the results together. the message will be empty if no files are missing.

### DIFF
--- a/clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
+++ b/clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
@@ -137,15 +137,14 @@ class T1FreeSurfer(cpe.Pipeline):
         # T1w file:
         try:
             t1w_files = clinica_file_reader(
-                self.subjects, self.sessions, self.bids_directory, T1W_NII
+                self.subjects, self.sessions, self.bids_directory, T1W_NII, False
             )
         except ClinicaException as e:
             err_msg = (
                 "Clinica faced error(s) while trying to read files in your BIDS directory.\n"
                 + str(e)
             )
-            raise ClinicaBIDSError(err_msg)
-
+            print(ClinicaBIDSError(err_msg))
         # Save subjects to process in <WD>/<Pipeline.name>/participants.tsv
         folder_participants_tsv = os.path.join(self.base_dir, self.name)
         save_participants_sessions(
--- a/clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
+++ b/clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
@@ -144,9 +144,8 @@ class T1FreeSurfer(cpe.Pipeline):
                 "Clinica faced error(s) while trying to read files in your BIDS directory.\n"
                 + str(e)
             )
-        print("t1w_files_pipeline:  ", t1w_files)
         if error_message != "":
-            print(error_message)
+            cprint(error_message, lvl="warning")
         if t1w_files == []:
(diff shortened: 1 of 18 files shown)


clinica/utils/inputs.py
```diff
@@ -329,7 +334,20 @@ def clinica_file_reader(
...
+        )
+        if "needed_pipeline" in information.keys():
+            if information["needed_pipeline"]:
+                error_message += (
+                    "Please note that the following clinica pipeline(s) must "
+                    f"have run to obtain these files: {information['needed_pipeline']}\n"
+                )
+        for msg in error_encountered:
+            error_message += msg
+    return results, error_message
```
ghisvail, line 350: There is a large potential for refactoring with the `raise_exception` case. Basically the error message construction code is the same, only what is done with it (raise in an exception or return the error message as a result) actually changes.

ghisvail: That would make the code clearer : one section building the error message, one section returning or raising based on the input arguments.

ghisvail (changes requested):

clinica/utils/inputs.py
```diff
@@ -312,24 +317,24 @@ def clinica_file_reader(
...
+            )
         for msg in error_encountered:
             error_message += msg
+    if len(error_encountered) > 0 and raise_exception is True:
         if is_bids:
             raise ClinicaBIDSError(error_message)
         else:
             raise ClinicaCAPSError(error_message)
-    return results
+    return results, error_message
```
omar-rifai, line 337: An alternative to changing the return type (and hence updating all the calls to the function), is to add a new parameter variable (e.g `print_errors=False`) which prints the error messages if activated. I'm not saying we should go ahead with this solution, but maybe to keep in mind.

omar-rifai:
Currently if the first file of a tsv is missing, the pipeline seems to skip the file and execute the second subject with no warning message. I did not wait for it to finish execution but I was wondering whether that is the expected behavior?

omar-rifai:
@ghisvail , is all good with the required modification? If so we can merge.

omar-rifai (approved):

clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
```diff
@@ -136,15 +136,18 @@ def build_input_node(self):
...
+            t1w_files, error_message = clinica_file_reader(
+                self.subjects, self.sessions, self.bids_directory, T1W_NII, False
             )
         except ClinicaException as e:
             err_msg = (
                 "Clinica faced error(s) while trying to read files in your BIDS directory.\n"
                 + str(e)
             )
-            raise ClinicaBIDSError(err_msg)
+        if error_message != "":
```
ghisvail, line 147:
```suggestion
        if error_message:
```

clinica/pipelines/t1_freesurfer/t1_freesurfer_pipeline.py
```diff
@@ -136,15 +136,18 @@ def build_input_node(self):
...
+            cprint(error_message, lvl="warning")
+        if t1w_files == []:
```
ghisvail, line 149:
```suggestion
        if not t1w_files:
```

ghisvail (changes requested):
Have a look of the suggestions, the rest looks good to me 👍

ghisvail (approved):

PR #528, opened 2021-12-14
Use change ID instead of branch name for docs path

### LINKED ISSUE #526: CI fails if branch name contains slashes
Dependabot submits pull requests for automatic security updates to our dependencies. The branch naming scheme for those PR follows a namespace convention, i.e. starts with `dependabot/`, and includes slashes. The latter don't play well with the section of the CI building the documentation which uses `${BRANCH_NAME}` as a path and yields error like:

`mv: cannot move 'site' to 'dependabot/pip/lxml-4.6.5': No such file or directory`

This CI section needs to be more robust towards any branch naming scheme.

ghisvail:
This PR makes documentation generation by the CI more robust to
arbitrary branch naming schemes, like those containing slashes.

Closes #526


mdiazmel (approved):
LGTM, can be merged.

PR #532, opened 2021-12-14
Normalize subprocess calls to dcm2niix

### LINKED ISSUE #527: Allow paths with space characters in subprocess calls
A user [reported](https://groups.google.com/g/clinica-user/c/FbYaIX6tosg) encountering an error running `adni-to-bids`  with space characters in the path. This is due to `dcm2niix` reading the path as two separate command line arguments.

Clinica should be robust to this use case or at least have a more explicit warning message.

ghisvail:
All calls to dcm2niix now transit through the `run_dcm2niix` helper
function. Do not use shell=True to allow proper escaping of white
spacing in path arguments.

Closes #527

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -761,7 +761,13 @@ def json_from_dcm(dcm_dir, json_path):
         cprint(msg=f"No DICOM found at {dcm_dir}", lvl="warning")
 
 
-def run_dcm2niix(command):
+def run_dcm2niix(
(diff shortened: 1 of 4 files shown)


omar-rifai (approved):
Looks good ! From what I understand `shell=True` is for running the command through the shell (and also changes the way parameters are parsed). Is that correct? Won't there be issues with finding the tool `dcm2niix` if not ran through the shell?  The non-regression tests seem to confirm this works so I would suppose it's OK

PR #533, opened 2021-12-14
Update the cronjob to run every friday at midnight

omar-rifai:


ghisvail (approved):

mdiazmel:
@omar-rifai , @ghisvail ,

I would prefer to start the cronjob earlier, if anything goes wrong we will be able to stop it (or fix it) before it runs all the night.  

Besides this, I think that we can split these tests to run in different machines (now that we have multiple agents available). What do you think?

omar-rifai:
@mdiazmel  Sounds good. I think the point here was to specifically specify the day of the week in order to avoid it running inadvertently if it fails during the weekend.

mdiazmel:
> @mdiazmel Sounds good. I think the point here was to specifically specify the day of the week in order to avoid it running inadvertently if it fails during the weekend.

Ahh ok. I think that having `5` \`FRY` or `6`\`SAT` doesn't change anything ([e.g. here](https://crontab.guru/#*_*_*_*_6)). In fact, I think that the CI was triggering because a bad configuration of the Multiproject pipeline in the Jenkins interface (it used to trigger the CI when the `dev` branch was updated). I changed this behavior.

PR #535, opened 2021-12-20
Add model for BIDS dataset description

ghisvail:
This implementation uses attrs for model definition and cattrs for
serialization.

Closes #430

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -435,19 +435,16 @@ def write_modality_agnostic_files(study_name, bids_dir):
     from os import path
 
     import clinica
+    from clinica.iotools.bids_dataset_description import BIDSDatasetDescription
 
-    dataset_dict = {"Name": study_name, "BIDSVersion": "1.4.1", "DatasetType": "raw"}
-    dataset_json = json.dumps(dataset_dict, skipkeys=True, indent=4)
-    f = open(path.join(bids_dir, "dataset_description.json"), "w")
-    f.write(dataset_json)
-    f.close()
+    with open(path.join(bids_dir, "dataset_description.json"), "w") as f:
+        BIDSDatasetDescription(name=study_name, bids_version="1.4.1").write(to=f)
 
-    file = open(path.join(bids_dir, "README"), "w")
-    file.write(
-        f"This BIDS directory was generated with Clinica v{clinica.__version__}.\n"
-        f"More information on http://www.clinica.run\n"
-    )
-    file.close()
+    with open(path.join(bids_dir, "README"), "w") as f:
+        f.write(
+            f"This BIDS directory was generated with Clinica v{clinica.__version__}.\n"
+            f"More information on http://www.clinica.run\n"
+        )
 
     validator_dict = {
         "ignore": [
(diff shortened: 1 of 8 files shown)


omar-rifai:
Did this mess up something with the instantiation tests? It seems none have passed.

clinica/iotools/bids_utils.py
```diff
@@ -435,19 +435,16 @@ def write_modality_agnostic_files(study_name, bids_dir):
...
     import clinica
+    from clinica.iotools.bids_dataset_description import BIDSDatasetDescription
 
-    dataset_dict = {"Name": study_name, "BIDSVersion": "1.4.1", "DatasetType": "raw"}
-    dataset_json = json.dumps(dataset_dict, skipkeys=True, indent=4)
-    f = open(path.join(bids_dir, "dataset_description.json"), "w")
-    f.write(dataset_json)
-    f.close()
+    with open(path.join(bids_dir, "dataset_description.json"), "w") as f:
+        BIDSDatasetDescription(name=study_name, bids_version="1.4.1").write(to=f)
```
omar-rifai, line 441: Why are we using a hardcoded version here while we have a global variable for BIDS?

ghisvail: To remain consistent with the old code. I guess BIDS version `1.4.1`was the standards against which the first converters were coded and no one checked whether they are compatible with the latest standards version. If we are confident they are indeed compliant with the latest version, then we could drop setting it explicitly.

omar-rifai:
Looks good to me. We can wait for @mdiazmel review then merge.

ghisvail:
I removed the commits touching the pipelines. It is ready for review.

clinica/iotools/bids_utils.py
mdiazmel, line 441: Didn't get why here there is `1.4.1` version while in the class definition (`clinica/iotools/bids_dataset_description.py `) there is a variable `BIDS_VERSION = "1.6.0"` ? (Some of our `dataset_description.json` files are `1.4.1` and others `1.6.0`. I'm probably missing something...).

mdiazmel (approved):
Thanks @ghisvail .

Seems ok, can be merged.

clinica/iotools/bids_utils.py
ghisvail: `BIDS_VERSION` should be updated to whatever the current BIDS standards version supported is (1.6.0 as of today). So a converter running today will output a `dataset_description.json` with a BIDS version set to 1.6.0. Recent converters we have worked on (NIFD, OASIS-3, HABS) are compliant with this version.

Now for the oldest ones (AIBL, ADNI, ...) had BIDS standards version set to `1.4.1`, which probably was the latest version _at the time they were implemented_. This PR does not assume their implementation is compliant with `1.6.0`, hence why the initial version of `1.4.1`is left. They are _probably_ compatible with `1.6.0`, but this verification is a topic for a different PR in my opinion.

ghisvail: Most notably, version `1.6.0` introduced the definitive specification for PET imaging, whereas we used to do our own thing back then.

mdiazmel: Ok, I see. Thanks for the clarification!

PR #539, opened 2022-01-03
Automate management of stale issues

ghisvail:
Issues are automatically labelled `stale` after 14 days of inactivity.
Issues are automatically closed after further 14 days of inactivity.

See: https://github.com/actions/stale


mdiazmel:
Don't know where the error on the convertor test come from (cf. CI)? It's not certainly due to this PR...

ghisvail:
> Don't know where the error on the convertor test come from (cf. CI)? It's not certainly due to this PR...

It's the additional `dataset_description.json` file that was added to the OASIS-3 converter for #535.

PR #540, opened 2022-01-03
Do not let nipype check for updates

ghisvail:
This feature leads to unnecessary spamming in the debug logs.

See: https://github.com/nipy/nipype/blob/master/nipype/utils/config.py#L64


mdiazmel (approved):
Cool,
It should avoid unnecessary requests to the nipype server!

PR #548, opened 2022-01-17
Disable DWI tests waiting for fix

omar-rifai:
DWI tests currently fail on weekly non regression tests, preventing other tests from completing. This PR disables these tests until a fix has been found.

### DIFF
--- a/Jenkinsfile_nonregression
+++ b/Jenkinsfile_nonregression
@@ -184,6 +184,42 @@ pipeline {
                 }
               }
             }
+            /* stage('DWI:nonreg') {
+              environment {
+                PATH = "/usr/local/Modules/bin:$PATH"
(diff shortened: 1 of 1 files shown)


ghisvail:
Maybe better to comment them out (with `/* */`) instead of a delete?

mdiazmel (approved):
LGTM, thanks @omar-rifai !

I'll merge it. 

BTW, do we need to create an issue to remember to re-activate these tests once the problem solved?

omar-rifai:
> LGTM, thanks @omar-rifai !
> 
> I'll merge it.
> 
> BTW, do we need to create an issue to remember to re-activate these tests once the problem solved?

Good idea, done in [issue 549].(https://github.com/aramis-lab/clinica/issues/549)

PR #550, opened 2022-01-18
Add a forgotten output to clinica_file_reader

MatthieuJoulot:
When clinica_file_reader is called, if the inputs and outputs are the same (meaning the inputs have already been processed),  the function will exit early with a return.  One output of this return was missing.


omar-rifai:
> An early output of clinica. file reader was forgotten, corresponding to an early output in which case the. input and output are corresponding.

Can you please reformulate this sentence. It doesn't seem clear to me.

omar-rifai (approved):

PR #554, opened 2022-01-22
Run project-wide format before release

ghisvail:
The current state of the code does not lint without errors, meaning code style drifted from the expected config.

This PR runs a project-wide format to normalize the formatting back to a state where it lints without errors.

### DIFF
--- a/clinica/iotools/bids_dataset_description.py
+++ b/clinica/iotools/bids_dataset_description.py
@@ -2,8 +2,8 @@ from enum import Enum
 from typing import IO
 
 import attr
-from cattr.preconf.json import make_converter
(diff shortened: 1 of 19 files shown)


ghisvail:
After inspection of the diff, there is nothing major. Just a bunch of import reordering, line breaks and extra spaces removed.

mdiazmel (approved):
Seems that somebody forgot to activate the isort hook ;)

Thanks @ghisvail , I merge it!

ghisvail:
Indeed. It should probably be enforced by the PR process, though. Pre-commit hooks are an additional setup step that people may forget about.

PR #555, opened 2022-01-22
Update copyright notice in license file

ghisvail:
- Bump year to 2022
- Use HTTPS in URL


omar-rifai:
Thanks.

mdiazmel:
> * Use HTTPS in URL

It should be ideal to point to the https url. But, I think that the server hosting the site web have some problems with the ssl certificate and not sure that it works every time.

ghisvail:
@mdiazmel I believe the issue you are referring is for `clinica.run`?

The ARAMIS website works properly with HTTPS.

PR #556, opened 2022-01-22
Upgrade scikit-learn and scikit-image

### LINKED ISSUE #552: Upgrade scikit-learn and scikit-image to their latest versions
I tried provisioning a fresh Python venv (no Conda involved) which took an uncomfortable amount of time (a dozen of minutes).

The culprits were by and large `scikit-learn` and `scikit-image`.  Both of them are pinned at ancient versions, for which there are no wheels published for recent releases of MacOS and so they must be built from source. `scikit-image` even failed to install on Python 3.9.

ghisvail:
Which will facilitate deployment of Clinica in environments powered by
more recent versions of Python and MacOS.

Closes #552


ghisvail:
I tried a deployment on Linux with Python 3.9 and it failed as well, sadly.

We should probably consider merging this early in the next dev cycle to raise the availability of Clinica for more recent Python stacks.

mdiazmel:
Upgrading to `scikit-learn = "^1.0"` seems a huge jump. Maybe we should run the non-regression tests for the concerned pipelines before merging, non? (at least the tests for the machine learning pipelines).

ghisvail:
Agreed

mdiazmel:
> grading to `scikit-learn = "^1.0"` seems a huge jump. Maybe we should run the non-regression tests for the concerned pipelines before merging, non? (at least the tests for the machine learning pipelines).

Done! 

I'll merge it!

PR #557, opened 2022-01-24
Add metrics to mergetsv output

omar-rifai:
Add more metrics to the output of `merge-tsv`: `thickness` for `t1-freesurfer-longitudinal`, and `MD` `RD` and `AD` for `DWI-DTI`.

### DIFF
--- a/clinica/iotools/utils/pipeline_handling.py
+++ b/clinica/iotools/utils/pipeline_handling.py
@@ -118,10 +118,6 @@ def t1_freesurfer_longitudinal_pipeline(
     Returns:
          final_df: a DataFrame containing the information of the bids and the pipeline
     """
-
-    from clinica.iotools.converters.adni_to_bids.adni_utils import (
-        replace_sequence_chars,
-    )
     from clinica.utils.stream import cprint
     from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


clinica/iotools/utils/pipeline_handling.py
```diff
@@ -196,7 +211,7 @@ def t1_freesurfer_longitudinal_pipeline(
                 )
                 atlas_df = pd.read_csv(atlas_path, sep="\t")
                 label_list = [
-                    "t1-freesurfer-segmentationVolumes_" + x
+                    "t1-fs-long_segmentationVolumes_" + x
```
ghisvail, line 214: If a change of naming convention is necessary, it'd be nice to adopt a naming scheme which is BIDS compatible.

i.e. `[<prefix1>-<label1>]_[<prefix2>-<label2>]_[...]_<suffix>`

where `<prefix>` is purely alphabetic, `<label>` is alphanumerical and `<suffix>` alphanumerical. It looks to me that we are introducing entities without prefix, or with multiple `-` which are reserved for separating key-value pairs.

ghisvail (changes requested):

ghisvail (approved):

PR #558, opened 2022-01-25
CI: Use Makefile to bootstrap environments

ghisvail:


Jenkinsfile
```diff
@@ -296,15 +298,16 @@ pipeline {
...
             sh '''
               source "${CONDA_HOME}/etc/profile.d/conda.sh"
-              conda create -p "${WORKSPACE}/env" python=3.8 poetry
-              conda activate "${WORKSPACE}/env"
-              poetry install --extras docs
-              mkdocs build -d "${CHANGE_ID:-$BRANCH_NAME}"
+              make env.conda
+              conda activate "${CONDA_ENV}"
+              make doc
+              mv site "${CHANGE_ID:-$BRANCH_NAME}"
```
mdiazmel, line 310: I just realized that for some unknown reason this command is not working... see PR #530 . The url for the documentation of this PR should be copied to [this url](https://aramislab.paris.inria.fr/clinica/docs/public/PR-558/) , but it's not the case.

mdiazmel (commented):
Thanks fot his @ghisvail .

We need to investigate why the command is not copying the files to the right folder before merging (see comment inside the code).

Jenkinsfile
ghisvail: That's because the correct URL is this [one](https://aramislab.paris.inria.fr/clinica/docs/public/558/).

mdiazmel: I see. We lost coherence with the documentation for old PRs where the url was `server/PR-XYZ` ... (sorry, my mistake because this change escaped me in PR#530). We should consider changing it to be uniform for all PRs (new and old ones).

ghisvail: Do we really _need_  that URL to be `<base_url>/PR-XXX` vs `<base_url>/XXX`?

If yes, then it means the simple substitution trick with `CHANGE_ID` and `BRANCH_NAME` would need to substituted with more complex logic.

> We should consider changing it to be uniform for all PRs (the new ones and the already published).

Is there something relying on the old behavior? Otherwise, I don't see a problem. The only reference that needs to remain consistent is the canonical URL to the stable / latest docs. Docs built as part of PRs are volatile and should be garbage collected eventually.

mdiazmel: That's true, the builts of the doc for old PRs are removed periodically. So, the inconsistency will be temporary. This is really **no critic**, we can left it like that. 

I'll merge!

PR #574, opened 2022-02-08
Fix dwi t1

MatthieuJoulot:
Dwi_preprocessing_using_t1 isn't currently working because of the deprecation of a few commands used in the code. This PR fixes the problem.

### DIFF
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
@@ -167,7 +167,7 @@ def ants_combine_transform(fix_image, moving_image, ants_warp_affine):
     out_warped = os.path.abspath("out_warped.nii.gz")
 
     cmd = (
-        f"antsApplyTransforms -o {out_warp_field} -i {moving_image} -r {fix_image} "
+        f"antsApplyTransforms -o [{out_warp_field},1] -i {moving_image} -r {fix_image} "
         f"-t {ants_warp_affine[0]} -t {ants_warp_affine[1]} -t {ants_warp_affine[2]}"
     )
     os.system(cmd)
@@ -179,17 +179,6 @@ def ants_combine_transform(fix_image, moving_image, ants_warp_affine):
     os.system(cmd1)
 
     return out_warp_field, out_warped
-
-
-# def create_jacobian_determinant_image(imageDimension, deformationField, outputImage):
-#     import os
-
-#     outputImage = os.path.abspath(outputImage)
-
-#     cmd = f"CreateJacobianDeterminantImage {str(imageDimension)} {deformationField} {outputImage}"
-#     os.system(cmd)
-
-#     return outputImage
 
 
 def init_input_node(t1w, dwi, bvec, bval, dwi_json):
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
@@ -88,7 +88,6 @@ def change_itk_transform_type(input_affine_file):
             else:
                 new_file_lines.append(line)
 
-
     updated_affine_file = os.path.join(os.getcwd(), "updated_affine.txt")
 
     with open(updated_affine_file, "wt") as f:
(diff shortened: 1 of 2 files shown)


clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -168,31 +167,20 @@ def ants_combine_transform(fix_image, moving_image, ants_warp_affine):
...
+        f"antsApplyTransforms -o [{out_warp_field},1] -i {moving_image} -r {fix_image} "
+        f"-t {ants_warp_affine[0]} -t {ants_warp_affine[1]} -t {ants_warp_affine[2]}"
     )
     os.system(cmd)
 
     cmd1 = (
-        f"antsApplyTransforms -o out_warped.nii.gz -i {moving_image} -r {fix_image} "
-        f"-t {ants_warp_affine[0]} {ants_warp_affine[1]} {ants_warp_affine[2]}"
+        f"antsApplyTransforms -o {out_warped} -i {moving_image} -r {fix_image} "
+        f"-t {ants_warp_affine[0]} -t {ants_warp_affine[1]} -t {ants_warp_affine[2]}"
```
ghisvail, line 177: Note for later, use nipype's interface for `antsApplyTransforms` instead of unsafe `os.system` calls.

ghisvail (approved):

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
ghisvail: Could be done in a follow-up PR if time is a constraint.

omar-rifai: If the interface exists, we ought to use it directly.

omar-rifai (changes requested):
We need to use the nipype interface as discussed.

MatthieuJoulot:
the old code has been left in comment in case the upgrade does not work.

MatthieuJoulot:
The results given by the `antsApplyTransforms` using nipype and without using nipype differ despite the command appearing to be exactly the same. So we chose to keep the version without nipype which gives a better result.

Screenshot 2022-02-14 at 09 23 05


Above,  the result of the pipeline without nipype on the left, and the version using nipype on the right.

NicolasGensollen:
@MatthieuJoulot if you have a MWE, it could be worth opening an issue upstream, WDYT?

MatthieuJoulot:
Not sure that it is worth it. Maybe it is. We use the nipype version of the command in other places (such as dti) and it works. Here, we have no idea why it does not work, which makes very little to work with for the nipype or ants folks.

ghisvail:
> Here, we have no idea why it does not work, which makes very little to work with for the nipype or ants folks.

Which is exactly why the issue should be forwared upstream (to `nipype` first, then `ANTs` transitively).

I second @NicolasGensollen suggestion to provide a MWE (if time allows it) to accelarate triage. It would increase the probability to get a fix quick.

Even if the issue never gets fixed, it would definitely help the rest of the community who is or will be affected by the very same issue by providing a single point for status tracking.

ghisvail:
> So we chose to keep the version without nipype which gives a better result.

In that case, let me suggest splitting the affected function into two -- one with image transformation, one with transformation field warping -- and wrap them in respective nipype nodes. Also maybe consider [replacing](https://docs.python.org/3/library/subprocess.html#replacing-os-system) `os.system`.

Node wiring would be easier to maintain and, in the event the issue gets fixed upstream, replacing with the appropriate nipype nodes would yield a smaller diff.

ghisvail:
> to provide a MWE (if time allows it) to accelarate triage

I can take care of it, if you give me the files used as inputs to the failing case.

ghisvail (approved):

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -160,36 +160,34 @@ def rotate_bvecs(in_bvec, in_matrix):
...
     import os
+    import subprocess
 
-    outputImage = os.path.abspath(outputImage)
+    out_warp_field = os.path.abspath("out_warp_field.nii.gz")
 
-    cmd = f"CreateJacobianDeterminantImage {str(imageDimension)} {deformationField} {outputImage}"
-    os.system(cmd)
+    cmd = (
+        f"antsApplyTransforms -o [{out_warp_field},1] -i {moving_image} -r {fix_image} "
```
omar-rifai, line 185: I understand the idea of creating two Nipype nodes in case the interface is used again. But for the moment it just seems we are redundantly calling the same command twice. I think a comment in the code to justify this would be useful.

ghisvail: > calling the same command twice

It's not the *same* command. Granted both share the same set of inputs, but output differs (warped image vs deformation field).

> the idea of creating two Nipype nodes in case the interface is used again

Using two nodes puts us closer to how we would have used the actual nipype interface for `antsApplyTransforms` if it were working as intended. More importantly, this setup also gives nipype more freedom to schedule workflow execution, since both transforms can be run in parallel (same inputs, different outputs).

ghisvail (changes requested):
I don't think we'have gained anything from this refactoring, tbh.

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
MatthieuJoulot: I've added another comment in the workflow part. I've also factorized this function. Do you think it is clearer ?

ghisvail: Ok, I think I have got an idea that may satisfy everyone:

ghisvail: let's define a function with the following signature:

```python
def antsApplyTransform(
    fixed_image: str,
    moving_image: str,
    transforms: List[str],
    warped_image: str,
    output_warped_image: bool = True
) -> None:
    import subprocess

    # Whether we want the warped image or transformation field as output.
    output = f"{warped_image}" if output_warped_image else f"[{warped_image}, 1]"
    
    # Common template for the command.
    cmd = (
        f"antsApplyTransforms -o {output} -i {moving_image} -r {fixed_image} "
        f"-t {transforms[0]} -t {transforms[1]} -t {transforms[2]}"
    )

    # Perform the actual call.
    subprocess.run(cmd, shell=True)
```

Wrap it in a Function node and set `warped_image` and `output_warped_image` accordingly when constructing the workflow. That would put us very close to the nipype interface, naming-wise and functionality-wise, whilst making a future replacement by `nipype.interfaces.ants.Registration` seamless.

omar-rifai: All right, so what I was missing is that the "output" is actually an argument that defines the behavior of the function. The proposal looks good.

ghisvail (approved):

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
ghisvail: Code updated with missing absolute path conversion.

MatthieuJoulot:
Since we had made extensive modifications to the pipelines, I have rerun the pipeline to confirm it works. It does.

PR #575, opened 2022-02-08
Update Clinica version in pyproject.toml

omar-rifai:
Update the clinica version in the `pyproject.toml` file to be inline with latest release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "clinica"
-version = "0.5.4"
+version = "0.5.5"
 description = "Software platform for clinical neuroimaging studies"
 license = "MIT"
 authors = ["ARAMIS Lab"]


mdiazmel:
If the metadata for the project wasn't updated at the moment of releasing the new version, I wonder where Pypi got the information to decide that it was the [release 0.5.5](https://pypi.org/project/clinica/0.5.5/)?
I'm definitely missing something...

omar-rifai:
I would guess it's from the tag, since this is what we are releasing.

ghisvail:
If I install Clinica from `dev`, I get:

```txt
$ poetry install
[...]
Installing the current project: clinica (0.5.4)
```

Then edit `pyproject.toml` to change version to `0.5.5`:

```txt
$ poetry install
[...]
Installing the current project: clinica (0.5.5)
```

Now if you were to publish the package without committing the modification to the version field, you could end up with this situation.

omar-rifai:
Yes I agree, the current version will be off if installed from `dev` (hence this PR ).  However if installing from `PyPi`, it will indeed install version `0.5.5`. Granted, if fetching the version number within the code (as with prov for example) this can be problematic: mea culpa.  Maybe it is worth considering a way to update this field automatically. I will look into it.

mdiazmel:
> I would guess it's from the tag, since this is what we are releasing.

The Pypi metadata is certainly not extracted from the git tag!  I continue asking where this information is extracted from?

omar-rifai:
The issue came from the release being made after the `tag` has been pushed but not the local `dev` branch.  We fixed the issue by "force" pushing the local `dev`. The commit in this merged PR hence does not show anymore.

PR #582, opened 2022-02-15
Activation of DWI CI tests

### LINKED ISSUE #549: Reactivate DWI non-regression tests once they have been validated locally
With the [latest modifications](https://github.com/aramis-lab/clinica/pull/547) to the DWI pipelines, the non-regression tests were disabled temporarily from scheduled execution (in Jenkinsfile scheduled). 

After local validation of these pipeline, we need to reactivate those tests.

MatthieuJoulot:
This PR makes the necessary changes to reactivate the DWI CI tests.

Closes #549

### DIFF
--- a/test/nonregression/pipelines/test_run_pipelines_dwi.py
+++ b/test/nonregression/pipelines/test_run_pipelines_dwi.py
@@ -17,8 +17,8 @@ warnings.filterwarnings("ignore")
 
 @pytest.fixture(
     params=[
-        # "DWIPreprocessingUsingT1",
-        # "DWIPreprocessingUsingPhaseDiffFieldmap",
+        "DWIPreprocessingUsingT1",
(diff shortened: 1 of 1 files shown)


ghisvail:
Requires #579

ghisvail (approved):

ghisvail:
Subject to validation on next Friday's meeting.

omar-rifai (approved):
Looks good. Are we OK with the current comparison function? Do we have other reserves on this PR?

MatthieuJoulot:
> Looks good. Are we OK with the current comparison function? Do we have other reserves on this PR?

Yes, the comparison function does what we want. 
Furthermore, the `similarity_measure` and `likeliness_measure` function are not doing exactly the same thing, so replacing one with the other would require more extensive researches on the reasons for which we use them.

Issue #588, opened 2022-03-01
Store source identifiers in BIDS metadata during conversion


ghisvail:
Whilst reviewing #587, it struck me that some parts of Clinica implement conversions from BIDS normalized IDs back to the source dataset IDs for subjects and sessions. It's probably not the only place where this is done, which leads to the potential for accidental code duplication of dataset-specific logic.

One solution to this would be to just store the source subject and session IDs in their respective metadata files, such as:

`participants.tsv`:

| participant_id | source_participant_id | group | age | sex |
| :-- | :-- | :-- | :-- | :-- |
| sub-ADNI011S0002 | 011_S_0002 | control | 22 | F |

`sub-<label>_sessions.tsv`:

| session_id | source_session_id | medhist* | npi* |
| :-- | :-- | :--| :-- | 
| ses-M00 | bl | n/a | n/a

This way we could assume that any dataset converted using Clinica provides source IDs *systematically* and *consistently*.

NicolasGensollen:
@AliceJoubert, again, no rush but also worth taking a look at this when you have some time.

PR #589, opened 2022-03-01
UKB Converter

MatthieuJoulot:
Draft of ukbiobank converter.  Based on the OASIS3 converter.


clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    import pandas as pd
+
+    from clinica.utils.stream import cprint
+
+    # Read the xls
+    try:
+        image_data_file = list(Path(clinical_data_directory).glob("*.csv"))
+    except StopIteration:
+        raise FileNotFoundError("Imaging collection file not found.")
```
omar-rifai, line 20: Shouldn't it be clinical data files not found?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # Assign the dfs
+    df_list = []
+    for i in range(0, len(image_data_file)):
+        df_list.append(pd.read_csv(str(image_data_file[i])))
+        df_list[i] = df_list[i].set_index(df_list[i].axes[1][0])
+        if df_list[i].index.name == "eid":
+            df_clinical = df_list[i]
```
omar-rifai, line 28: I'm not sure I understand this part,  we taking the last item in `df_list` with "eid" as index and assigning it to `df_clinical`?  Do we really need to maintain the list in that case?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+def read_imaging_data(imaging_data_directory: PathLike) -> DataFrame:
+    from pathlib import Path
+
+    import pandas as pd
+
+    source_path_series = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
+
+    source_dir_series = source_path_series.apply(lambda x: Path(str(x)).parent).rename(
```
omar-rifai, line 58: nitpicking the first "series" is the only one not dealing with "Path" and it's the only one with Path in its name

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+def find_imaging_data(path_to_source_data: PathLike) -> Iterable[PathLike]:
+    from pathlib import Path
+
+    for path in Path(path_to_source_data).rglob("*.nii.gz"):
+        yield str(path.relative_to(path_to_source_data))
+
+
+def intersect_data(df_source: DataFrame, dict_df: dict) -> Tuple[DataFrame, DataFrame]:
+    import pandas as pd
```
omar-rifai, line 90: The name of the function is not very informative, can we add a docstring to explain what we are trying to achieve with it?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+    df_source = df_source.assign(
+        ses=lambda df: df.session.apply(
+            lambda x: f"ses-M{ (5-(len(str(x)))) * '0' + str(int(x))}"
+        )
+    )
+
+    df_source = df_source.join(
+        df_source.modality.map(
```
omar-rifai, line 128: I feel this function can be modularized to reduce cognitive complexity

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+    return df_source, df_small
+
+
+def dataset_to_bids(
+    df_source: DataFrame, df_small: DataFrame
+) -> Tuple[DataFrame, DataFrame, DataFrame]:
+    # Build participants dataframe
+    df_participants = (
+        df_small[["participant_id", "ageAtEntry", "M/F", "Hand", "Education"]]
```
omar-rifai, line 190: More of an open question but shouldn't we rather have this mapping of name in a separate file (tsv?) for easier manipulation?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: Yes indeed.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: We set the first column as index, and we check that it is eid. That way, we know it is the file we are looking for.
It is probably a good security to have in case the users provides us. with a directory with more than one tsv

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,367 @@
...
+
+    # Read the xls
+    try:
+        image_data_file = list(Path(clinical_data_directory).glob("*.tsv"))
+    except StopIteration:
+        raise FileNotFoundError("Clinical data file not found.")
+
+    # Assign the dfs
+    df_list = []
+    for i in range(0, len(image_data_file)):
```
NicolasGensollen, line 25: suggestion: you could use `enumerate` here

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
NicolasGensollen: I agree with @omar-rifai you could just loop over the files, assign the right DataFrame to df_clinical, and not maintain the list since you're not using it afterwards.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: No problem ! I added a docstring, tell me if you feel it is not enough :)

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        cprint(msg="All clinical data have been found", lvl="info")
+    else:
+        raise FileNotFoundError("Clinical data not found or incomplete")
+
+    return df_clinical
+
+
+def read_clinical_data(
+    clinical_data_directory: PathLike,
+) -> Tuple[DataFrame, DataFrame, DataFrame, DataFrame, DataFrame]:
```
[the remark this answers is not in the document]
MatthieuJoulot: This comes from the original converter this one is inspired from (OASIS3) in which we had many dataframes, which lead to creating a dictionary of dataframes. Here there is only one though, so i have changed the signature.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+    # Assign the dfs
+    df_list = []
+    for i in range(0, len(image_data_file)):
+        df_list.append(pd.read_csv(str(image_data_file[i]), sep="\t"))
+        df_list[i] = df_list[i].set_index(df_list[i].axes[1][0])
+        if df_list[i].index.name == "eid":
+            df_clinical = df_list[i]
+
+    # Warn if a dataframe is missing
+    if "df_clinical" in locals():
```
NicolasGensollen, line 32: It is usually better to avoid this kind of pattern if possible. I would simply initialize df_clinical to None before the for loop and test here whether it was assigned or not.

```python
if df_clinical is not None:
    cprint("bla bla")
else:
    raise("bla bla")
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+        cprint(msg="All clinical data have been found", lvl="info")
+    else:
+        raise FileNotFoundError("Clinical data not found or incomplete")
+
+    return df_clinical
+
+
+def read_clinical_data(
```
NicolasGensollen, line 40: Is there a reason you created this function? It looks like it is not doing anything more than `find_clinical_data`.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+
+def read_imaging_data(imaging_data_directory: PathLike) -> DataFrame:
+    from pathlib import Path
+
+    import pandas as pd
+
+    source_path_series = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
+    # list of the files we want to build the bids for each mopdalkty
```
NicolasGensollen, line 55:
```suggestion
    # list of the files we want to build the bids for each modality
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+        for f in ZipFile(z).namelist():
+            if f.endswith(".nii.gz"):
+                yield [str(z.relative_to(path_to_source_data)), f]
+
+
+def intersect_data(
+    df_source: DataFrame, dict_df: DataFrame
+) -> Tuple[DataFrame, DataFrame]:
+    """
+    This function merges the two dataframe given as inputs based on the subject id, and uses the existing data
```
NicolasGensollen, line 103:
```suggestion
    This function merges the two dataframes given as inputs based on the subject id, and uses the existing data
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
omar-rifai: It seems somewhat contrived if the purpose is to check that we don't have many `.csv`files.  Moreover, if we are making the hypothesis that there can possible be more than one, why are we taking the last (as opposed to the first?).

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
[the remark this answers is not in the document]
omar-rifai: and the return type?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+        install_sidecars(
+            zipfile=str(dataset_directory) + "/" + metadata["source_zipfile"],
+            filename=metadata["sidecars"],
+            bids_path=to / metadata["bids_sidecar_path"],
+        )
+    return
+
+
+def install_nifti(zipfile: str, filename: str, bids_path: str) -> None:
```
NicolasGensollen, line 328: Might be just me, but I find the names (esp. the `install` prefix) a bit misleading.
It looks like these functions are extracting files from an archive and writing them to some location.
Also, for DRYness, it looks like you could easily combine `install_nifti` with `install_json` since they have almost exactly the same code.

NicolasGensollen (commented):
Thanks @MatthieuJoulot ! 👍 
Just a few suggestions and questions:

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: Same as others, it comes from the structure of the original converter the code is inspired from. you are right i could do without.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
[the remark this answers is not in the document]
MatthieuJoulot: I have changed it as well 😄

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: simplified !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: Yes, i was thinking about it as well. I may not combine `install_nifti`and `install_json`, but i will probably combine `install_json`and `install_sidecars`

omar-rifai: I'd agree with @NicolasGensollen on this one,  as the content is almost identical, maybe rename the function to something more generic as `extract_data` and reuse.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,369 @@
...
+            f.write(fs.cat(filename))
+
+
+def install_sidecars(zipfile: str, filename: str, bids_path: str) -> None:
+    """Install a NIfTI file from a source archive to the target BIDS path."""
+    import fsspec
+
+    fo = fsspec.open(zipfile)
+    fs = fsspec.filesystem("zip", fo=fo)
+    for i in range(0, len(filename)):
```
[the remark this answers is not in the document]
omar-rifai: (ignore above comment if refactoring is done as suggested by @NicolasGensollen )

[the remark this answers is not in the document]
ghisvail: Agreed, `filename: str` -> `filename: List[str]`

Also, replace:

```
for i in len(filenames):
    filenames[i]
```

by:

```python
for filename in filenames:
    # do something with filename
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,319 @@
...
+
+    for z in Path(path_to_source_data).rglob("*.zip"):
+        for f in ZipFile(z).namelist():
+            if f.endswith(".nii.gz"):
+                yield [str(z.relative_to(path_to_source_data)), f]
+
+
+def intersect_data(
+    df_source: DataFrame, df_clinical_data: DataFrame
+) -> Tuple[DataFrame, DataFrame]:
```
NicolasGensollen, line 88: This function currently does two different things:

- It performs a `merge` on `df_clinical_data` and `df_source` and assign it to `df_clinical` 
- Does some DataFrame operations on `df_clinical` without needing anything else
- And it returns `df_source` without doing any modifications on it 

It looks like you could split this in at least two functions:

- `intersect_source_and_clinical_data(df_source, df_clinical_data) -> df_clinical`
- `clean_clinical_data(df_clinical) -> df_clinical`

Or something like this, WDYT?

ghisvail: Sounds like a sensible proposal. Having a cleaning / sanitization process for each dataframe, followed by a merge step.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: Done !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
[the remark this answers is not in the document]
MatthieuJoulot: done as well :)

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: Sounds good, I'll look into it right now 😄

MatthieuJoulot: @ghisvail @NicolasGensollen I have changed the intersect_data function and split in two. Is it what you were thinking of ?

ghisvail: This is how a *somewhat* generic converter looks like to me:

```mermaid
flowchart LR
LCD --> CCD --> MCI
LID --> CID --> MCI
MCI --> PCI --> WCI
```

with:
- LCD, LID: Load clinical / imaging data
- CCD, CID: Clean clinical / imaging data
- MCI: Merge clinical and imaging data
- PCI: Prepare merged data for writing
- WCI: Write merged data to BIDS

If we all agree with this, then we could keep those names.

ghisvail: I have been thinking about this for a while (since we talked about migrating to `pydra`), and having a generic structure for a converter could help standardizing towards a workflow compatible design.

clinica/iotools/converters/ukb_to_bids/ukb_to_bids.py
```diff
@@ -0,0 +1,50 @@
...
+    path_to_clinical: PathLike,
+) -> List[PathLike]:
+    """Convert the entire dataset in BIDS.
+
+    Scans available files in the path_to_dataset,
+    identifies the patients that have images described by the JSON file,
+    converts the image with the highest quality for each category.
+    """
+    from .ukb_utils import (
+        complete_clical,
```
NicolasGensollen, line 19:
```suggestion
        complete_clinical,
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,305 @@
...
+    """
+    import pandas as pd
+
+    df_clinical = df_clinical_data.merge(
+        df_source, how="inner", right_on="Subject", left_on="eid"
+    )
+    return df_clinical
+
+
+def complete_clical(df_clinical: DataFrame) -> DataFrame:
```
NicolasGensollen, line 98:
```suggestion
def complete_clinical(df_clinical: DataFrame) -> DataFrame:
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,305 @@
...
+    for participant_id, data_frame in sessions.groupby(["participant_id"]):
+        session = data_frame.droplevel(["participant_id", "suffix"]).drop_duplicates()
+
+        session_filepath = to / participant_id / f"{participant_id}_sessions.tsv"
+        with fs.open(session_filepath, "w") as sessions_file:
+            write_to_tsv(session, sessions_file)
+
+    scans = scans.set_index(["bids_full_path"], verify_integrity=True)
+    for bids_full_path, metadata in scans.iterrows():
+        copy_file_to_bids(
```
NicolasGensollen, line 270: Looks like this can be factorized easily

NicolasGensollen (commented):
Thanks @MatthieuJoulot !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
NicolasGensollen: @ghisvail I agree. If all converters follow (or could potentially be adapted to follow) this pattern, then it would definitely be beneficial to standardize them.

MatthieuJoulot: Okay. I won't change this right now since other modification might change our point of view on this, but i will come back to you both when all the modalities are included in the converter.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: done !

ghisvail:
@MatthieuJoulot forget about the source IDs, let's deal with them for all converters at once when addressing #588. That will keep this review simpler.

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
+<!-- markdownlint-disable MD046 -->
+# `ukb-to-bids` – Conversion of the UK Biobank (UKB) to BIDS
+
+!!! quote "Description reproduced from the [UK Biobank webpage](https://www.ukbiobank.ac.uk/)"
+    UK Biobank is a large-scale biomedical database and research resource, containing in-depth genetic and health information from half a million UK participants. The database is regularly augmented with additional data and is globally accessible to approved researchers undertaking vital research into the most common and life-threatening diseases. It is a major contributor to the advancement of modern medicine and treatment and has enabled several scientific discoveries that improve human health
+
```
omar-rifai, line 6: missing .

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+## Dependencies
+
+If you installed the core of Clinica, this converter needs no further dependencies.
```
omar-rifai, line 9: Doesn't it require dcm2niix?

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Downloading UK Biobank
+
+The UK Biobank to BIDS converter requires the user to have downloaded the UK Biobank imaging and clinical data.
```
omar-rifai, line 13: This doesn't make much sense given the next sentence. Do you rather mean "requires the user to have registered"?

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+
+Once you have access to it, the imaging data can be downloaded and extracted using this tutorial: https://biobank.ctsu.ox.ac.uk/~bbdatan/Accessing_UKB_data_v2.3.pdf, and the clinical data can be extracted using the information from this repository https://github.com/kenhanscombe/ukbtools.
+
+
+Be careful about two things: 
+    - the converter only takes in the nifti provided by UK Biobank, not the dicoms.
```
omar-rifai, line 19: maybe add a note in parenthesis for `fmri`

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+    - the converter will not work if you do not have among the clinical data the year of birth, the age at recruitment and at sessions and the sex of the subject.
```
omar-rifai, line 20: This seems difficult for a user to check. Maybe just add this message as an error message in the code if we can not find them ?

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+Please note that this converter processes the following modalities : 
+- T1W
+- T2 Flair
+- DWI
+- SWI
+- tfMRI
+- rsfMRI
+
+!!! Chosen files
+    Whenever possible, we use the rawest files in order not to impose opinions onto the user. When available we get the associated json.
```
omar-rifai, line 33: "in order not to impose opinions" --> "so as to leave it to the user to choose the processing needed"

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+    - T1W -> T1.nii.gz: cropped so there is no neck and defaced. The rawest image would be the simply defaced one, but since we have no interest in the neck we don't use it. We do not want any other corrections.
```
omar-rifai, line 34: Have these as a table it'll be easier to read.

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+    - T2 Flair -> T2_FLAIR: cropped so there is no neck and defaced. The reasons for choosing this image are the same as for T1.
+    - DWI -> AP.nii.gz/PA.nii.gz + bvals and bvecs. It is the rawest images we can get.
+    - rsfMRI -> rfMRI.nii.gz
+    - tfMRI -> tfMRI.nii.gz or dicom equivalent to get the json associated
+    - SWI -> SWI.nii.gz which is the combined coil version. We would get a rawer version, but dcm2niix seems to have trouble handling this format, so it is simpler to go with this version. Plus, swi is not fully integrated to bids so changes may be coming, hence a simple version for now.
```
omar-rifai, line 39: "dcm2niix seems to have trouble handling this format" is too broad and unspecific. Reformulate.

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,87 @@
...
+
+The converter can be run with the following command line:
+
+```Text
+clinica convert ukb-to-bids [OPTIONS] DATASET_DIRECTORY CLINICAL_DATA_DIRECTORY BIDS_DIRECTORY 
+```
+
+where:
+
+- `DATASET_DIRECTORY` is the path to the original UK BIobank imaging directory, which content should look like:
```
omar-rifai, line 53: "which content" --> "whose content"

omar-rifai (changes requested):

docs/Converters/UKBtoBIDS.md
MatthieuJoulot: not right now since we take only the nifti as inputs.

docs/Converters/UKBtoBIDS.md
MatthieuJoulot: I think it is basic data to have (not required by bids but always required by any study). And not very hard to check either ! Do you still want me to add a message in the code ?

ghisvail: Let's place ourselves in the perspective of our users. First, they rarely reach for documentation *first*, only in case there is an issue (oftentimes) or when they are curious (sometimes). Second, imagine I run the converter on a datatset freshly downloaded from UKB and get a stack trace. Where do I get from here? Errors are usually for *exceptional* circumstances, not specific invariants that were not met that we deliberatly know of.

> the converter will not work if you do not have among the clinical data the year of birth, the age at recruitment and at sessions and the sex of the subject.

In my opinion, this is an invariant (or constraint) of our converter. Given any downloaded UKB datasets, we can determine whether these attributes are available as a validation step, prior to the conversion. So I'd expect a validation error with an appropriate message be thrown early in the execution, rather than a stack trace complaining of a missing `age` attribute in some DataFrame deep within the converter's logic.

Does it make sense to you?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,344 @@
...
+    fo = fsspec.open(zipfile)
+    fs = fsspec.filesystem("zip", fo=fo)
+    for filename in filenames:
+        if fs.exists(filename):
+            bids_path_extension = str(bids_path) + "." + (filename.split(".", 1)[1])
+            with fsspec.open(bids_path_extension, mode="wb") as f:
+                f.write(fs.cat(filename))
+
+
+def select_session(x):
```
NicolasGensollen, line 340: Is there a reason you didn't add type hints for this function?

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,344 @@
...
+    df_scan = df_clinical.filter(items=list(df_ref["scan"]))
+    return df_participants, df_session, df_scan
+
+
+def write_to_tsv(dataframe: DataFrame, buffer: Union[PathLike, BinaryIO]) -> None:
+    # Save dataframe as a BIDS-compliant TSV file.
+    dataframe.to_csv(buffer, sep="\t", na_rep="n/a", date_format="%Y-%m-%d")
+
+
+def write_bids(
```
NicolasGensollen, line 281: It feels like this function isn't specific to the `ukb_to_bids` converter. It could be worth having it as a utility function and sharing it across converters. WDYT?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
I have a few suggestions/questions below. Feel free to take or leave.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
omar-rifai: I'd second that opinion. But for the sake of restricting the scope of this PR, I'd suggest just moving the function to a common utility module and not refactoring the interface with other converters.

With "recent" converters such as `HABS` or `Oassis3` the integration should require minimal effort but might be a bit more costly with other converters. At any rate, best to separate this effort from the core converter.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,344 @@
...
+        with fs.open(session_filepath, "w") as sessions_file:
+            write_to_tsv(session, sessions_file)
+    scans = scans.set_index(["bids_full_path"], verify_integrity=True)
+    for bids_full_path, metadata in scans.iterrows():
+        copy_file_to_bids(
+            zipfile=str(dataset_directory) + "/" + metadata["source_zipfile"],
+            filenames=[metadata["source_filename"]] + metadata["sidecars"],
+            bids_path=to / bids_full_path,
+        )
+    return
```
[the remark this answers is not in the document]
MatthieuJoulot: Good catch !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: There is ! I have forgotten to add it 😅

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: The write_bids function may be a little different for each converter depending on the data we have ! but the write_to_tsv function right above was moved 😄

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,386 @@
...
+        "ukb_ref.csv",
+    )
+    df_ref = pd.read_csv(path_to_ref_csv, sep=";")
+
+    df_clinical = df_clinical.set_index(
+        ["participant_id", "session", "modality", "bids_filename"],
+        verify_integrity=True,
+    )
+
+    # Build participants dataframe
```
NicolasGensollen, line 282: I think you can factorize a little bit here. Maybe something like this:?

```python
return (df_clinical.filter(items=list(df_ref[_])) for _ in ["participants", "session", "scan"])
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,386 @@
...
+    import pandas as pd
+
+    df_clinical = df_clinical_data.merge(
+        df_source, how="inner", right_on="source_id", left_on="eid"
+    )
+    return df_clinical
+
+
+def complete_clinical(df_clinical: DataFrame) -> DataFrame:
+    """This funbction uses the existing data to create the columns needed for the bids hierarchy (subject_id, ses, age_at _session, ect.)"""
```
NicolasGensollen, line 134:
```suggestion
    """This function uses the existing data to create the columns needed for the bids hierarchy (subject_id, ses, age_at _session, ect.)"""
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,386 @@
...
+def convert_dicom_to_nifti(zipfiles: str, bids_path: str) -> None:
+    """Install the requested files in the BIDS  dataset."""
+    import os
+    import subprocess
+    import tempfile
+    import zipfile
+
+    zf = zipfile.ZipFile(zipfiles)
+    try:
+        os.makedirs(bids_path.parent)
```
NicolasGensollen, line 368: The function signature says that `bids_path` is assumed to be a string while this is assuming a Path object without any cast or checks.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,386 @@
...
+    except OSError:
+        # Folder already created with previous instance
+        pass
+    with tempfile.TemporaryDirectory() as tempdir:
+        zf.extractall(tempdir)
+        command = ["dcm2niix", "-w", "0", "-f", bids_path.name, "-o", bids_path.parent]
```
NicolasGensollen, line 374: I think that adding a docstring with the command you use in plain text (and even a few explanations for the options selected) would be super useful for maintenance. WDYT?

MatthieuJoulot: Yes it feels like a good idea !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
MatthieuJoulot: nice one !

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,390 @@
...
+        command += [tempdir]
+        subprocess.run(command)
+    return
+
+
+def select_session(x: int) -> int:
+    if x["source_session_number"] == "2":
+        return x.age_when_attended_assessment_centre_f21003_2_0
+    elif x["source_session_number"] == "3":
+        return x.age_when_attended_assessment_centre_f21003_3_0
```
ghisvail, line 390: Static analysis complains and rightfully so. `x` is definitely not an int here.

Could this be made simpler? At the end of the day all we need is a mapping:

```python
{
    "2": "age_when_attended_assessment_centre_f21003_2_0",
    "3": "age_when_attended_assessment_centre_f21003_3_0",
}
```

or a function:

```python
def get_age_column(session: int) -> str:
    return f"age_when_attended_assessment_centre_f21003_{session}_0"
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+
+    df_clinical = df_clinical_data.merge(
+        df_source, how="inner", right_on="source_id", left_on="eid"
+    )
+    return df_clinical
+
+
+def complete_clinical(df_clinical: DataFrame) -> DataFrame:
+    """This function uses the existing data to create the columns needed for the bids hierarchy (subject_id, ses, age_at _session, ect.)"""
```
ghisvail, line 132: PEP8 complains the line is too long.

```suggestion
    """
    This function uses the existing data to create the columns needed for
    the BIDS hierarchy (subject_id, ses, age_at _session, etc...).
    """
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,390 @@
...
+
+    import pandas as pd
+
+    source_path_series_nifti = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
+    source_path_series_dicom = pd.Series(
+        find_dicom_data(imaging_data_directory), name="source_path"
+    )
+    # justifications regarding the chosen files can be found in clinica's documentation: https://github.com/aramis-lab/clinica/blob/dev/docs/Converters/UKBtoBIDS.md
```
ghisvail, line 56: Silence PEP8 + rewording:

```suggestion
    # Rationales for this choice of files can be found in the documentation:
    # https://github.com/aramis-lab/clinica/blob/dev/docs/Converters/UKBtoBIDS.md
```

ghisvail (changes requested):
First batch of review from what PyCharm warns about.

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,86 @@
...
+
+If you installed the core of Clinica, this converter needs dcm2niix.
+
+## Downloading UK Biobank
+
+The UK Biobank to BIDS converter requires the user to have applied to the data and to have downloaded the UK Biobank imaging and clinical data.
+
+Once you have access to it, the imaging data can be downloaded and extracted using this tutorial: https://biobank.ctsu.ox.ac.uk/~bbdatan/Accessing_UKB_data_v2.3.pdf, and the clinical data can be extracted using the information from this repository https://github.com/kenhanscombe/ukbtools.
+
+Be careful, the converter only takes in the nifti provided by UK Biobank, not the dicoms.
```
omar-rifai, line 17: this part is no longer true based on the recent fmri efforts

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,86 @@
...
+- rsfMRI (dicom)
+
+!!! Chosen files
+    Whenever possible, we use the rawest files so as to leave it to the user to choose the processing needed. When available we get the associated json.
+| Modality    | Chosen image(s) | Justification |
+| :----------:|:---------------:|:-------:|
+| T1W                     | T1.nii.gz       | Defaced and cropped so there is no neck. The rawest image would be the simply defaced one, but since we have no interest in the neck we don't use it. We do not want any other corrections. |
+| T2 Flair    | T2_FLAIR.nii.gz | Defaced and cropped so there is no neck. The reasons for choosing this image are the same as for T1. |
+| DWI         | AP/PAnii.gz     | It is the rawest images we can get. bvals and bvec are also available. |
+| rsfMRI      | rsfMRI.dcm | Since the nii.gz version doesn't always include a json and in consideration for it's usage, we convert the dicom. |
```
omar-rifai, line 37: "in consideration for it's usage" sounds a bit awkward --> "The dicom sources are used to extract BIDS required metadata not found otherwise"

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,86 @@
...
+| tfMRI       | tfMRI.dcm   | Same as rsfMRI.|
+| SWI         | SWI.nii.gz      | Combined coil version. We would get a rawer version, but dcm2niix has trouble handling the slices direction, so it is simpler to go with this version. Plus, swi is not fully integrated to bids so changes may be coming, hence a simple version for now. |
```
omar-rifai, line 39: A bit too lengthy and informal --> "Combined coil version for dcm2niix compatibility."

clinica/iotools/converters/ukb_to_bids/ukb_to_bids.py
```diff
@@ -0,0 +1,50 @@
...
+from os import PathLike
+from typing import List
+
+
+def convert_images(
+    path_to_dataset: PathLike,
+    bids_dir: PathLike,
+    path_to_clinical: PathLike,
+) -> List[PathLike]:
+    """Convert the entire dataset in BIDS.
```
omar-rifai, line 12: "to BIDS?"

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    import pandas as pd
+
+    source_path_series_nifti = pd.Series(
+        find_imaging_data(imaging_data_directory), name="source_path"
+    )
+    source_path_series_dicom = pd.Series(
+        find_dicom_data(imaging_data_directory), name="source_path"
+    )
+    # Rotionales for this choice of files can be found in the documentation:
```
omar-rifai, line 55: Rationale*

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,393 @@
...
+    )
+
+    return df_source
+
+
+def find_dicom_data(path_to_source_data: PathLike) -> Iterable[PathLike]:
+    from pathlib import Path
+
+    for z in Path(path_to_source_data).rglob("*.zip"):
+        if str(z).split("_")[1] == "20217":
```
omar-rifai, line 104: maybe add a docstring with a link to the specs that makes sense of these IDS

omar-rifai (commented):
This is starting to look really neat. Is there a part that still needs working on from your side before we test?

clinica/iotools/converters/ukb_to_bids/ukb_to_bids.py
omar-rifai, line 12:
```suggestion
    """Convert the entire dataset to BIDS.
```

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -394,3 +396,21 @@ def select_sessions(x: DataFrame) -> Series:
         return x.age_when_attended_assessment_centre_f21003_2_0
     elif x["source_sessions_number"] == "3":
         return x.age_when_attended_assessment_centre_f21003_3_0
+
+
+def import_event_tsv(bids_path: str) -> None:
```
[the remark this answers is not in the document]
MatthieuJoulot: I agree on a theorical standpoint, but i really dislike the idea of installing it there, since it is meant to be used only by tfmri data and the tsv would be above not only fmri but the other modalities as well.

That being said, either options is okay with me. wdyt @ghisvail

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,422 @@
...
+
+    from fsspec.implementations.local import LocalFileSystem
+
+    fs = LocalFileSystem(auto_mkdir=True)
+    path_to_event_tsv = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))),
+        "resources",
+        "fmri",
+        "ukb_event.tsv",
+    )
```
ghisvail, line 414: I wonder whether it would not make sense to just store this file under something like `resources/ukb/task-task-facesshapesemotion_events.tsv` and copy it as is to the output BIDS directory.

That would avoid the extra code for renaming and computing the relative path with `.parent` which looks a bit weird to me.

clinica/iotools/converters/ukb_to_bids/ukb_utils.py
```diff
@@ -0,0 +1,420 @@
...
+    path_to_event_tsv = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))),
+        "resources",
+        "fmri",
+        "task-facesshapesemotion_bold_events.tsv",
+    )
+
+    bids_path_extension = (
+        str(bids_path) + "/" + "task-facesshapesemotion_bold_events.tsv"
+    )
```
ghisvail, line 418: typo:

```suggestion
    fs = LocalFileSystem(auto_mkdir=True)
    path_to_event_tsv = os.path.join(
        os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))),
        "resources",
        "fmri",
        "task-facesshapesemotion_events.tsv",
    )

    bids_path_extension = (
        str(bids_path) + "/" + "task-facesshapesemotion_events.tsv"
    )
```

clinica/iotools/bids_utils.py
```diff
@@ -818,3 +818,8 @@ def run_dcm2niix(
             ),
             lvl="warning",
         )
+
+
+def write_to_tsv(dataframe: DataFrame, buffer: Union[PathLike, BinaryIO]) -> None:
```
omar-rifai, line 823: I'm getting a NameError on the types here. Are they properly defined?

mdiazmel, line 823: This line should be something like this:
```suggestion
def write_to_tsv(dataframe: pd.DataFrame, buffer: Union[PathLike, BinaryIO]) -> None:
```

mdiazmel: Also need to add some lacking imports for the `PathLike` and `BinaryIO` type annotations, something like this:
```
from os import PathLike
from typing import List, Optional, Union, BinaryIO
```
at the beginning of the file.

MatthieuJoulot: I have only imported DataFrame from pandas not to import the whole library. Is that okay ?

MatthieuJoulot: They were not.

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+!!! quote "Description reproduced from the [UK Biobank webpage](https://www.ukbiobank.ac.uk/)"
+    UK Biobank is a large-scale biomedical database and research resource, containing in-depth genetic and health information from half a million UK participants. The database is regularly augmented with additional data and is globally accessible to approved researchers undertaking vital research into the most common and life-threatening diseases. It is a major contributor to the advancement of modern medicine and treatment and has enabled several scientific discoveries that improve human health.
+
+## Dependencies
+
+If you installed the core of Clinica, this converter needs dcm2niix.
+
+## Downloading UK Biobank
+
+The UK Biobank to BIDS converter requires the user to have applied to the data and to have downloaded the UK Biobank imaging and clinical data.
```
mdiazmel, line 13:
```suggestion
The UK Biobank to BIDS converter assumes that the user have already applied to get access to the data and that the imaging and clinical data have already been downloaded locally.
```

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+
+Once you have access to it, the imaging data can be downloaded and extracted using this tutorial: https://biobank.ctsu.ox.ac.uk/~bbdatan/Accessing_UKB_data_v2.3.pdf, and the clinical data can be extracted using the information from this repository https://github.com/kenhanscombe/ukbtools.
```
mdiazmel, line 15:
```suggestion
The imaging data can be downloaded and extracted using [this tutorial](https://biobank.ctsu.ox.ac.uk/~bbdatan/Accessing_UKB_data_v2.3.pdf). The clinical data can be extracted using the information from [this repository](https://github.com/kenhanscombe/ukbtools).
```

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+
+Be careful, the clinical_data.tsv needs to contain these informations about the subject: sex, year of birth, age at recruitment, age at sessions.
```
mdiazmel, line 17:
```suggestion
Be careful, the file `clinical_data.tsv` needs to contain the following nformations about the subject: sex, year of birth, age at recruitment, age at sessions.
```

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+Please note that this converter processes the following modalities : 
+- T1W (nifti)
+- T2 Flair (nifti)
+- DWI (nifti)
+- SWI (nifti)
+- tfMRI (dicom)
+- rsfMRI (dicom)
+
+!!! Chosen files
+    Whenever possible, we use the rawest files so as to leave it to the user to choose the processing needed. When available we get the associated json.
```
mdiazmel, line 30:
```suggestion
    Whenever is possible, the converter uses the rawest files found. This decision allows the user to choose the processing needed. When available the ocnverter get the associated json.
```

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+| Modality    | Chosen image(s) | Justification |
+| :----------:|:---------------:|:-------:|
+| T1W                     | T1.nii.gz       | Defaced and cropped so there is no neck. The rawest image would be the simply defaced one, but since we have no interest in the neck we don't use it. We do not want any other corrections. |
```
mdiazmel, line 33:
```suggestion
| T1W                     | T1.nii.gz       | Defaced and cropped so there is no neck. The rawest image would be the simply defaced one, but brain studies usually are not interested in the region of the neck. No other corrections are included. |
```

docs/Converters/UKBtoBIDS.md
```diff
@@ -0,0 +1,85 @@
...
+| T2 Flair    | T2_FLAIR.nii.gz | Defaced and cropped so there is no neck. The reasons for choosing this image are the same as for T1. |
+| DWI         | AP/PAnii.gz     | It is the rawest images we can get. bvals and bvec are also available. |
+| rsfMRI      | rsfMRI.dcm | Since the nii.gz version doesn't always include a json and in consideration for it's usage, we convert the dicom. |
+| tfMRI       | tfMRI.dcm   | Same as rsfMRI.|
+| SWI         | SWI.nii.gz      | Combined coil version. We would get a rawer version, but dcm2niix has trouble handling the slices direction, so it is simpler to go with this version. Plus, swi is not fully integrated to bids so changes may be coming, hence a simple version for now. |
```
mdiazmel, line 38:
```suggestion
| SWI         | SWI.nii.gz      | Combined coil version. We would get a rawer version, but `dcm2niix` has trouble handling the slices direction, so it is simpler to go with this version. In addition, SWI modality is not fully integrated to BIDS specification and some changes may be coming. A simple version is available in the current version of this converter. | 
```

mdiazmel (changes requested):
Seems almost ready!

There are some small bugs (importing issues) to solve in the file `clinica/iotools/bids_utils.py`. Otherwise, it runs without problems.

I also suggest some changes in the documentation.

clinica/iotools/bids_utils.py
mdiazmel: > I have only imported DataFrame from pandas not to import the whole library. Is that okay ?

Yes, that's the right way to do it! Please run the `isort` hook before committing, it seems that the order of the imports is not the right one.

ghisvail: > Please run the isort hook before committing

or `make format`

docs/Converters/UKBtoBIDS.md
omar-rifai: *information

docs/Converters/UKBtoBIDS.md
omar-rifai: *has already applied

docs/Converters/UKBtoBIDS.md
omar-rifai: *converter

clinica/iotools/bids_utils.py
MatthieuJoulot: the `isort`hook has run the imports should be in the correct order

omar-rifai:
Thanks @MatthieuJoulot ! Will merge as it passes CI and seems working. If further changes needed we can create a different PR.

PR #590, opened 2022-03-02
Bump BIDS version to 1.7.0

ghisvail:
See https://github.com/bids-standard/bids-specification/releases/tag/v1.7.0


ghisvail:
> Does this mean we satisfy BIDS 1.7.0 requirements in the converters?

Indeed (for those not specifying a version explictly).

PR #596, opened 2022-03-07
Fix parser warning in AIBL reader

ghisvail:
ParserWarning: Falling back to the 'python' engine because the 'c'
engine does not support regex separators (separators > 1 char and
different from '\s+' are interpreted as regex); you can avoid this
warning by specifying engine='python'.


NicolasGensollen (approved):
LGTM, thanks!
[link](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.errors.ParserWarning.html?highlight=parserwarning) to the warning's documentation.

PR #597, opened 2022-03-07
Fix Pandas warning in AIBL writer

ghisvail:
Instead of normalizing the participant's dataframe inplace row per row,
perform data transformations column by column. This is both more
effective and makes this warning disappear.

### DIFF
--- a/clinica/iotools/converters/aibl_to_bids/aibl_utils.py
+++ b/clinica/iotools/converters/aibl_to_bids/aibl_utils.py
@@ -719,17 +719,16 @@ def create_participants_df_AIBL(
             # Add the extracted column to the participant_df
             participant_df[participant_fields_bids[i]] = pd.Series(field_col_values)
 
-    # Adding participant_id column with BIDS ids
-    for i in range(0, len(participant_df)):
-        value = participant_df["alternative_id_1"][i]
-        participant_df["participant_id"][i] = "sub-AIBL" + str(value)
-        participant_df["date_of_birth"][i] = re.search(
-            "/([0-9].*)", str(participant_df["date_of_birth"][i])
-        ).group(1)
-        if participant_df["sex"][i] == 1:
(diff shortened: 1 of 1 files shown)


clinica/iotools/converters/aibl_to_bids/aibl_utils.py
```diff
@@ -719,17 +719,16 @@ def create_participants_df_AIBL(
...
+    # Compute BIDS-compatible participant ID.
+    participant_df["participant_id"] = "sub-AIBL" + participant_df["alternative_id_1"]
+
+    # Keep year-of-birth only.
+    participant_df["date_of_birth"] = participant_df["date_of_birth"].str.extract(
+        r"/(\d{4}).*"
+    )
+
+    # Normalize sex value.
+    participant_df["sex"] = participant_df["sex"].map({1: "M"}).fillna("F")
```
omar-rifai, line 731: what happens if there is a value of `0`  in this case?

ghisvail: Any value other than 1 becomes "F". That's equivalent to the old imperative implementation.

omar-rifai: I'd suggest having an explicit map from `0` to `F`  (if it is indeed what encodes `F` in the specs) rather than defaulting to `F`. This might prevent us from converting `n/a` or otherwise invalid values to `F`.

ghisvail: After verification, "F" is mapped to 2.

Before:

```
0    1
1    2
2    2
3    1
4    1
Name: sex, dtype: int64
```

After:

```

0    M
1    F
2    F
3    M
4    M
Name: sex, dtype: object
```

omar-rifai:
Great, thanks !

PR #601, opened 2022-03-09
Fix working directory for pet-surface

ghisvail:
This PR makes the `pet-surface` pipeline write to its reserved folder (like other PET pipelines do), instead of the root working directory directly.


omar-rifai:
LGTM, will merge.

PR #602, opened 2022-03-11
Fix BIDS compatibility of DWI converter

ghisvail:


ghisvail:
Can be merged without changes to the CI dataset, since
- the latter does not include DWI test data for ADNI
- search patterns in DWI pipelines are flexible enough

omar-rifai:
For all the entities that we consider as superfluous as per the "entity parsimony principle in BIDS", we ought to provide the information in a `json` metadata file, no?

ghisvail:
> For all the entities that we consider as superfluous as per the "entity parsimony principle in BIDS", we ought to provide the information in a `json` metadata file, no?

If they ought to be used, yes. The recommended metdata for DWI are [here](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#other-recommended-metadata_1).

That being said, it does not look like our DWI processing pipelines are using the acquisition entity.

ghisvail:
> the latter does not include DWI test data for ADNI

I stand corrected by our venerable CI. There is some renaming to be done in the test dataset for ADNI.

NicolasGensollen:
> If they ought to be used, yes.

I am not sure to understand what you mean by that. How do you know this information wouldn't be useful for some users, or for future work? For some reason, I feel like all available information in the inputs should be available in the output if it does fit within the BIDS specs.

omar-rifai:
It's probably going to be a pain but I'd suggest we get into the habit of systematically moving the information we are taking out of the entities in `.json` files.

ghisvail:
I'd agree in principle with the following nuance:

The DWI spec lists some metadata information (such as the phase encoding direction) as `RECOMMENDED`. So, as far as this PR is concerned, we are not in breach of the spec.

Besides, as far as Clinica is concerned, this change has no incidence. In other terms, the superfluous acquisition entity was neither parsed nor used in our DWI pipelines.

Now the downside of not providing this piece of metadata is that *other* processing pipelines not (yet) in Clinica may require such information, since it is qualified as recommended afterall. By not providing these recomended metadata, we are limitting the potential range of application for derivatives, whether inside or outside of Clinica.

That's beyond the scope of this PR though, and should be addressed as part of a larger effort to expose more of these metadata from all of our converters.

omar-rifai:
If you have a strong feeling about not implementing this here then so be it but personally I feel like removing this entity in here without adding the info elsewhere is somewhat of a regression. At any rate, this is not urgent as you have pointed out as it is not used elsewhere.

ghisvail:
> Personally I feel like removing this entity without adding the info elsewhere is somewhat of a regression.

There is no direct correspondance from `acq-axial` to the corresponding metadata (assuming this ever was the intent). In the end, there is no information lost beyond a tag in the filename.

> If you have a strong feeling about not implementing this here, then so be it.

My intent was only to remove this acquisition label, which we don't know the meaning of and have no use for. It would be more [work](https://mrtrix.readthedocs.io/en/latest/concepts/pe_scheme.html#fixed-phase-encoding) to find the adequate value for `PhaseEncodingDirection`, and other recommended metadata, for each dataset.

Hence why I am suggesting a different PR for it.

omar-rifai:
> There is no direct correspondance from `acq-axial` to the corresponding metadata (assuming this ever was the intent). In the end, there is no information lost beyond a tag in the filename.

If that is the case then sure, it's a different story ! The comment above was relative to entities for which we do find a corresponding metadata.

omar-rifai:
Actually, it seems that the ".json" metadata is already generated by the converter:
```
{
	"Manufacturer": "GE",
	"ManufacturersModelName": "DISCOVERY_MR750",
	"AcquisitionNumber": 1,
	"InstitutionName": "UIRR",
	"DeviceSerialNumber": "6cnQ6wuFVm2vRB7ybMfpFTk",
	"SoftwareVersions": "27_LX_MR_Software_release:DV25.1_R03_1802.a",
	"ScanningSequence": "EP_SE",
	"SequenceVariant": "NONE",
	"SeriesDescription": "Axial_DTI",
	"BodyPartExamined": "BRAIN",
	"ProtocolName": "Axial_DTI",
	"ImageType": ["ORIGINAL", "PRIMARY", "OTHER"],
	"AcquisitionTime": "10:35:14.000000",
	"MagneticFieldStrength": 3,
	"FlipAngle": 90,
	"EchoTime": 0.0604,
	"RepetitionTime": 7.8,
	"PhaseEncodingLines": 116,
	"ConversionSoftware": "dcm2niix",
	"ConversionSoftwareVersion": "v1.0.20170624 Clang7.3.0"
}
```

omar-rifai (approved):

alexandreroutier:
Hi!

I believe that `_acq-axial` key-value was used to stress that DWI data in ADNI came from "Axial DTI" protocol or something named like this (cf `SeriesDescription` in JSON file from @omar-rifai's comment).

I am ccing @nburgos to confirm or not but I remember that there exists another protocol in recent ADNI flavours. In particular, there were some patients which underwent two DWI protocols. In the end, the "majority"  protocol was converted but we used this key to keep a trace of it (and possibly convert the other protocol in the future?).

See e.g. https://adni.loni.usc.edu/methods/documents/ the document [ADNI3 MRI Scanner Protocols](https://adni.loni.usc.edu/methods/documents/) 

I think this PR reveals the fact there is an ambiguity between information on the Clinica Wiki:
> DWI We select images containing 'DTI' in the sequence name, that are not multiband, processed nor enhanced.

and the resulting BIDS files with  `_acq-axial` embedded.

nburgos:
> I am ccing @nburgos to confirm or not but I remember that there exists another protocol in recent ADNI flavours. In particular, there were some patients which underwent two DWI protocols. In the end, the "majority" protocol was converted but we used this key to keep a trace of it (and possibly convert the other protocol in the future?).

I remember the same thing but don't have the details in mind. It's good to keep trace of what was actually converted but I guess we can do it with the json and by clarifying the documentation.

ghisvail:
> I believe that `_acq-axial` key-value was used to stress that DWI data in ADNI came from "Axial DTI" protocol or something named like this (cf `SeriesDescription` in JSON file from @omar-rifai's comment).

Thank you for confirming that 👍 

> I remember that there exists another protocol in recent ADNI flavours.

Indeed, both ADNI2/GO and ADNI3 declare their respective DWI protocols.

> In the end, the "majority" protocol was converted but we used this key to keep a trace of it (and possibly convert the other protocol in the future?).

Since @omar-rifai confirmed the presence of this piece of information in the JSON metadata, we have not lost much.

Should other DWI protocols be converted in the future, then we would revisit the use of the acquisition entitiy for file distinction.

omar-rifai:
Can be re-based with dev for passing CI

PR #624, opened 2022-03-18
Fix PET entity parsing in iotools

ghisvail:
Entity parsing in some iotools rely on positional location of tokens, which is very fragile. This PR adjusts the location to the new standards adopted for PET, but a long term solution should be to resort to entity prefix parsing instead.

Requires synchronization of data for CI to pass.


ghisvail:
I also took the opportunity to fix a few easy warnings reported by PyCharm.

omar-rifai (approved):

PR #625, opened 2022-03-23
BIDS-Compliance: rename `subjects` dir with name of pipeline

omar-rifai:
This PR's purpose is to rename the root directory of the derivative outputs from `subjects` into `<pipeline-name>` which is what is recommended for BIDS derivatives. 

As the "subjects" directory is hard-coded in several places in the code, it would require to:

- [x] use the pipeline name in `container_from_filename()`
- [ ] replace hard-coded instances of "subjects" in `clinica/utils/inputs.py`  determine_caps_or_bids. As this is used by `clinica_file_reader` moving forward is delicate. Open to suggestions on this.
- [ ] replace "subjects" in pipelines not handled by `container_from_filename()`(e.g t1-freesurfer, pet-surface`
- [ ] Update the doc to reflect these changes

### DIFF
--- a/clinica/pipelines/dwi_dti/dwi_dti_pipeline.py
+++ b/clinica/pipelines/dwi_dti/dwi_dti_pipeline.py
@@ -132,12 +132,14 @@ class DwiDti(cpe.Pipeline):
         # Find container path from filename
         container_path = npe.Node(
             nutil.Function(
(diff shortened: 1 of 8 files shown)


omar-rifai:
Effort stalled for investigating pydra migration.

PR #626, opened 2022-03-24
Disable macos non-regression tests in Jenkinsfile_nonregression

omar-rifai:
The non-regression tests fails because of a environment set-up problem in macos. Last run of scheduled tests failed to finish on linux as well. This PR comments out all the tests on macos temporarily to pinpoint the issue.


mdiazmel (approved):
LGTM! I'll merge!

PR #627, opened 2022-03-28
Modify time before issues considered stale

omar-rifai:


ghisvail:
A year is a bit much. May I suggest to increase both counts to 28 days instead? That means an issue would be considered stale after a month without activity, and be automatically closed after another month.

Otherwise, we might as well get rid of this altogether.

omar-rifai:
> A year is a bit much. May I suggest to increase both counts to 28 days instead? That means an issue would be considered stale after a month without activity, and be automatically closed after another month.
> 
> Otherwise, we might as well get rid of this altogether.

from the current state of the issues, some date more than a year back and that still haven't been addressed. The rationale behind the 365 days is that the hook would serve a  reminder for us to take a look at those issues again. Maybe something in between would work ? 100 days?

ghisvail:
> from the current state of the issues, some date more than a year back and that still haven't been addressed. 

Indeed, hence why the following start date is used `start-date: '2022-01-01T00:00:00Z'`. Those past issues will not be considered by the stale bot.

> The rationale behind the 365 days is that the hook would serve a reminder for us to take a look at those issues again.

I believe that an issue opened a year ago is _very_ unlikely to be ever taken care of. Both the devs and users have moved on for a while past that point. There are a few projects I contributed an issue to which had a stale config this large, and my initial reaction when I got the reminder email months after was 🙄 whatever.

 > Maybe something in between would work ? 100 days?

The question is to find the right amount from which both devs and users are happy with a reminder. Could be 30, 60 maybe 90 days. Definitely not a year in my opinion.

ghisvail (approved):

PR #629, opened 2022-03-29
Skip empty dirs in compare_folders()

omar-rifai:
Quick fix for non-regression tests to skip empty directories when comparing


omar-rifai:
> Thanks @omar-rifai .
> 
> Can you add a word on the docstring of this function (`tree`) to precise the new behavior on empty folders?
> 
> Potentially, there can be some border effects on the test for the converters (where the `compare_folders` function is also used. To verify during the CI.

Done for the docstring. I'll run the `scheduled_tests` tonight to validate that all goes well.

mdiazmel (approved):

PR #633, opened 2022-03-31
add na=False to str.contains() to handle nan values

### LINKED ISSUE #632: Error with adni-to-bids
What is happening?

This is the first time for my team and me to use the clinica.

We have downloaded sMRI baseline data from ADNI（When I downloaded data from ADNI, I choosed ADNI Baseline, ADNIGO Screening MRI ADNI2 Screening MRI-New Pt and ADNI3 Initial Visit-Cont Pt）and try to get data from baseline. 

We encountered an error while running `clinica convert adni-to-bids $ADNI_PATH $CLINICAL_DATA_DIRECTORY $BIDS --modalities T1`
```
2022-03-31 16:10:08,969:INFO:	Reading clinical data file: VITALS.csv
Traceback (most recent call last):
  File "/home/dell/miniconda3/envs/clinicaEnv/bin/clinica", line 8, in <module>
    sys.exit(main())
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/clinica/cmdline.py", line 89, in main
    cli()
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/clinica/iotools/converters/adni_to_bids/adni_to_bids_cli.py", line 70, in cli
    adni_to_bids.convert_clinical_data(
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/clinica/iotools/converters/adni_to_bids/adni_to_bids.py", line 126, in convert_clinical_data
    adni_utils.create_adni_sessions_dict(
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/clinica/iotools/converters/adni_to_bids/adni_utils.py", line 720, in create_adni_sessions_dict
    df_subj_session = df_subj_session[
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/pandas/core/frame.py", line 3495, in __getitem__
    if com.is_bool_indexer(key):
  File "/home/dell/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/pandas/core/common.py", line 144, in is_bool_indexer
    raise ValueError(na_msg)
ValueError: Cannot mask with non-boolean array containing NA / NaN values
```
When I go to the `adni_utils.py`, it shows

```
    # Nv/None refer to sessions whose session is undefined. "sc" is the screening session with unreliable (incomplete)
    # data.
    df_subj_session = df_subj_session[
        df_subj_session.session_id.str.contains("ses-M\d+")
    ]
```
Should I download more data  or do something else to avoid the error?

Hopefully to receive the reply

Thanks.

omar-rifai:
Closes #632

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -733,7 +733,7 @@ def create_adni_sessions_dict(
     # Nv/None refer to sessions whose session is undefined. "sc" is the screening session with unreliable (incomplete)
     # data.
     df_subj_session = df_subj_session[
-        df_subj_session.session_id.str.contains("ses-M\d+")
+        df_subj_session.session_id.str.contains("ses-M\d+", na=False)
     ]
     write_adni_sessions_tsv(df_subj_session, bids_subjs_paths)
 


ghisvail:
This solves the issue but wouldn't it be more robust to just drop all session with a NaN value as session ID in the first place? They don't make any sense and may be the source for other bugs down the line.

omar-rifai:
> This solves the issue but wouldn't it be more robust to just drop all session with a NaN value as session ID in the first place? They don't make any sense and may be the source for other bugs down the line.

I guess it could be more explicit but it boils down to that as we are filtering the data frame based on that mask, no?

ghisvail:
> I guess it could be more explicit

Which is always a good thing.

> but it boils down to that as we are filtering the data frame based on that mask, no?

The remedy is the same, agreed, but the cause remains unknown.

First, we should ensure that `df_subj_session` **never** contains rows with either of its IDs set to `n/a`. In the context of this function where `df_subj_session` is constructed by iterative accumulation, that means `df_filtered` brings the n/a values.

Now, I suspect our n/a value were generated from [here](https://github.com/aramis-lab/clinica/blob/f039a9e572cef2453d3509e0cec0fa5c88da7395/clinica/iotools/converters/adni_to_bids/adni_utils.py#L710). Which brings the real issue: what is the cause for the missing mapping from visit to session for this particular instance?

omar-rifai:
> > but it boils down to that as we are filtering the data frame based on that mask, no?
> 
> The remedy is the same, agreed, but the cause remains unknown.
> 

The cause is simply the unreliability of the input data from `adni` the few exceptions we handle in `get_visit_id()` were not fully documented. 

> Now, I suspect our n/a value were generated from [here](https://github.com/aramis-lab/clinica/blob/f039a9e572cef2453d3509e0cec0fa5c88da7395/clinica/iotools/converters/adni_to_bids/adni_utils.py#L710). Which brings the real issue: what is the cause for the missing mapping from visit to session for this particular instance?

I guess we could change the default return values in [`get_visit_id()`](https://github.com/aramis-lab/clinica/blob/f039a9e572cef2453d3509e0cec0fa5c88da7395/clinica/iotools/converters/adni_to_bids/adni_utils.py#L627). Otherwise I'll be happy for you to suggest a modification you seem fit and I'll merge.

ghisvail:
> Otherwise I'll be happy for you to suggest a modification you seem fit and I'll merge.

Perhaps a better alternative is to add `df_filtered.dropna(subset=["session_id"], inplace=True)` after [line 716](https://github.com/aramis-lab/clinica/blob/f039a9e572cef2453d3509e0cec0fa5c88da7395/clinica/iotools/converters/adni_to_bids/adni_utils.py#L716).

ghisvail:
> The cause is simply the unreliability of the input data

I'd say it's rather our lack of robustness that is at fault here.

The entire web works with unreliable data.

omar-rifai:
I was referring to the cause of `NaN` as session ids. If you think we shouldn't have added them to the dataframe in the first place then I'd argue that your fix with a `dropna()` does not solve the issue. 

I feel like we're splitting hair here. If you want to push ahead with your solution, please provide the commit and I'll merge otherwise I feel like the initial solution does the job.

ghisvail:
> I was referring to the cause of NaN as session ids.

The mapping line 711 may not be exhaustive enough, which would be cause for getting NaN values. It could be done on purpose for filtering purposes, or be completely accidental. Who knows.

Another source for NaN could be buggy code in `update_sessions_df`, which is called right after with `df_filtered`.

> I'd argue that your fix with a dropna() does not solve the issue.

It solves the non exhaustive mapping issue described above. After `dropna`, we are now guaranteed that there are no rows with NaN session_id after mapping. If we are still having this issue, then we know *for sure* the problem lies further downstream, most likely in `update_sessions_df`. Then, we could fix this as well.

Since this code is full of side-effects, it is better from a maintenance perspective to provide as much guarantees as we can, as early as we possibly can. We could just apply the `dropna` right before where the error happens (which is similar to what you propose) with the downside of shadowing multiple root causes upstream.

> I feel like we're splitting hair here.

I am just giving you a perspective from a data engineering mindset. If I failed to convince you, that's fine really.

omar-rifai:
> The mapping line 711 may not be exhaustive enough, which would be cause for getting NaN values. It could be done on purpose for filtering purposes, or be completely accidental. Who knows.
> 

That's what I suggested changing when you first mentioned the cause. 

> I guess we could change the default return values in [get_visit_id()](https://github.com/aramis-lab/clinica/blob/f039a9e572cef2453d3509e0cec0fa5c88da7395/clinica/iotools/converters/adni_to_bids/adni_utils.py#L627)


> Another source for NaN could be buggy code in `update_sessions_df`, which is called right after with `df_filtered`.
> 

I'd argue that by reading the code we can simple see that no, it's not the case.


> > I'd argue that your fix with a dropna() does not solve the issue.
> 
> It solves the non exhaustive mapping issue described above. After `dropna`, we are now guaranteed that there are no rows with NaN session_id after mapping. If we are still having this issue, then we know _for sure_ the problem lies further downstream, most likely in `update_sessions_df`. Then, we could fix this as well.
> 

Same comment as above. The issue's "source" is `get_visit_id()` which is not exhaustive.


> Since this code is full of side-effects, it is better from a maintenance perspective to provide as much guarantees as we can, as early as we possibly can. We could just apply the `dropna` right before where the error happens (which is similar to what you propose) with the downside of shadowing multiple root causes upstream.
> 

Agreed. And from my comments above, my code has the same effect as what you suggested. I'd be happy to discuss it further during a call if need be.

> > I feel like we're splitting hair here.
> 
> I am just giving you a perspective from a data engineering mindset. If I failed to convince you, that's fine really.

I appreciate your input but _again_  my initial proposal was to change The mapping line 711  which indeed is the initial construction of the data frame. If all the discussion boils down to this than we could have agreed on it from the start. That's all I'm saying.

ghisvail (approved):

ghisvail:
> I appreciate your input but again my initial proposal was to change The mapping line 711 which indeed is the initial construction of the data frame.

Actually, lines 706 and 707 are where this dataframe gets constructed, not 711.

 > If all the discussion boils down to this than we could have agreed on it from the start.

It's not but I consider it my failure not managing to explain this better.

> That's all I'm saying.

In hindsight, that code is so full of side effects that I am not sure providing further guarantees really is worth fighting for. I approved the PR to avoid further frustration on your end.

omar-rifai:
> > I appreciate your input but again my initial proposal was to change The mapping line 711 which indeed is the initial construction of the data frame.
> 
> Actually, lines 706 and 707 are where this dataframe gets constructed, not 711.
> 

I meant the part of interest to this conversation (namely the part where we map the session ids). 

> In hindsight, that code is so full of side effects that I am not sure providing further guarantees really is worth fighting for. I approved the PR to avoid further frustration on your end.

I'm happy to have a constructive conversation on this. The frustration came from nitpicking on comments instead  of moving forward with a commit as I suggested you'd propose. 
I'm completely open for you to have a go at it.

omar-rifai:
Closing as discussed. Will be handled by a different PR.

PR #635, opened 2022-04-04
Remove superfluous code in T1Linear pipeline

ghisvail:
This PR removes superflous node connections and substitutions from a past behaviour where the output of the intermediate bias field corrected image used to be saved as an output of the workflow. Now that this is no longer the case (not even with a parameter), it is worth getting rid of the dead code which only adds confusion to the process of transitionning the output of T1Linear to a BIDS compliant format.


mdiazmel (approved):
LGTM!

ghisvail:
Merged since CI issues are unrelated to this PR.

PR #639, opened 2022-04-07
Enh output node

omar-rifai:


clinica/pipelines/t1_linear/t1_linear_utils.py
```diff
@@ -1,3 +1,32 @@
+from clinica.utils.exceptions import ClinicaParserError
+
+
+def extract_source_entities(ref_file):
```
NicolasGensollen, line 4: I feel like these functions are not specific to the `t1_linear` pipeline, they extract entities from bids file name and vice-versa. They could probably be shared across pipelines. WDYT?

clinica/pipelines/t1_linear/t1_linear_utils.py
```diff
@@ -1,3 +1,32 @@
...
+    import pathlib
+
+    # TODO: smarter entities extraction
+    source_entities = ("_").join(pathlib.Path(ref_file).stem.split("_")[:-1])
+    ext = "".join(pathlib.Path(ref_file).suffixes)
+
+    return source_entities, ext
+
+
+def construct_derivative_entities(pipeline_entities, pipeline_parameters):
```
NicolasGensollen, line 14: Could also be worth adding type hints for consistency (even though the other functions of this file do not seem to have them...).

clinica/pipelines/t1_linear/t1_linear_utils.py
omar-rifai: Indeed, that is the goal ultimately. This currently is sort of a POC to see if it works :)

clinica/pipelines/t1_linear/t1_linear_utils.py
omar-rifai: From what I remember, type hints did not couple well with Nipype Nodes - I can check again.

omar-rifai:
Effort stalled for investigating Pydra route.

PR #640, opened 2022-04-08
[INFRA] Add unittests to CI

NicolasGensollen:
Now that we have (some) unit tests, this PR proposes : 

- to fix a typo in a filename (need to check that this is not breaking something expecting the existence of this file down the line though...)
- to add unit tests to the CI


ghisvail:
The test suite passes, but the logs suggest we have resource leaks:

```
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name=0 mode='r' encoding='UTF-8'>
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name=1 mode='w' encoding='UTF-8'>
```

PR #642, opened 2022-04-19
Remove vendored mrtrix3 interfaces

ghisvail:
Clinica used to vendor some of nipype's interfaces because the base nipype version supported by Clinica was frozen.

Now that Clinica supports the latest version of nipype, this PR removes the vendored code and use nipype directly.


NicolasGensollen (approved):
LGTM, thanks @ghisvail !

Linking #492 as it was similarly removing vendors interfaces when Nipype `1.7.0` came out.
Also linking Nipype's PR enabling this update : https://github.com/nipy/nipype/pull/3445

PR #644, opened 2022-04-28
[DOC] Fix documentation on providing covariates for `statistics_surface` pipe…

NicolasGensollen:
The current documentation recommends providing covariates through `--covariates 'age sex'` which results in an error when evaluating the design matrix down the line. This PR updates the docs to recommend using `-c age -c sex`.


omar-rifai (approved):
Great, thanks!

PR #645, opened 2022-04-29
Update unitest with swi folder in bidsignore

omar-rifai:
After changes in UKB to `.bidsignore` the unittest failed on `test_write_modality_agnostic_files`. This PR updates the test with `swi/` that was added to `.bidsignore` writer.


NicolasGensollen (approved):
LGTM, Thanks for fixing this @omar-rifai !

PR #647, opened 2022-05-03
Add b-spline alternative value

sophieloiz:
Modify the b-spline fitting distance according to the modalty (T1W or FLAIR)


omar-rifai:
LGTM, thanks :)

PR #650, opened 2022-05-13
Use Gunzip interface directly

ghisvail:
This PR retires our old wrapper function for decompressing nifti files and uses the `Gunzip` interface directly instead. It remains functionally equivalent, since the wrapper function was essentially calling this interface underneath.

As far as compression is concerned, I filed this [issue](https://github.com/nipy/nipype/issues/3471) upstream to collect opinions on whether it would make sense to extend this interface to produce compressed nifti files. I believe it would.


omar-rifai:
Thanks ! will merge. But as discussed offline, let's keep in mind - for other custom functions- the drawbacks/benefits tradeoff of adding new Nipype interfaces in the midst of migrating the project to Pydra.

PR #653, opened 2022-05-16
Add entry point for the FLAIR documentation

MatthieuJoulot:


ghisvail (changes requested):
Needs an update [here](https://github.com/aramis-lab/clinica/blob/dev/docs/index.md) too.

I'd suggest to group FLAIR and T1 together in the anatomical pipeline section.

Issue #294, opened 2021-06-23
Replace the class loader functionality


omar-rifai:
A class loader functionality which was historically in Clinica (but ununsed up until now) has been removed in [PR275](https://github.com/aramis-lab/clinica/pull/275). We could replace it with a more idiomatic 'Click' mechanism if it is needed.

ghisvail:
I believe third-party extension of Clinica is an exciting topic, but won't come back in the form of a class loader. Instead, we would likely come up with a proper plugin architecture for the relevant parts of Clinica which may benefit from runtime extensions.

PR #658, opened 2022-05-25
Initialize Pydra Engine - I/O interfaces

omar-rifai:
**Update**: Currently the features this PR are

- [x] A Pipeline class for a uniform objects across clinica pipelines 
- [x] An `input` interface** created dynamically based on the `input_spec` of the core workflow 
- [x] An `output` interface** created dynamically based on the `output_spec` of the core workflow 
- [x]  Add (hidden) pydra version of t1-linear. The command is invocable with  `clinica run pydra-t1-linear <input-dir> <output-dir>`
- [x] Add unit tests for the different functions

Important things to improve:
- Definition of splitting criteria if multiple inputs are available (define placeholder for this information in the Pipeline class)
- Placeholder for remote execution specification
- Design of a plugin architecture  

 ** Currently the interfaces serve as POC but more in-depth implementation to be done in different PR


ghisvail:
I am bit worried that a PR may not be the right medium to discuss this design decision. The risk is to introduce early bias towards implementation details (with the code associated to this PR) instead of widening our reach for alternative blueprints.

> The idea is to agree on the main features of the engine here before implementing any pipelines.

As far as this proposal is concerned, by defining a Clinica pipeline as a subclass of `pydra.Workflow`, I believe we would end up suffering from the same limitations as we currently do with the current design with `nipype` down the road.

Starting with excessive coupling with the underlying pipeline implementation, which is the main reason why Clinica is so difficult to extend today. Most of our pipelines break [SOLID principles](https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design) one way or another, which is a sign to me that the **is a** relationship between Clinica's pipelines and nipype / pydra workflows introduced by subclassing should be challenged.

omar-rifai:
In order to move forward, a list of required features (below) were defined with @ghisvail. The design of the new engine should be flexible enough to accommodate these requirements.

### Domain Requirements
- BIDS compliance for input and output (custom BIDS data reader/ BIDS data writer)
- Provenance information in the pipelines
- Possibility to define/set different execution environments (local/remote/container..)

### Technical Requirements
- Allow pipelines to be extendable in engine (can easily add nodes before and after) 
- Make pipelines (pydra-) engine agnostic to the extent possible (hide pydra specifics)
- Decoupling of domain (image processing arguments) and technical pipeline parameters

ghisvail:
Perhaps we could also add "support for third-party pipelines" to the DR and "Design a plugin architecture to support pipelines maintained outside of Clinica" to the TR to address #294?

clinica/pipelines/engine_pydra.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self._input_workflow.set_output(
+            [
+                (field, getattr(self._input_workflow.bids_reader_task.lzout, field))
+                for field in data_keys
+            ]
+        )
+
+    def run(self):
+        with Submitter(plugin="cf") as submitter:
```
ghisvail, line 97:
```suggestion
    def run(self):
        # FIXME: Allow parameterization of submitter.
        with Submitter(plugin="cf") as submitter:
```

test/unittests/pydra/test_engine.py
```diff
@@ -0,0 +1,92 @@
+from distutils import core
```
ghisvail, line 1: I wonder whether we are testing the right thing here.

From what I read, we are saying the implementation works if nodes named `bids_reader_task` or `bids_writer_task_0` are present anywhere in the workflow. In my opinion, we are testing implementation details rather than actual properties of what the decorator achieves.

If I change the name of the task in the implementation, it does affect that I can still read or write BIDS data in the workflow, yet the test would fail.

An alternative could be to test the reader with a task that simply accumulates paths received sequentially as input from the BIDS reader and check that its results include all paths expected from the BIDS dataset provided as input.

Same for the writer. Making a task that yields a bunch of arbitrary files and checking that the writer wrote all of them to their expected location.

What do you think?

ghisvail (changes requested):

test/unittests/pydra/test_engine.py
omar-rifai: Yes I agree and I actually hesitated to write out tests for these functions. In reality, since the interfaces are bound to change (not within the scope of this PR). the only reason these tests exists is to preserve the good practice of having unit tests.
Once the `bids_reader_task` and the `bids_writer_task` are actually written to what we want them to achieve, what you are suggesting will make sense.

omar-rifai: I updated the description of the PR to reflect that point.

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+
+    pipeline.add(input_workflow)
+
+    # connect input workflow to core workflow
+
+    for field in list_core_inputs:
+        read_data = getattr(input_workflow.lzout, field)
+        setattr(core_workflow.inputs, field, read_data)
+
+    return field
```
ghisvail, line 68: "Local variable 'field' might be referenced before assignment"

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+            for field in data_keys
+        ]
+    )
+    return input_workflow
+
+
+def build_output_workflow(
+    pipeline: Workflow, core_workflow: Workflow, output_dir: PurePath
+) -> None:
+    """Example of an output workflow.
```
ghisvail, line 97: Missing parameters in docstring
Unexpected parameter pipeline in docstring

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for field in list_core_inputs:
+        read_data = getattr(input_workflow.lzout, field)
+        setattr(core_workflow.inputs, field, read_data)
+
+    return field
+
+
+def add_input_task(input_workflow: Workflow, query_dict) -> Workflow:
+    """
```
ghisvail, line 72: Missing parameters in docstring
Unexpected parameter pipeline in docstring

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+
+def clinica_io(func):
+    """
+    Prepend an input workflow (BIDS reader) and
+    append an output_workflow (BIDS writer)
+    to any Pydra workflow
+    """
+
+    @functools.wraps(func)
+    def run_wrapper(name: str, input_dir: PathLike, output_dir: PathLike, **kwargs):
```
ghisvail, line 20: Parameter 'kwargs' value is not used

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+            input_spec=["input_dir"],
+            input_dir=input_dir,
+        )
+
+        split_key = build_input_workflow(pipeline, core_workflow)
+
+        # TODO: define condition on split if multiple fields
+        pipeline.add(core_workflow.split(split_key))
+
+        build_output_workflow(pipeline, core_workflow, output_dir)
```
ghisvail, line 35: Expected type 'PurePath', got 'PathLike' instead

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
...
+def list_out_fields(wf: Workflow) -> list:
+    """
+    :return: list of workflow "wf" fields in output_spec
+    """
+    return [x[0] for x in wf.output_spec.fields if not x[0].startswith("_")]
+
+
+def list_in_fields(wf: Workflow) -> list:
+    """
+    :return: list of worfklow wf fields in input_spec
```
ghisvail, line 22: typo

```suggestion
    :return: list of workflow wf fields in input_spec
```

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    return [x[0] for x in wf.input_spec.fields if not x[0].startswith("_")]
+
+
+def bids_query(keys: list) -> dict:
+    """
+    keys: a list of BIDS items to query
+    returns: a dictionary formated for input to BIDSDataGrabber()
```
ghisvail, line 30: typo

```suggestion
    returns: a dictionary formatted for input to BIDSDataGrabber()
```

clinica/pydra/interfaces.py
```diff
@@ -0,0 +1,32 @@
...
+    print("printing core vars in output_workflow", output_file)
+    print("should go to location", output_dir)
+
+    return "dummy_output_string"
+
+
+def bids_reader(query_bids: dict, input_dir: PurePath):
+    """
+    :query_bids: input to BIDSDataGrabber (c.f https://nipype.readthedocs.io/en/latest/api/generated/nipype.interfaces.io.html#bidsdatagrabber)
+    :input_dir: the BIDS input directory
```
ghisvail, line 24: numpydoc

clinica/pydra/t1_linear/t1_linear.py
```diff
@@ -0,0 +1,117 @@
+from multiprocessing import Pipe
```
ghisvail, line 1: unused import

```suggestion
```

test/unittests/pydra/test_engine.py
```diff
@@ -0,0 +1,92 @@
...
+import pip
+import pytest
+from pydra import Workflow
+from pathlib import Path, PurePath
+from os import PathLike, path
```
ghisvail, line 6: unused imports

```suggestion
import pytest
from pathlib import Path, PurePath
from os import PathLike, path
```

test/unittests/pydra/test_engine.py
```diff
@@ -0,0 +1,92 @@
...
+    input_workflow.inputs.input_dir = Path(CURRENT_DIR) / "data"
+
+    add_input_task(input_workflow, bids_query)
+
+    assert "bids_reader_task" in [x.name for x in input_workflow.nodes]
+    return
+
+
+def test_build_output_workflow(pipeline: Workflow, core_workflow: Workflow):
+    from clinica.pydra.engine import build_input_workflow
```
ghisvail, line 82: Unused import as a result of the commented code. Either uncomment the statement or delete it and remove the unused import.

test/unittests/pydra/test_engine.py
```diff
@@ -0,0 +1,92 @@
...
+
+@pytest.fixture
+def core_workflow():
+    wf = Workflow("smoothing_t1w", input_spec=["T1w"])
+    wf.add(
+        smooth_image(
+            name="smooth_image",
+            interface=smooth_image,
+            input_image=wf.lzin.T1w,
+        )
```
ghisvail, line 48: I wonder whether we really need to rely on `nilearn` for testing?

We could just echo the input image to the working directory and consider this copy a derivative for testing purposes, no?

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+            (field, getattr(input_workflow.bids_reader_task.lzout, field))
+            for field in data_keys
+        ]
+    )
+    return input_workflow
+
+
+def build_output_workflow(
+    pipeline: Workflow, core_workflow: Workflow, output_dir: PurePath
+) -> None:
```
ghisvail, line 96: typing

```suggestion
) -> Workflow:
```

omar-rifai:
@ghisvail please let me know when the review is over so that I can take the time to go over all the points.

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    bids_keys_available = {
+        "T1w": {"datatype": "anat", "suffix": "T1w", "extension": [".nii.gz"]}
+    }
+
+    return {key: bids_keys_available[key] for key in keys}
+
+
+def run(wf):
```
ghisvail, line 39: Missing docstring and typing information

clinica/pydra/t1_linear/t1_linear_cli.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@click.command(name=pipeline_name, hidden=True)
+@cli_param.argument.bids_directory
+@cli_param.argument.caps_directory
+def cli(
+    bids_directory: str,
+    caps_directory: str,
+) -> None:
+    """Affine registration of Flair images to the MNI standard space (Pydra engine)."""
```
ghisvail, line 17:
```suggestion
    """Affine registration of T1w images to the MNI standard space (Pydra engine)."""
```

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
+from pydra import Submitter, Workflow
+
+
+def list_keys(query_bids: dict) -> list:
+    """
+    :query_bids: query dictionary to BIDSDataGrabber
+    :return: IDS of queried elements
+    """
+
+    return list(query_bids.keys())
```
ghisvail, line 10: I don't think this utility is particularly useful. It's not specific to BIDS (i.e. it works on any dictionary) and it's used once in the code base (`add_input_task`).

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def list_out_fields(wf: Workflow) -> list:
+    """
```
ghisvail, line 14: numpydoc

clinica/pydra/engine_utils.py
```diff
@@ -0,0 +1,45 @@
...
+def list_in_fields(wf: Workflow) -> list:
+    """
+    :return: list of worfklow wf fields in input_spec
+    """
+    return [x[0] for x in wf.input_spec.fields if not x[0].startswith("_")]
+
+
+def bids_query(keys: list) -> dict:
+    """
+    keys: a list of BIDS items to query
```
ghisvail, line 29: numpydoc

clinica/pydra/engine.py
```diff
@@ -0,0 +1,121 @@
...
+    :return: The output workflow.
+    """
+
+    output_attrs = []
+
+    for i, field in enumerate(pu.list_out_fields(core_workflow)):
+
+        pipeline.add(bids_writer(name="bids_writer_task_" + str(i)))
+
+        writer_task = getattr(pipeline, "bids_writer_task_" + str(i))
```
ghisvail, line 109: I wonder whether it would be more preferable to use the field name instead of its index from enumeration. It would make introspection easier, i.e. it's easier to know what a task named `write_t1w_corrected` compared to `write_1`.

ghisvail (changes requested):
Some more comments on the implementation side

ghisvail:
> @ghisvail please let me know when the review is over so that I can take the time to go over all the points.

You can start with this batch.

A more general comment is that I guess we are losing the ability to restrict processing to a custom list of subjects and sessions, am I correct? Just to be sure that's on purpose if that's the case.

omar-rifai:
> > @ghisvail please let me know when the review is over so that I can take the time to go over all the points.
> 
> You can start with this batch.
> 

I'd rather wait for your review to be ready.

> A more general comment is that I guess we are losing the ability to restrict processing to a custom list of subjects and sessions, am I correct? Just to be sure that's on purpose if that's the case.

I'm not sure I understand what you mean by losing as we are starting from scratch? If comparing to the implementation we have in Nipype yes, a lot of the features have not been implemented in this initial PR as agreed upon.

test/unittests/pydra/test_engine.py
omar-rifai: both work. Currently as this does not add a burden on our dependencies I don't really see the point in changing. But if we see further reasons down the line, we can reconsider.

clinica/pydra/engine.py
omar-rifai: Good point, changed.

PR #659, opened 2022-05-27
Fix multiple issues in data_handling module

ghisvail:
I came across some warnings in my IDE while browsing code.

This PR fixes issues with typos and incomplete docstrings, and removes some dead code. There is also the `Variable referenced before assignment` which is a potentially harmful bug if this code path ever gets triggered at runtime.


NicolasGensollen (approved):
LGTM, thanks @ghisvail ! 👍

PR #662, opened 2022-05-30
Parallel file reader

14thibea:
Hello :)

I am currently using the `extract` function of ClinicaDL which relies on `clinica_file_reader`, and on a large data set this takes forever.
This is why I propose to parallelize the `clinica_file_reader` with `joblib`.
I also changed the version of black used for the environment installation of pre-commit as I encountered the same error as in https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click (now it works).

Tell me if you need more explanations or if you think this should be done differently.
Best,

### DIFF
--- a/clinica/utils/inputs.py
+++ b/clinica/utils/inputs.py
@@ -198,7 +198,10 @@ def clinica_file_reader(
                                                         file
         raise_exception: if True (normal behavior), an exception is raised if errors happen. If not, we return the file
                         list as it is
-        n_procs: number of cores used to run in parallel.
+        n_procs: int, optional
+            Number of cores used to run in parallel with Joblib.
+            If set to 0, subjects and sessions will be processed sequentially.
+            Default=0.
 
     Returns:
          list of files respecting the subject/session order provided in input,
--- a/clinica/utils/inputs.py
+++ b/clinica/utils/inputs.py
@@ -181,7 +181,7 @@ def clinica_file_reader(
     input_directory,
     information,
     raise_exception=True,
-    n_procs=0,
+    n_procs: int = 1,
 ):
     """Read files in BIDS or CAPS directory based on participant ID(s).
 
@@ -199,9 +199,8 @@ def clinica_file_reader(
         raise_exception: if True (normal behavior), an exception is raised if errors happen. If not, we return the file
                         list as it is
         n_procs: int, optional
-            Number of cores used to run in parallel with Joblib.
-            If set to 0, subjects and sessions will be processed sequentially.
-            Default=0.
+            Number of cores used to fetch files in parallel (defaults to 1).
+            If set to 1, subjects and sessions will be processed sequentially.
 
     Returns:
          list of files respecting the subject/session order provided in input,
@@ -325,7 +324,7 @@ def clinica_file_reader(
     if len(subjects) == 0:
         return [], ""
 
-    if n_procs > 0:
+    if n_procs > 1:
         # results is the list containing the results
         shared_results = manager.list()
         # error is the list of the errors that happen during the whole process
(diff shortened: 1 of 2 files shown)


clinica/utils/inputs.py
```diff
@@ -193,6 +198,7 @@ def clinica_file_reader(
                                                         file
         raise_exception: if True (normal behavior), an exception is raised if errors happen. If not, we return the file
                         list as it is
+        n_procs: number of cores used to run in parallel.
```
NicolasGensollen, line 201: I would add a bit more information.
Maybe something like this:

```suggestion
        n_procs: int, optional
            Number of cores used to run in parallel with Joblib.
            If set to 0, subjects and sessions will be processed sequentially.
            Default=0.
```

clinica/utils/inputs.py
```diff
@@ -251,14 +257,40 @@ def clinica_file_reader(
...
+
     from clinica.utils.exceptions import (
         ClinicaBIDSError,
         ClinicaCAPSError,
         ClinicaException,
     )
 
+    manager = Manager()
+
+    def find_sub_ses_pattern(input_directory, sub, ses, error_encountered, results):
```
NicolasGensollen, line 273: Is there a reason you didn't extract this function out of `clinica_file_reader`?
Also, I think adding a very short docstring could be useful.

NicolasGensollen (commented):
Thanks @14thibea !
Looks good, I just have a couple suggestions :

clinica/utils/inputs.py
14thibea: It's just that it is only used in this function to parallelize the for loop and also this is the way we did it in ClinicaDL (so I just copied the pattern).
Then there is no good reason actually ^^
We can put it outside if you prefer!

clinica/utils/inputs.py
```diff
@@ -251,14 +260,40 @@ def clinica_file_reader(
...
+        if is_bids:
+            origin_pattern = join(input_directory, sub, ses)
+        else:
+            origin_pattern = join(input_directory, "subjects", sub, ses)
+
+        current_pattern = join(origin_pattern, "**/", pattern)
```
omar-rifai, line 282: It looks like `pattern` is not defined until afterwards, or am I missing something?

omar-rifai (changes requested):
Thanks, looks good ! I just have a small doubt on the pattern variable.

clinica/utils/inputs.py
NicolasGensollen: Ok, thank you for the explanations. I'm fine with both solutions but a little bit more in favor of extracting it into a private helper function.
Extracting inner functions into private functions is usually good practice (if it doesn't use variables from the outer function's scope) as it yields a simpler code (code is less nested, functions are shorter...) and ease code reusability.

clinica/utils/inputs.py
omar-rifai: No sorry my bad, the `pattern` variable seems defined in the outer function before the function call. That being said, it might be clearer if `pattern` was an explicit argument to the inner function (which sort of echoes @NicolasGensollen's comment).

NicolasGensollen: This will work but it would be cleaner to pass `pattern` as a proper arg (also it won't work if we decide to extract it out of the outer function as suggested in my previous comment).

ghisvail: If you keep it inside, `find_sub_ses_pattern` becomes a closure which *inherits* the context of the caller (including `pattern`). If you move it out, it becomes a standalone function and needs a proper parameter for passing the value.

It's just to say @14thibea that you did nothing wrong, both usages work. For testing purposes however, I second @NicolasGensollen and @omar-rifai recommendations to extract the code to a separate function instead of using a closure.

14thibea:
I answered to your comments in the last commit :)

ghisvail:
> [...] and on a large data set this takes forever.

Just out of curiosity, what's the size of the dataset?

How much of a gain did you manage to get with this?

omar-rifai (approved):

PR #664, opened 2022-05-31
Fix warnings in dwi utility module

ghisvail:


ghisvail:
>  Just wondering, why didn't you add type annotations for the other functions of this file?

For now, I am mostly focusing on warnings (soft and hard), and the absence of function annots is not considered one. I add them if a warning concerns a function signature, like missing docstrings for a parameter or an ambiguous type signature.

PR #665, opened 2022-05-31
Call read_csv from the top-level pandas module

ghisvail:
Using `pd.io.parsers` triggers a `Cannot find reference` warning from code introspection in PyCharm. Anyway, the officially documented import for `read_csv` is the top-level `pandas` module, so we should normalize our code base to it.


NicolasGensollen (approved):
LGTM, thanks!

PR #667, opened 2022-06-01
Fix usage of attrs module

ghisvail:
`attr` is the old package name, we should be using `attrs` instead now


NicolasGensollen (approved):
LGTM, thanks!

PR #668, opened 2022-06-01
Fix default value in MissingModsTracker constructor

ghisvail:
`mod_list` can be `None` or `[]`, using `False` leads to a type mismatch warning.


NicolasGensollen (approved):
LGTM, thanks!

PR #669, opened 2022-06-02
Python implementation for StatisticsSurface

### LINKED ISSUE #643: Find an alternative to ClinicaSurfStat
Clinica implements its own wrapper around [SurfStat](https://math.mcgill.ca/keith/surfstat/), a MATLAB toolbox for statistical analysis of surface and volumetric data. MATLAB runtime management is a pain and `nipype` has been of great help to offset that pain away. That being said, the least amount of custom MATLAB code we maintain within Clinica the better.

There has been a new toolbox in the block called [BrainStat](https://www.biorxiv.org/content/10.1101/2022.01.18.476795v1.full.pdf), inspired by SurfStat, but also implemented in Python and available under a permissive license. This may be a good candidate for replacing `ClinicaSurfStat` in the long run.

NicolasGensollen:
Fixes #643 

## Description

This PR proposes to replace the current implementation of the `StatisticsSurface` pipeline, which relies on the MATLAB [SurfStats](https://www.math.mcgill.ca/keith/surfstat/) toolbox, by a pure Python implementation relying on [BrainStat](https://brainstat.readthedocs.io/en/master/).

There are several reasons motivating this migration, the main ones being that `SurfStats` isn't maintained anymore and that it is written in MATLAB. The current solution implemented in Clinica is to vendor the MATLAB toolbox with a custom wrapper to enable calling it from Python, which is obviously far from ideal...

### POC repo

First of all, here is the repo with the proof-of-concept I made before opening this PR: https://github.com/NicolasGensollen/POC_Stat_Pipeline

It is possible to experiment with the code more easily than through Clinica's pipelining architecture.

### Run the StatisticsSurface pipeline in pure Python

Obviously, this requires to have Brainstat installed (since it is not a dependency of Clinica yet). This can be done easily with pip:

```bash
$ pip install brainstat
```

For now, I made the minimum amount of work to integrate it into Clinica. 

So there is still a decent amount of work to do in order to have a clean integration into Clinica.

Nonetheless, it should be possible to run the pipeline (without the plots which are crashing atm for some reason...):

```bash
$ clinica run statistics-surface ./GitRepos/clinica_data_ci/data_ci/StatisticsSurface/in/caps/ UnitTest t1-freesurfer group_comparison ./GitRepos/clinica_data_ci/data_ci/StatisticsSurface/in/subjects.tsv group --covariates age --covariates sex -np 1 -wd $HOME/WD
```

Feel free to try it and take a look at the code.

Feedbacks are welcome as always! 😃


ghisvail:
I have added `brainstat` to the project and refreshed the lock file. It brings quite a bit of transitive dependencies as a result. That's something to keep in mind.

It could be an argument for investigating an alternative with just `nilear` (assuming this is possible), since Clinica already depends on it.

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,157 @@
...
+        .. warning::
+            The contrast needs to be in the design matrix.
+
+    glm_type : {"group_comparison", "correlation"}
+        Type of GLM to run:
+            - "group_comparison": Performs group comparison.
+              For example "AD - ND".
+            - "correlation": Performs correlation analysis.
+
+    group_label : Label for thr group.
```
omar-rifai, line 101:
```suggestion
    group_label : Label for the group.
```

clinica/pipelines/statistics_surface/_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+import pandas as pd
+
+MISSING_TERM_ERROR_MSG = Template(
+    "Term ${term} from the design matrix is not in the columns of the "
+    "provided TSV file. Please make sure that there is no typo."
+)
+
+
+def _is_categorical(df: pd.DataFrame, column: str) -> bool:
```
omar-rifai, line 11: I wonder if this module is not a bit overkill for the task it contains. Maybe an inline code like that does the below would suffice ? Or perhaps moving `_is_categorical` to a utils module across pipelines.

```
def is_categorical(array_like):
    return array_like.dtype.name == 'category'
```

clinica/pipelines/statistics_surface/_outputs.py
```diff
@@ -0,0 +1,226 @@
...
+    ----------
+    results : Results to write to disk.
+    filename_root : Filename root common to all output files.
+    verbose : Verbose mode.
+    """
+    import json
+
+    out_json_file = str(filename_root) + "_results.json"
+    if verbose:
+        print(f"--> Writing results to JSON in {out_json_file}...")
```
omar-rifai, line 25: Let's have all the print messages use `cprint` to be able to channel through the logger. That would automatically handle the verbose flag.

clinica/pipelines/statistics_surface/_outputs.py
```diff
@@ -0,0 +1,226 @@
...
+    """Write provided results to BIDS format.
+
+    .. warning::
+        This is not implemented yet.
+
+    """
+    warnings.warn("Writing results to BIDS is not implemented yet.")
+
+
+WRITERS = {
```
omar-rifai, line 95: Move up for better visibility.

clinica/pipelines/statistics_surface/_outputs.py
```diff
@@ -0,0 +1,226 @@
...
+    Parameters
+    ----------
+    mesh : Mesh of the surface to plot.
+    texture : Texture of the surface to plot.
+    threshold : Threshold to be used for plotting.
+    title : Title to display on the plot.
+    verbose : Verbose mode.
+    """
+    try:
+        from nilearn.plotting import plot_surf_stat_map
```
omar-rifai, line 164: I don't really understand the reason for encapsulating this import in a try catch statement as it is in our dependencies

clinica/pipelines/statistics_surface/_outputs.py
```diff
@@ -0,0 +1,226 @@
...
+    results : Dictionary of results.
+    filename_root : The common root for the output filenames.
+    mesh : The mesh to be used for plotting.
+    verbose : Verbose mode.
+    """
+    RESULTS_NO_PLOT = set(["coefficients"])
+    for name, result in results.items():
+        if name not in RESULTS_NO_PLOT:
+            if isinstance(result, dict):
+                texture = result["P"]
```
omar-rifai, line 198: What is "P" here? Maybe add an info in the docstring.

clinica/pipelines/statistics_surface/_outputs.py
```diff
@@ -0,0 +1,226 @@
...
+    provided results dictionary.
+
+    Parameters
+    ----------
+    results : Dictionary of results.
+    filename_root : The common root for the output filenames.
+    mesh : The mesh to be used for plotting.
+    verbose : Verbose mode.
+    """
+    RESULTS_NO_PLOT = set(["coefficients"])
```
omar-rifai, line 194: Is there a reason you're not building a set directly?

omar-rifai (changes requested):
Hi @NicolasGensollen, Thanks for this work ! LGTM. There are a few suggestions in comments.

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,157 @@
...
+    from scipy.stats import t
+
+    results = dict()
+    results["coefficients"] = np.nan_to_num(model.coef)
+    results["TStatistics"] = np.nan_to_num(model.t)
+    results["uncorrectedPValue"] = dict()
+    results["uncorrectedPValue"]["P"] = 1 - t.cdf(results["TStatistics"], model.df)
+    results["uncorrectedPValue"]["mask"] = mask
+    results["uncorrectedPValue"]["thresh"] = threshold_uncorrected_pvalue
+    results["FDR"] = model._fdr()
```
ghisvail, line 48: PyCharm complains of an access to a protected member here. We are likely supposed to access this information through other means.

I glanced quickly at the implementation of the SLM class, and it looks like this information could be stored in the `model.Q` attribute computed in `model.multiple_comparison_corrections`.

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,157 @@
...
+    tsv_file: PathLike,
+    design_matrix: str,
+    contrast: str,
+    glm_type: str,
+    group_label: str,
+    freesurfer_home: PathLike,
+    surface_file: PathLike,
+    feature_label: str,
+    parameters: Dict,
+    verbose: bool = True,
```
ghisvail, line 69: Those parameters are either unused or undocumented.

clinica/pipelines/statistics_surface/_outputs.py
ghisvail: I fixed it when sweeping through the warnings reported by the code checker.

clinica/pipelines/statistics_surface/_outputs.py
NicolasGensollen: Thanks for the suggestion @omar-rifai ! 👍 
Implemented in 10b4c9c

clinica/pipelines/statistics_surface/clinica_surfstat.py
NicolasGensollen: Indeed, the docstring was incomplete. Thanks for noticing!
I improved it a bit in d1fdf48

omar-rifai:
@NicolasGensollen, I'd suggest to drop the vendored surfstat in a seperate PR. It will make the review process easier. Also it'll isolate the creation from the deletion in case we need to fall back to the previous version for whichever reason.

NicolasGensollen:
@omar-rifai sounds like a good plan. 
I rebased and dropped the commit deleting `SurfStat`.

ghisvail:
We won't get tests for Python 3.10 because of VTK. There is a prerelease available, which we could use in case the next release has not happened on time.

ghisvail:
Support for Python 3.10 needs wheels from the prerelease version of VTK 9.2 and this [fix](https://github.com/MICA-MNI/BrainSpace/pull/75) I submitted upstream.

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+
+    def plot(self, result: StatisticsResults, method: str):
+        plotter = self._get_plotter(method)
+        return plotter(result)
+
+    def _get_plotter(self, method: str):
+        if method == "nilearn_plot_surf_stat_map":
+            return self._plot_stat_maps
+        else:
+            raise NotImplementedError(f"Plotting method {method} is not implemented.")
```
ghisvail: Do we really need dynamic dispatch? It messes with static code analysis of the `plot` message and it raises the cognitive overhead for the code.

ghisvail: I solved the static analysis part. Still not sure what dynamic dispatch provides here.

ghisvail (changes requested):
I have done a first pass just from a static analysis standpoint. I fixed what was straightforward, and left some comments on stuff that may need further info.

test/unittests/pipelines/statistics_surface/test_model.py
```diff
@@ -0,0 +1,323 @@
...
+    assert serializer.mat_extension == ".mat"
+    with pytest.raises(
+        NotImplementedError, match="Serializing method foo is not implemented."
+    ):
+        serializer.save(results, "foo")
+    serializer.save(results, "json")
+    assert os.path.exists(tmp_path / Path("out/dummy_results.json"))
+    with open(tmp_path / Path("out/dummy_results.json"), "r") as fp:
+        serialized = json.load(fp)
+    # assert results.to_dict(jsonable=True) == serialized
```
ghisvail, line 300: To be removed?

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+                    "please carefully check the format of your tsv files."
+                )
+                return GroupGLMWithInteraction(
+                    design, df, self.feature_label, contrast, **kwargs
+                )
+            return GroupGLM(design, df, self.feature_label, contrast, **kwargs)
+        raise ValueError(
+            f"GLM factory received an unknown GLM type: {glm_type}."
+            f"Only 'correlation' and 'group_comparison' are supported."
+        )
```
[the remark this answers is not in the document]
ghisvail: Imo, it should be the concrete class responsibility, not the factory method, to handle consistency logic such as:

```python
        elif glm_type == "group_comparison":
            if "*" in design:
                # Raise warning or error.
```

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+
+
+class PValueResults:
```
ghisvail, line 437: Good idea to implement our own record type for the GLM results. I'd suggest to turn them into a [dataclass](https://docs.python.org/3/library/dataclasses.html) as a result.

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+        tstats: Array of T-statistics.
+        df: Subjects DataFrame.
+        mask: Binary mask.
+        threshold: Threshold.
+        """
+        from scipy.stats import t
+
+        return cls(1 - t.cdf(tstats, df), mask, threshold)
+
+    def to_dict(self, jsonable: bool = True):
```
ghisvail, line 479: I am not sure what `jsonsable` provides. Maybe you want both a `to_dict` and a `to_json` method?

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+
+    def __init__(
+        self,
+        pvalues: np.ndarray,
+        cluster_pvalues: np.ndarray,
+        mask: np.ndarray,
+        threshold: float,
+    ) -> None:
+        super().__init__(pvalues, mask, threshold)
+        self.C = cluster_pvalues
```
ghisvail, line 518: `self.C`breaks standard Python naming convention. Why not keeping `cluster_pvalues`?

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+    def save(self, result: StatisticsResults, method: str) -> None:
+        """Save provided `StatisticsResults` to disk with provided method.
+
+        Parameters
+        ----------
+        result: StatisticsResults to be saved.
+        method: Saving method to use.
+        """
+        writer = self._get_writer(method)
+        return writer(result)
```
ghisvail, line 671: `def save(...) -> None`

```suggestion
        writer(result)
```

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,151 @@
...
+        - "clusterthreshold": Threshold to be used to declare clusters as
+          significant. Default=0.001.
+    """
+    # Load subjects data
+    df_subjects = _read_and_check_tsv_file(tsv_file)
+    if "sizeoffwhm" in parameters:
+        fwhm = parameters["sizeoffwhm"]
+    else:
+        fwhm = DEFAULT_FWHM
+        parameters["sizeoffwhm"] = fwhm
```
ghisvail, line 131: Can be simplified to:

```suggestion
    fwhm = parameters.setdefault("sizeoffwhm", DEFAULT_FWHM)
```

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+    cprint("After correction (Cluster-wise Correction for Multiple Comparisons): ")
+    df = model.P["clus"][0]
+    cprint(df)
+    cprint(f"Clusters found: {len(df)}")
+    cprint(
+        f"Significative clusters (after correction): {len(df[df['P'] <= threshold])}"
+    )
+
+
+def _is_categorical(df: pd.DataFrame, column: str) -> bool:
```
ghisvail, line 45: This function deals with two responsibilities, figure whether a column is categorical and whether a column exist in the input DataFrame which raises an exception. The latter use case cannot be deduce from the function name nor the docstring.

Instead maybe changing the logic to enumerate categorical columns from an input DataFrame?

To test whether a column [is categorical](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html#equality-semantics):

```python
column_dtype = dataframe.dtypes[column_name]
is_categorical = column_type == "category"
```

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+            sub_terms = [_.strip() for _ in term.split("*")]
+            model_term = reduce(
+                lambda x, y: x * y, [_build_model_term(_, df) for _ in sub_terms]
+            )
+        else:
+            model_term = _build_model_term(term, df)
+        model.append(model_term)
+    if len(model) == 1:
+        return model[0]
+    return reduce(lambda x, y: x + y, model)
```
ghisvail, line 112: Since `FixedEffect` implements `__add__`, we can fold with `sum`:

```suggestion
    return sum(model)
```

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+    for term in terms:
+        # Intercept is automatically included in brainstat
+        if term == "1":
+            continue
+        # Handles the interaction effects
+        if "*" in term:
+            sub_terms = [_.strip() for _ in term.split("*")]
+            model_term = reduce(
+                lambda x, y: x * y, [_build_model_term(_, df) for _ in sub_terms]
+            )
```
ghisvail, line 106: Can be folded with `sum`:

```suggestion
            model_term = sum([_build_model_term(_, df) for _ in sub_terms])
```

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+        else:
+            model_term = _build_model_term(term, df)
+        model.append(model_term)
+    if len(model) == 1:
+        return model[0]
+    return reduce(lambda x, y: x + y, model)
+
+
+def _build_model_term(term: str, df: pd.DataFrame) -> FixedEffect:
```
ghisvail, line 115: Same comment as the `_is_categorical` helper function. We are again checking that a specific column in the DataFrame, and I don't think it is the responsibility of this utility function to do it either.

We are probably missing some sort of sanitization step upstream, which given an input DataFrame and a list of terms, checks they are compatible before preparing the model.

This way

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+    used in the Clinica SurfaceStatistics pipeline.
+    """
+
+    def __init__(
+        self,
+        design: str,
+        df: pd.DataFrame,
+        feature_label: str,
+        contrast: str,
+        **kwargs,
```
ghisvail, line 144: We should probably get rid of the `kwargs`. All the arguments with default values are known a priori, so we would be better off listing them explicitly rather than using `kwargs.pop`.

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,151 @@
...
+    surface_file = _get_t1_freesurfer_custom_file_template(input_dir)
+    thickness = _build_thickness_array(input_dir, surface_file, df_subjects, fwhm)
+
+    # Load average surface template
+    fsaverage_path = freesurfer_home / Path("subjects/fsaverage/surf")
+    average_surface, average_mesh = _get_average_surface(fsaverage_path)
```
ghisvail, line 137: Apparently, BrainStat provides its own [fetchers](https://github.com/MICA-MNI/BrainStat/blob/6399b74e0076f648cc1e64fda0cd22740a5dbb78/brainstat/datasets/base.py#L110). Perhaps they could be useful?

clinica/pipelines/statistics_surface/clinica_surfstat.py
```diff
@@ -0,0 +1,153 @@
...
+def clinica_surfstat(
+    input_dir: PathLike,
+    output_dir: PathLike,
+    tsv_file: PathLike,
+    design_matrix: str,
+    contrast: str,
+    glm_type: str,
+    group_label: str,
+    freesurfer_home: PathLike,
+    surface_file: PathLike,
```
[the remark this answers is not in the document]
NicolasGensollen: Indeed. I think the intend was to enable users to provide the surface file themselves, and only rely on the t1 FreeSurfer template when nothing was provided. I added this in [8fe4f4a](https://github.com/aramis-lab/clinica/pull/669/commits/8fe4f4abf1e29ff7ca438e9d220db407c97f510c)

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+    run with one specific contrast.
+
+    Attributes
+    ----------
+    coefficients: The beta coefficients of the fitted GLM model.
+    tstats: The corresponding T-statistics.
+    uncorrected_p_value: The corresponding uncorrected p values,
+        stored in a `PValueResults` instance.
+    FDR: The corresponding False Discovery Rate.
+    corrected_p_value: The corresponding corrected p values,
```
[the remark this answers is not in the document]
NicolasGensollen: Yes, I agree. I kept those weird attribute names to match the output results of the previous Matlab-based pipeline (they are written as labels in output Matlab files). In [ea38312](https://github.com/aramis-lab/clinica/pull/669/commits/ea38312dc350a13e5f3ddadb5befdeb00a98d755) I propose another solution which replaces the attribute names with proper Pythonic snake case names, but adds camel case properties to access them. Let me know if you think this is a reasonable solution.

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: I implemented it this way as I thought we could add other plotting methods with little cost. I was thinking at other, more advanced, surface plotting functions provided by Nilearn, like [view_img_on_surf](https://nilearn.github.io/dev/modules/generated/nilearn.plotting.view_img_on_surf.html#nilearn.plotting.view_img_on_surf) for instance. We could also use plotting capabilities of Brainstat although I'm not familiar with them.
Looking at it now, I agree this might not be the best implementation. So, if you think we won't need additional plotting capabilities and this solution is too complex, I can simplify.

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: Good suggestion. Implemented in [b558a64](https://github.com/aramis-lab/clinica/pull/669/commits/b558a6430ecebd33d5346b1a84d6e3d94abae4a7)

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: I agree with this comment and the one below about duplicated `to_dict` method.
I made some refactoring in [3931d6e](https://github.com/aramis-lab/clinica/pull/669/commits/3931d6e5684151e63d8dffc396da84f1e71495b6). Let me know what you think.

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: See my previous comment.

clinica/pipelines/statistics_surface/_model.py
```diff
@@ -0,0 +1,728 @@
...
+        )
+        return cls(
+            np.nan_to_num(model.coef),
+            tstats,
+            uncorrected_p_values,
+            model.Q,
+            corrected_p_values,
+        )
+
+    def to_dict(self, jsonable: bool = True):
```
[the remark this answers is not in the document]
NicolasGensollen: Refactoring done in [3931d6e](https://github.com/aramis-lab/clinica/pull/669/commits/3931d6e5684151e63d8dffc396da84f1e71495b6)

clinica/pipelines/statistics_surface/clinica_surfstat.py
NicolasGensollen: Great suggestion ! 👍

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: I agree for the responsibility issue. I made some changes in [1a1e9bf](https://github.com/aramis-lab/clinica/pull/669/commits/1a1e9bf7d22797d5ceb7dbbd8445cecd85af3065) to improve this.

I might be missing something however for the way you suggest we test for categorical columns. I agree this would be the way to test it given a properly defined (meaning with dtypes set) DataFrame, but the DataFrame we are dealing with here comes from a brutal tsv read with no a priori information, typically on the types of the data we are reading. So the columns with categorical data will just have the `object` dtype. 
Does this make sense?

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: The lambda function is the product function, not the sum.

ghisvail:
My bad, should be [math.prod](https://docs.python.org/3.8/library/math.html#math.prod) then.

________________________________

clinica/pipelines/statistics_surface/_model.py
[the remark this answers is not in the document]
NicolasGensollen: Yes, the `GLMFactory` class implementation might indeed be a bit too complex with no real benefit compared to a simple factory function.
I implemented your suggestions (both the change to a factory method and displacement of the consistency logic within the concrete class) in [3bb21f7](https://github.com/aramis-lab/clinica/pull/669/commits/3bb21f7f15dd68a26ddc68af72fa4ff72496270e)

clinica/pipelines/statistics_surface/clinica_surfstat.py
NicolasGensollen: We can discuss this next week, I'm not really convinced that using their fetchers would be a better option compared to the solution implemented here.

ghisvail:
I fixed the merge conflict and updated the `brainspace` dependency to use the new published version. It looks like there are still issues with unit tests though.

NicolasGensollen:
Thanks for fixing the conflict @ghisvail 
I hadn't updated the tests (now done in [3cd895f](https://github.com/aramis-lab/clinica/pull/669/commits/3cd895fdff07813fe9dc9e1560fd841043d0b192))
I should have addressed all your comments and suggestions except for the `sum()` and `prod()` aggregations which are giving unexpected results. I opened https://github.com/MICA-MNI/BrainStat/issues/290 a few days ago and reverted back to lambda aggregation functions in the mean time.

ghisvail:
> Thanks for fixing the conflict @ghisvail I hadn't updated the tests (now done in [3cd895f](https://github.com/aramis-lab/clinica/pull/669/commits/3cd895fdff07813fe9dc9e1560fd841043d0b192)) I should have addressed all your comments and suggestions except for the `sum()` and `prod()` aggregations which are giving unexpected results. I opened [MICA-MNI/BrainStat#290](https://github.com/MICA-MNI/BrainStat/issues/290) a few days ago and reverted back to lambda aggregation functions in the mean time.

You did well. It's a weird bug indeed.

clinica/pipelines/statistics_surface/_model.py
ghisvail: > I agree for the responsibility issue. I made some changes in [1a1e9bf](https://github.com/aramis-lab/clinica/pull/669/commits/1a1e9bf7d22797d5ceb7dbbd8445cecd85af3065) to improve this.

That's better indeed 👍 

> DataFrame we are dealing with here comes from a brutal tsv read with no a priori information

So how do you figure whether a column is categorical or not from a raw string format without additional metadata, then? This is a very good question.

The implementation says anything whose dtype name does not start with "float". Which is a proxy for testing whether a column contains values which are non-float. So any column which dtype is either strings, objects or integers would be considered as categorical. Is it really the right logic, though?

Let's take the following example: if I have an age column successfully parsed as integers, then it's categorical. If it's parsed as floats, then it's not. Date-like columns would be considered categorical too. Also, `object` dtype is attributed when the `pandas` sniffers fail to detect a precise dtype from a column. Those columns may be categorical or malformed.

Should we decide to keep this logic, then it would be better to rename this filter to `is_nonfloat` rather than keep `is_categorical` in my opinion. Another option would be to enhance the reader to analyze columns which have been successfully converted as strings after `convert_dtype`, compute their histograms, apply a conservative heuristic between the range of different values over the total of lines and automatically convert them to a proper categorical dtype. It would be significantly more work though.

clinica/pipelines/statistics_surface/_model.py
ghisvail: Yeah, I meant `math.prod`. Sorry for the hasty copy pasting from the previous comment.

ghisvail (changes requested):
Another batch of comments and suggestions. I stopped the numpydoc suggestions after I realized most of them are missing typing informations for parameters and returned values.

Please consider making another pass at them with the [style guide](https://numpydoc.readthedocs.io/en/latest/format.html) open for a refresher.

NicolasGensollen:
For some reason, I thought Numpydoc had converged on the automatic integration of type hints like [Napoleon already does](https://github.com/sphinx-doc/sphinx/blob/master/doc/usage/extensions/napoleon.rst#type-annotations)...
But you're right, this is still an open issue: https://github.com/numpy/numpydoc/issues/196 so these docstrings aren't compliant. I'll change them.

clinica/pipelines/statistics_surface/_inputs.py
```diff
@@ -0,0 +1,150 @@
...
+        raise ValueError(f"The TSV data in {tsv_file} should have at least 2 columns.")
+    if tsv_data.columns[0] != TSV_FIRST_COLUMN:
+        raise ValueError(
+            f"The first column in {tsv_file} should always be {TSV_FIRST_COLUMN}."
+        )
+    if tsv_data.columns[1] != TSV_SECOND_COLUMN:
+        raise ValueError(
+            f"The second column in {tsv_file} should always be {TSV_SECOND_COLUMN}."
+        )
+    return tsv_data
```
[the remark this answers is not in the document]
NicolasGensollen: Good suggestion, implemented in [b225b69](https://github.com/aramis-lab/clinica/pull/669/commits/b225b69c7f8cadf4b64ddc605a598559ae8109b2)

clinica/pipelines/statistics_surface/_model.py
NicolasGensollen: The current implementation is legacy code I left as it was. I do agree it's far from perfect and might lead to spurious failures down the road as you explained above.

I liked the heuristic solution you proposed and went for an implementation in [66bc1c9](https://github.com/aramis-lab/clinica/pull/669/commits/66bc1c91b623a27601cefa542789842e98f185eb).
Let me know if this is what you had in mind !

One thing though is that using `convert_dtype` is currently breaking things in `Brainstat`.
I think this is a bug on their side, so I filled a bug report (https://github.com/MICA-MNI/BrainStat/issues/292), and opened a fix (https://github.com/MICA-MNI/BrainStat/pull/293).
We'll see how it goes... In the mean time, tests are broken, but pass with the proposed upstream fix.

PR #670, opened 2022-06-03
Fix warnings in ADNI converter

ghisvail:


clinica/iotools/converters/adni_to_bids/adni_modalities/adni_t1.py
```diff
@@ -455,17 +454,18 @@ def select_scan_from_qc(scans_meta, mayo_mri_qc_subj, preferred_field_strength):
...
-        multiple_mag_strength = True
+    if multiple_mag_strength:
         # Save for later the scans with the non preferred magnetic field strength
         not_preferred_scan = scans_meta[
             scans_meta.MagStrength != preferred_field_strength
         ]
         # Filtering to keep only the scans with the preferred magnetic field strength
         scans_meta = scans_meta[scans_meta.MagStrength == preferred_field_strength]
+    else:
+        not_preferred_scan = None
```
NicolasGensollen, line 468: I'm wondering whether we should add a check either before calling `select_scan_no_qc(scans_meta)`, or within `select_scan_no_qc` to raise a meaningful error if `scans_meta` is `None`. 
Atm, `scans_meta` cannot be `None` when reaching the return statement, but it's not super obvious and could be a source of error when trying to refactor this function later. 
I don't have a strong opinion, we can leave it like this if you think this would be superfluous.

clinica/iotools/converters/adni_to_bids/adni_to_bids.py
```diff
@@ -38,15 +38,17 @@ def get_bids_subjs_info(
 
 
 class AdniToBids(Converter):
-    def get_modalities_supported(self):
+    @classmethod
```
NicolasGensollen, line 41: Is there a reason you decided to use a class method instead of a static method (or a `supported_modalities` property) here?

NicolasGensollen (approved):
Thanks @ghisvail !
LGTM!

clinica/iotools/converters/adni_to_bids/adni_to_bids.py
ghisvail: This code is not specific to any class instance (`self` is unused) so it makes sense to turn into a class method.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_t1.py
ghisvail: > [...] but it's not super obvious and could be a source of error when trying to refactor this function later. I don't have a strong opinion, we can leave it like this if you think this would be superfluous.

I believe the overall data flow and error detection need to be redesigned for both ADNI and AIBL, beyond just this function. It should be revisited in a follow-up refactoring PR in my opinion.

PR #671, opened 2022-06-03
Fix accidental bypass of sorting the conversion versions

ghisvail:
Found that one whilst dealing with #670.

It still worked because the preceding `os.listdir` operation seem to return a sorted iterable. This PR ensures sorting does happen no matter what `os.listdir` returns.


omar-rifai:
Thanks !

PR #676, opened 2022-06-07
Fix warnings in NIFD converter

ghisvail:


clinica/iotools/converters/nifd_to_bids/nifd_to_bids_cli.py
```diff
@@ -8,13 +10,13 @@
 @cli_param.clinical_data_directory
 @cli_param.bids_directory
 def cli(
-    dataset_directory: str,
-    clinical_data_directory: str,
-    bids_directory: str,
+    dataset_directory: PurePath,
```
NicolasGensollen, line 13: It'd be great to have consistency across converters. At the moment, some converters (like `ADNI_to_bids`) have string types for paths, others (like `ukb_to_bids`) have `PathLike`, and this one will have `PurePath`. WDYT?

clinica/iotools/converters/nifd_to_bids/nifd_utils.py
```diff
@@ -310,8 +309,8 @@ def convert_dicom(sourcedata_dir: PathLike, bids_filename: PathLike) -> None:
 
     # Run conversion with dcm2niix with anonymization and maximum compression.
     run_dcm2niix(
-        input_dir=sourcedata_dir,
-        output_dir=output_dir,
+        input_dir=str(sourcedata_dir),
+        output_dir=str(output_dir),
```
NicolasGensollen, line 313: Why do you need to cast to string here? `output_dir` should already be a string after casting at line 302, right?

NicolasGensollen (approved):
LGTM, thanks @ghisvail !

clinica/iotools/converters/nifd_to_bids/nifd_to_bids_cli.py
ghisvail: Agreed. We are in that weird sport where the historic path manipulation library (`os.path`) is focused on strings, whilst the modern `pathlib` alternative uses `PurePath`.

`os.PathLike` provides a protocol to bridge between both at the expense of internal casting (to `str` or `Path` / `PurePath`) to please static type checking.

So far, this PR focused on fixing warnings such as type alignment issues. Whether to standardize to `PurePath` or `PathLike` would require a subsequent PR touching all converters.

PR #677, opened 2022-06-08
Enable install on Python 3.10

ghisvail:
I could not install Clinica on Fedora 36, which ships with Python 3.10 as default. The SciPy version in the lock file is too ancient to pull the appropriate wheel package. Since we have placed a lower bound to version 3.8, we could try raising the minimum SciPy version matching this constraint.

Another surprise is that SciPy now adds an upper bound to the supported Python 3 versions. Since we are explicitly depending on it, this upper bound needs to be carried along with us to keep Poetry's dependency resolver happy.


NicolasGensollen (approved):
Thanks @ghisvail !
I tried it locally and confirm it works for Python 3.8 to 3.10.

PR #679, opened 2022-06-10
Fix warnings on UKB converter

ghisvail:


clinica/iotools/converters/ukb_to_bids/ukb_to_bids_cli.py
```diff
@@ -27,7 +27,7 @@ def cli(
     check_dcm2niix()
 
     convert_images(dataset_directory, bids_directory, clinical_data_directory)
-    _write_bidsignore(bids_directory)
+    _write_bidsignore(str(bids_directory))
```
NicolasGensollen, line 30: I'm not sure to understand why you need this casting. `_write_bidsignore` expects either a string or a Path, and will cast it's input to a Path anyway.

NicolasGensollen (approved):
LGTM, thanks @ghisvail !

clinica/iotools/converters/ukb_to_bids/ukb_to_bids_cli.py
ghisvail: `bids_directory` is declared `PathLike`, so neither a string nor a path. Since `_write_bidsignore` wants a concrete type (`str | Path`), we need to cast explicitly to satsify static type checking.

PR #682, opened 2022-06-11
Compress and decompress with new Gzip interface

ghisvail:
I recently contributed a [new interface](https://github.com/nipy/nipype/pull/3472) enabling both compression and decompression of a file. This PR uses this interface instead of our homegrown function similar to #650.


ghisvail:
Closing in favor of a rewrite to Pydra.

Compression at write should be handled by a more featureful writer implementation (such as what `niworkflows` uses) instead of hacking our way with a combination of a generic `DataSink` composed with an arbitrary callable.

PR #685, opened 2022-06-14
Updates for release

omar-rifai:
Changes for release 0.7.1

PR #686, opened 2022-06-14
[DOC] Add change log to documentation

NicolasGensollen:
#685 made me realize that Clinica has a maintained change log, which is great!
But also that it is not currently included in the documentation although this is pretty standard to do so (scikit-learn, MNE, Nilearn, Scipy, Pandas for example all do this).
This PR proposes to add it to the docs.


omar-rifai:
LGTM! thanks.

PR #689, opened 2022-06-28
ci: Add caching support for unit tests

ghisvail:
- Use `snok/install-poetry` GH action to deploy Poetry configured with inplace venv support.
- Cache the venv created by Poetry by product of OS, Python version and hash of the lock file.
- Cache miss happens if the Python version changes or the lock file is updated.


NicolasGensollen (approved):
LGTM! Thanks @ghisvail ! 👍

PR #695, opened 2022-07-11
Deal with new data from ADNI3

14thibea:
Hello!

I recently downloaded new meta-data + T1 images of ADNI that may have never been converted with the converter of Clinica. There were 3 issues:
- one participant had no screening diagnosis --> its value for `diagnosis_sc` is now filled with `"n/a"`
- one session was poorly converted as it led to a `_Eq.nii` file --> it was added to the list of exceptions,
- the amyloid status was not computed for one session --> its value is now `"n/a"` (however, it would seem more relevant to code it as `Au`, but I did not remember exactly what corresponded to this label).
- the `viscode_to_session` function was modified to take into account the fact that "M00" may be encoded as "m0" and not only "bl" (case of `sub-ADNI003S6014` in `NPIQ.csv` file.

As some images are still missing because the downloading process partly failed, this PR should not be merged yet.


14thibea:
Hi @ghisvail :)

I set up a pre-commit hook for next time!

Actually there is a list of pairs of participant/session for each modality that are poorly converted and thus not converted by the Clinica converter. They can be found in files `iotools/converters/adni_to_bids/adni_modalities/adni_X.py`, in the corresponding method `compute_X_paths`, looking for the lines after the comment `# Exceptions` (where you should replace X by a modality, for example `t1`).
It is also described in the documentation [here](https://aramislab.paris.inria.fr/clinica/docs/public/latest/Converters/ADNI2BIDS/#using-the-converter), but it is a bit hidden as you have to click on "Known errors" to see all the list.
I remembered that at some point @MatthieuJoulot investigated and removed a lot of pairs from these lists by upgrading the versions of our dependencies (after that some images were actually correctly converted).

The amyloid status was described in [(Landau et al., 2013)](https://doi.org/10.2967/jnumed.112.109009), and we used it for example in [this paper](https://www.biorxiv.org/content/early/2018/04/06/274324).
For the tau status it is less clear, it may be not standard and we have to describe every time the threshold we are using... Maybe @nburgos has more information about this? 
And it seems that none of them is actually described in the documentation 🙃

14thibea:
About the p-tau threshold, it seems that other persons use the same one in other studies (for example [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2696350/)) but not sure if there is a reference / why they use this one.
I guess that on this topic, we need help from a clinician who knows what is going on.

nburgos:
Hello @14thibea and @ghisvail,

I'm not sure to understand what you want to know regarding the amyloid and tau thresholds. :/

Several β-amyloid thresholds exist depending on whether the measurement was obtained from a PET image (AV45 and FBB columns in ADNIMERGE) or a CSF sample (ABETA column in ADNIMERGE). Regarding tau, I'm not sure it makes much sense extracting a single value from PET images, which is anyway recent, so CSF is used instead. And usually people like looking at ratios (between total tau, phosphorylated tau and β-amyloid).

14thibea:
I downloaded everything now.
It's ready to be merged.

PR #699, opened 2022-07-13
Build the user documentation in a separate pipeline

mdiazmel:
This PR implements the build and publish of the user documentation as a different CI job. It uses a different Jenkins instance in order to unload the charge of the main instance.

### DIFF
--- a/.jenkins/doc_build
+++ b/.jenkins/doc_build
@@ -17,11 +17,11 @@ pipeline {
       }
       steps {
         sh '''
-                source "${CONDA_HOME}/etc/profile.d/conda.sh"
-                conda create -y -p "${CONDA_ENV}" python=3.9
-                conda activate "${CONDA_ENV}"
-                make doc
-                '''
+          source "${CONDA_HOME}/etc/profile.d/conda.sh"
+          conda create -y -p "${CONDA_ENV}" python=3.9
+          conda activate "${CONDA_ENV}"
+          make doc
+        '''
         stash(name: 'doc_html', includes: 'site/**')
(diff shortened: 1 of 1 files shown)


.jenkins/Jenkinsfile
```diff
@@ -295,31 +295,6 @@ pipeline {
...
-              mv site "${CHANGE_ID:-$BRANCH_NAME}"
-              scp -r "${CHANGE_ID:-$BRANCH_NAME}" aramislab:/srv/local/clinica/docs/public/
-            '''
-          }
-          post {
-            always {
-              cleanWs()
-            }
-          }
-        }
```
NicolasGensollen, line 322: Shouldn't we also remove the deploy stage below ?

ghisvail:
Do we really gain something from having separate `build` and `deploy`stages coupled by stashing and unstashing? Since the same node is capable of building and deploying, it sounds to me like an unnecessary overhead.

.jenkins/Jenkinsfile
NicolasGensollen: Indeed, didn't extend enough to realize that. Sorry for the noise...

ghisvail:
> And that let us be more flexible if we add more agents in the future

I agree that separating those stages allow for more recycling of nodes for other jobs. That being said, the cost of spinning up a new node (you might even have to wait for one to be available), stashing and unstashing is definitely not null, and `sshPublisher` is probably as quick, if not quicker to execute in comparison.

ghisvail (approved):

PR #701, opened 2022-07-19
Added pet linear to check_missing_processing

ravih18:
Added pet_linear to check_missing_processing.


clinica/utils/input_files.py
```diff
@@ -457,20 +457,27 @@ def pet_volume_normalized_suvr_pet(
 # pet-linear
 
 
-def pet_linear_nii(acq_label, suvr_reference_region, uncropped_image):
+def pet_linear_nii(acq_label, suvr_reference_region, uncropped_image, use_uniform):
```
ghisvail, line 460: To avoid breaking compatibility with old code.

```suggestion
def pet_linear_nii(acq_label, suvr_reference_region, uncropped_image, use_uniform: bool = False):
```

clinica/utils/input_files.py
```diff
@@ -457,20 +457,27 @@ def pet_volume_normalized_suvr_pet(
...
+        rec = "_rec-uniform"
+    else:
+        pipeline_description = ""
+        rec = ""
+
     information = {
         "pattern": os.path.join(
             "pet_linear",
-            f"*_trc-{acq_label}_pet_space-MNI152NLin2009cSym{description}_res-1x1x1_suvr-{suvr_reference_region}_pet.nii.gz",
+            f"*_trc-{acq_label}_pet_space-MNI152NLin2009cSym{description}_res-1x1x1_suvr-{suvr_reference_region}{rec}_pet.nii.gz",
```
ghisvail, line 478: Better to keep source entities ordered the same way as the [BIDS entity table](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html). 

```suggestion
            f"*_trc-{acq_label}{rec}_pet_space-MNI152NLin2009cSym{description}_res-1x1x1_suvr-{suvr_reference_region}_pet.nii.gz",
```

ghisvail (changes requested):

omar-rifai:
Thanks @ravih18, as discussed, we can add a test for this. Also ideally, the `pet_linear_nii` function should be separated in a different PR.

omar-rifai (changes requested):
As discussed, we can add a test for this. Also ideally, the `pet_linear_nii` function should be separated in a different PR.

PR #706, opened 2022-07-21
Homogeneize notations

14thibea:
Hi Clinica :)

I am currently launching `t1-volume` and I am always confused by the notations in the doc between `group_label` and `group_id` (where `group_id` = `group-<group_label>` but it is not written anywhere).
Here is a small PR to homogeneize.


ghisvail:
Great idea! Indeed, having `group_id` as `group-<group_label>` would be consistent with the BIDS taxonomy.

ghisvail:
> where group_id = group-<group_label> but it is not written anywhere

We should explicitly state it somewhere too.

omar-rifai:
Great, thanks @14thibea.

PR #711, opened 2022-07-25
Add import pathlib to nonregression tests

omar-rifai:
This PR fixes a regression introduced in #694 by importing `Path` into the non-regression modules.


MatthieuJoulot (approved):
LGTM !

PR #717, opened 2022-07-27
Catch error in non-regression tests to continue on failure

omar-rifai:
This PR allows non-regression tests to continue even if one of the stages fails.


MatthieuJoulot (approved):
LGTM

PR #722, opened 2022-08-08
Add BIDS reading task to Pydra

ghisvail:
Integrated from preliminary work done in [pydra-bids](https://github.com/aramis-lab/pydra-bids).

### DIFF
--- a/clinica/pydra/tasks/bids/utils.py
+++ b/clinica/pydra/tasks/bids/utils.py
@@ -46,12 +46,12 @@ def read_bids(output_query: Optional[dict] = None, **kwargs) -> FunctionTask:
 
         layout = BIDSLayout(ds_dir=base_dir)
 
-        results = {
-            key: layout.get(return_type="files", **query)
+        results = [
+            layout.get(return_type="files", **query)
             for key, query in list(output_query.items())
-        }
+        ]
 
-        return tuple(results.values())
+        return tuple(results) if len(results) > 1 else results[0]
 
     input_spec = SpecInfo(
         name="Input",
--- a/clinica/pydra/tasks/bids/utils.py
+++ b/clinica/pydra/tasks/bids/utils.py
@@ -1,0 +1,79 @@
+from typing import Optional
+
+from pydra.engine.task import FunctionTask
+
+DEFAULT_QUERY = {
+    "bold": {
+        "suffix": "bold",
+        "extension": [".nii", ".nii.gz"],
+    },
+    "T1w": {
+        "suffix": "T1w",
+        "extension": [".nii", ".nii.gz"],
+    },
+}
+
+
+def read_bids(output_query: Optional[dict] = None, **kwargs) -> FunctionTask:
+    """Generate a BIDS reading task.
+
+    Parameters
+    ----------
+    output_query : dict
+        Mapping between output and BIDS query
+    **kwargs
+        Additional arguments passed to the task constructor
+
+    Returns
+    -------
+    pydra.engine.task.FunctionTask
+        BIDS reading task
+
+    Examples
+    --------
+    >>> task = read_bids()
+    >>> task.output_names
+    ['bold', 'T1w']
+    """
+    from os import PathLike
+
+    from pydra.engine.specs import BaseSpec, SpecInfo
+
+    output_query = output_query or DEFAULT_QUERY
+
+    def inner(base_dir):
+        from ancpbids import BIDSLayout
+
+        layout = BIDSLayout(ds_dir=base_dir)
+
+        results = {
+            key: layout.get(return_type="files", **query)
+            for key, query in list(output_query.items())
+        }
+
+        return tuple(results.values())
+
+    input_spec = SpecInfo(
+        name="Input",
+        fields=[
+            (
+                "base_dir",
+                PathLike,
+                {"help_string": "Root directory of the BIDS dataset"},
+            ),
+        ],
+        bases=(BaseSpec,),
+    )
+
+    output_spec = SpecInfo(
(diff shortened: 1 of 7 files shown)


ghisvail:
This implementation passes the tests, but there are issues with type alignments between the nipype tasks and the BIDS reader task. I am investigating.

ghisvail:
Looks like I found the issue. Testing on T1 linear works as expected.

ghisvail:
I wonder whether we can do better than converting the query results to a tuple or single value and listing the keys to the output spec. Best would be to be able to say I am going to return a dictionary (directly resulting from the query) whose keys are listed in the output spec of the FunctionTask.

I have asked on Mattermost, we'll see. So far, this solution is doing the job.

clinica/pydra/tasks/bids/utils.py
```diff
@@ -0,0 +1,79 @@
...
+        from ancpbids import BIDSLayout
+
+        layout = BIDSLayout(ds_dir=base_dir)
+
+        results = [
+            layout.get(return_type="files", **query)
+            for key, query in list(output_query.items())
+        ]
+
+        return tuple(results) if len(results) > 1 else results[0]
```
omar-rifai, line 54: I don't understand why the function is changing the return type here for the two use cases?

Also  It seems that when I try to introduce another entity to be read (I tried with bold) the N4BiasFieldCorrectionInputSpec does not receive the correct type (it receives a list. Granted, this second problem was probably present in the previous implementation as well but it's worthwhile thinking about if we are to change the return type.

omar-rifai (commented):
Thanks for the PR. Beside the inline comment, it'd be worthwhile to check that we can indeed filter by subject session in this implementation. Otherwise looks good.

clinica/pydra/tasks/bids/utils.py
ghisvail: > I don't understand why the function is changing the return type here for the two use cases?

There is a difference between returning a tuple composed of a single list (such as -> Tuple[list[str]]) and a single list (-> list[str]). Pydra seems sensitive to that difference.

> N4BiasFieldCorrectionInputSpec does not receive the correct type (it receives a list.

You were right to test this. Sounds more like an issue with the splitter, not the reader.

ghisvail: > Pydra seems sensitive to that difference.

Probably due to this [section](https://github.com/nipype/pydra/blob/ac06f8cc691c62e3d56e9e94ef015a675c73851f/pydra/engine/task.py#L196) of the code.

omar-rifai (changes requested):
I think it makes sense to add the functionality to filter on subject/session in this PR.

clinica/pydra/tasks/bids/utils.py
ghisvail: > Can you illustrate with one use case where a tuple is needed?

You just tried a query with both `T1w` and `bold`. What do you think should be returned?

`-> Tuple[list(str), list(str)]`, one for the list of T1w files, one for the list of bold files.

ghisvail:
> I think it makes sense to add the functionality to filter on subject/session in this PR.

That's gonna take longer then. Is that ok?

clinica/pydra/tasks/bids/utils.py
omar-rifai: > > I think it makes sense to add the functionality to filter on subject/session in this PR.
> 
> That's gonna take longer then. Is that ok?

I mean it's primary reason brought forward motivating the use of `ancpbids` so it seems  the PR is not complete without this feature.

omar-rifai: > What do you think should be returned?

I'm not sure what should be returned but I'm asking why you chose a tuple over a list (of lists) for instance.

ghisvail: That's because if you want N named outputs (for instance `T1w` and `bold`), you need to return a tuple of N elements. Here is the relevant part of the [code](https://github.com/nipype/pydra/blob/ac06f8cc691c62e3d56e9e94ef015a675c73851f/pydra/engine/task.py#L199) for reference.

Returning a `list` or `list(list)` will yield only a single output, or fail with a `RuntimeError` if the shape of the returned type and the list of `output_names` don't match.

omar-rifai: > That's because if you want N named outputs (for instance `T1w` and `bold`), you need to return a tuple of N elements. Here is the relevant part of the [code](https://github.com/nipype/pydra/blob/ac06f8cc691c62e3d56e9e94ef015a675c73851f/pydra/engine/task.py#L199) for reference.
> 
> Returning a `list` or `list(list)` will yield only a single output, or fail with a `RuntimeError` if the shape of the returned type and the list of `output_names` don't match.

Clear. Thanks for the reference.

omar-rifai:
> > I think it makes sense to add the functionality to filter on subject/session in this PR.
> 
> That's gonna take longer then. Is that ok?

I mean it's primary reason brought forward motivating the use of ancpbids so it seems the PR is not complete without this feature

ghisvail:
I have iterated quite a bit on this since the initial design and eventually reached a design for tasks defining dynamic specifications that I am happy with. It is easily testable and easier to read than the initial design based on factories.

Since [pydra-bids](https://github.com/aramis-lab/pydra-bids) will likely appeal to a wider audience than just Clinica, I'd propose to publish it as its own package and have Clinica depend on it.

ghisvail:
> I'd propose to publish it as its own package and have Clinica depend on it.

https://pypi.org/project/pydra-bids/

PR #725, opened 2022-08-10
Pydra add t1volume tissue segmentation

omar-rifai:
Initial implementation of T1Volume Tissue Segmentation with Pydra
- [x] Write core workflow in Pydra
- [x] Successfully execute with `@clinica_io`
- [x] Integrate in hidden command line 
- [x] Write Unit tests for interfaces
- [x] Update docstrings in legacy `t1_volume_tasks` and `t1_volume_utils` functions

Not in scope (to be added to airtable taks):
- Unit tests for legacy functions
- Parameters passing in command line option (in future work when decorator ready)


clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+
+def initialize_tissues_spm_segment(
+    tissue_probability_map: Union[PathLike, None] = None
+) -> tuple:
+
+    """Prepare data structure for SPM segment interface
+
+    Parameters
+    ----------
+    tissue_probability_map: Union(PathLike, None)
```
ghisvail, line 16: numpydoc

```suggestion
    tissue_probability_map : PathLike, optional
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        Path to the niftii tissue probability map
```
ghisvail, line 17: typo

```suggestion
        Path to the nifti tissue probability map
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    Warns
+    -----
+    Prints out message for image <bids_name>
+    """
+    from clinica.utils.ux import print_end_image
+
+    print_end_image(bids_name)
+
+
+def zip_list_files(class_images: list, zip_files: bool = False) -> list:
```
ghisvail, line 87: Same comment as with the `zip_nii` utility function earlier. Sounds that this should be handled with a splitter on a `GZip` task.

ghisvail (changes requested):
That's a decent first attempt at converting `t1-volume-tissue-segmentation` to Pydra. I have got a feeling this is still too close to the original `nipype` implementation and it could be simplified by being more idiomatic to Pydra (see individual review items where applicable).

There is also quite a few items which could have been caught before review by proper static code analysis. I don't know about VSCode, but PyCharm does it pretty well.

test/unittests/pydra/test_t1volume.py
```diff
@@ -0,0 +1,21 @@
...
+    import clinica.pydra.t1_volume.t1_volume_tasks as t1vol_tasks
+
+    inv = t1vol_tasks.ApplySegmentationDeformation()
+    assert inv._jobname == "defs"
+    assert inv._jobtype == "util"
+    assert inv.input_spec().get_traitsfree() == {
+        "mask": 0,
+        "mfile": True,
+        "use_v8struct": True,
+    }
```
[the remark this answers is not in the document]
omar-rifai: In `ApplySegmentationDeformation` we extend SPM's Base interface through this set of parameters. In other words,  this set of parameters generate the specific interface we are after. The test is here to ensure these base parameters don't change.

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+    tissues = []
+
+    for i in range(1, 7):
+        n_gaussians = 2
+
+        if i == 4 or i == 5:
+            n_gaussians = i - 1
+
+        native_space = False
```
NicolasGensollen:
```suggestion
        native_space = i in tissue_classes
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        dartel_input = False
```
NicolasGensollen:
```suggestion
        dartel_input = i in dartel_tissues
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        warped_unmodulated = False
```
NicolasGensollen:
```suggestion
        warped_unmodulated = (i in tissue_classes) and save_warped_unmodulated 
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        warped_modulated = False
```
NicolasGensollen:
```suggestion
        warped_modulated = (i in tissue_classes) and save_warped_modulated
```

And delete the two if blocks afterwards.

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    produced segmentation is wrong (not the shape of a brain anymore)
+    """
+
+    import numpy as np
+
+    center = get_world_coordinate_of_center(nii_volume)
+
+    distance_from_origin = np.linalg.norm(center, ord=2)
+
+    if distance_from_origin < threshold_l2:
```
NicolasGensollen: Replace with 

```python
return distance_from_origin < threshold_l2
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    from os.path import isfile
+
+    import nibabel as nib
+    import numpy as np
+    from nibabel.filebasedimages import ImageFileError
+
+    # from clinica.utils.stream import cprint
+    # assert isinstance(nii_volume, str), "input argument nii_volume must be a str"
+
+    assert isfile(nii_volume), "input argument must be a path to a file"
```
NicolasGensollen: I would raise a proper error instead of using an assertion here...

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+
+    try:
+        orig_nifti = nib.load(str(nii_volume))
+    except ImageFileError:
+        print(f"File {nii_volume} could not be read by nibabel. Is it a valid NIfTI")
```
NicolasGensollen:
```suggestion
        print(f"File {nii_volume} could not be read by nibabel. Is it a valid NIfTI?")
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    bids_name : str
+        The BIDS name of the t1w image
+
+    Warns
+    -----
+    Prints out message for image <bids_name>
+    """
+    from clinica.utils.ux import print_end_image
+
+    print_end_image(bids_name)
```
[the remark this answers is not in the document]
omar-rifai: A legacy function. Removed

clinica/pydra/t1_volume/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_cli.py
```diff
@@ -0,0 +1,28 @@
+import click
+
+import clinica.pydra.engine_utils as pydra_utils
+import clinica.pydra.t1_volume.t1_volume_tissue_segmentation.t1_volume_tissue_segmentation_pipeline as pydra_t1vol
+from clinica.pipelines import cli_param
+
+pipeline_name = "pydra-t1vol-ts"
```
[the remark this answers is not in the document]
omar-rifai: Indeed, just for testing. We can decide on proper abbreviations in due time.

test/unittests/pydra/test_t1volume.py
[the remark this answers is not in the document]
ghisvail: I see. Now that we have mentioned it, the interface looks suspiciously similar to [this one](https://github.com/nipy/nipype/blob/1dee93965dfb6b0d0e50a6abd7ac89bf13dfbf8c/nipype/interfaces/spm/utils.py#L307), with parts of the input spec stripped out.

Might be worth looking whether swapping them yields the same results.

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    -------
+    ndarray
+        Voxel coordinates of the center of the volume
+    """
+    import numpy as np
+
+    center_x = header["dim"][1] / 2
+    center_y = header["dim"][2] / 2
+    center_z = header["dim"][3] / 2
+    return np.array([center_x, center_y, center_z])
```
NicolasGensollen: Can be simplified in : 

```python
return np.array([_ / 2 for _ in header["dim"][1:4]])
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    is used if sform_code = 0. The world coordinates are determined simply by scaling by the voxel size by their
+    dimension stored in pixdim.
+
+    References
+    ----------
+    https://brainder.org/2012/09/23/the-nifti-file-format/
+    """
+    import numpy as np
+
+    return np.array(coordinates_vol) * np.array(
```
NicolasGensollen: Can be simplified to :

```python
return np.array(coordinate_vol) * np.array([_ for _ in header["pixdim"][1:4]])
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        header["pixdim"][1], header["pixdim"][2], header["pixdim"][3]
+    )
+
+
+def vox_to_world_space_method_2(
+    coordinates_vol: ndarray, header: Nifti1Header
+) -> ndarray:
+    """Convert coordinates to world space (method 2)
+
```
NicolasGensollen: Extra blank line here

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        r[1, 1] = (a**2) + (c**2) - (b**2) - (d**2)
+        r[1, 2] = 2 * ((c * d) - (a * b))
+        r[2, 0] = 2 * ((b * d) - (a * c))
+        r[2, 1] = 2 * ((b * d) - (a * c))
+        r[2, 2] = (a**2) + (d**2) - (b**2) - (c**2)
+        return r
+
+    i = coordinates_vol[0]
+    j = coordinates_vol[1]
+    k = coordinates_vol[2]
```
NicolasGensollen: Maybe something like this ?

```python
i, j, k = coordinates_vol
```

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        print("q was " + str(q), ", now is 1")
+        q = 1
+    return np.dot(r_mat, np.array([i, j, q * k])) * np.array(
+        header["pixdim"][1:4]
+    ) + np.array([header["qoffset_x"], header["qoffset_y"], header["qoffset_z"]])
+
+
+def vox_to_world_space_method_3(coordinates_vol: ndarray, header: Nifti1Header):
+    """Convert coordinates to world space (method 3)
+
```
NicolasGensollen: Extra line

clinica/pydra/t1_volume/t1_volume_tasks.py
```diff
@@ -0,0 +1,251 @@
...
+    return out_file
+
+
+@task
+@annotate({"return": {"out_status": None}})
+def check_volume_location_in_world_coordinate_system(
+    nifti_input: Union[PathLike, list],
+    bids_dir: PathLike,
+    modality: str = "t1w",
+    skip_question: bool = False,
```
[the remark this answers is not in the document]
omar-rifai: I kept this parameter as it was present in the legacy function but currently not dealing with clinica's command line argument. I'll note this for a separate task.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: Good point, I'll look into it.

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+        # Should never be reached
+        raise ValueError("sform_code has a value > 0, so method 3 cannot be used")
+
+    homogeneous_coord = np.concatenate(
+        (np.array(coordinates_vol), np.array([1])), axis=0
+    )
+    return np.dot(aff, homogeneous_coord)[0:3]
+
+
+def vox_to_world_space_method_3_bis(coordinates_vol, header):
```
NicolasGensollen: Types are missing here

clinica/pydra/t1_volume/t1_volume_utils.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+    This method relies on the same technique as method 3, but for images created by FreeSurfer (MGHImage, MGHHeader).
+    Args:
```
NicolasGensollen: Can we convert to Numpydoc style?

NicolasGensollen (commented):
Some small suggestions as I was doing a quick first pass on the code.

clinica/pydra/t1_volume/t1_volume_tasks.py
```diff
@@ -0,0 +1,251 @@
...
+
+from clinica.pydra.t1_volume.t1_volume_utils import (
+    get_world_coordinate_of_center,
+    is_centered,
+)
+
+
+@task
+@annotate({"return": {"out_file": PathLike}})
+def zip_nii(in_var: Union[PathLike, list], same_dir: bool = False) -> PathLike:
```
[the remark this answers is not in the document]
omar-rifai: Removed for now as it was not being used in this pipeline.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: Also not used here. Also removed for now.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: I'm not sure I agree, it seems at this point `nii_volume` has been used in other functions (the input comes from another function call). Granted, in the current use case it has never changed but if we are to use an exception it would rather be in the function that first receives the nifti list.

The assert here seems to function more as a sanity check to verity that our previous functions have not modified this input's type.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: I'll change as suggested but I feel your version is more elegant computationally but not necessarily simpler.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: I'd prefer to leave as is for it being more explicit. If you have a strong feeling about it, we can change.

test/unittests/pydra/test_t1volume.py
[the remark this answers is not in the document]
omar-rifai: We are using some elements in the input specs which are not there in the interface you reference.

clinica/pydra/t1_volume/t1_volume_utils.py
ghisvail: Be careful with `_`, this really means the variable is unpacked but **unused** later on. This is not the case here.

clinica/pydra/t1_volume/t1_volume_utils.py
omar-rifai: The issue with an error here is that we will be notifying the user about a problem with an input that he would have not provided himself. I'm not sure I understand the purpose of that?

> If that's indeed the underlying reason, then this is very scary. Functions that modifies their inputs are a red flag.
Exactly. It would be extremely scary and hence raising an error seem appropriate? 

I'll leave it for now and we can discuss in person if necessary.

ghisvail: > The issue with an error here is that we will be notifying the user about a problem with an input that he would have not provided himself. I'm not sure I understand the purpose of that?

I believe Nicolas and I are talking in the more general sense of error handling. Take you custom implementation of a divison function and the case where the denominator is `0`. You may decide to `assert`, on it or raise a `ZeroDivisionError` like Python does in its standard implementation. The latter provides greater freedom to implement your own error handling strategy (even a YOLO mode `try / except: pass`) as opposed to the former.

Now back to our use case, there is nothing that prevents us for getting the same behaviour of the current `assert` implementation with a `try / catch` block.

> Exactly. It would be extremely scary and hence crashing seems appropriate, no(?)

Ensuring we are not using functions that modify their inputs would be more of an improvement. Otherwise, we are just reproducing bad habits from the legacy Clinica code of doing defensive programming and shotgun parsing everywhere.

I am not willing to argue on this further to be honest.

omar-rifai: I'm also talking about the more general sense of error handling. 

With the example you provided the distinction I'm making is from whereby the `0`comes. If it is a user input then yes, error handling make sense as you raise a proper exception that the user can then address. On the other hand, if the value you are checking comes from inner processing, then the assert is here as a failsafe for the code not to continue.
That being said, I have no strong opinions against handling this with an error, I was just pointing out to the fact that the error message (saying the variabel is not a nifti file) will not be informative to the user at this point in the code, that is all. Will change.

On the point of legacy code: legacy code is not the scope of this PR and this PR was solely renaming and reusing utility functions.The flaws directly in the legacy code can and should be addressed if properly planned for. 

On a side note when I said "I'll leave it for now and we can discuss in person if necessary." it was to avoid a further reply in here.

omar-rifai:
Will modify this PR to use the original functions instead of copying code to make the scope of the PR more transparent.

clinica/iotools/utils/data_handling.py
```diff
@@ -1010,29 +1012,44 @@ def check_relative_volume_location_in_world_coordinate_system(
...
+    bids_dir: PathLike,
+    modality: str = "t1w",
+    skip_question: bool = False,
+) -> bool:
+    """Check if images are centered around the origin of the world coordinate
+    Parameters
+    ----
+    nifti_list: list
+        list of path to nifti files or path
+    bids_dir: str
```
NicolasGensollen, line 1025:
```suggestion
    bids_dir: PathLike
```

clinica/iotools/utils/data_handling.py
```diff
@@ -1010,29 +1012,44 @@ def check_relative_volume_location_in_world_coordinate_system(
...
+        path to bids directory associated with this check
+    modality: str
```
NicolasGensollen, line 1027:
```suggestion
    modality: str, optional
```

clinica/iotools/utils/data_handling.py
```diff
@@ -1010,29 +1012,44 @@ def check_relative_volume_location_in_world_coordinate_system(
...
+        the modality of the image
```
NicolasGensollen, line 1028:
```suggestion
        the modality of the image. Default='t1w'.
```

clinica/iotools/utils/data_handling.py
```diff
@@ -1010,29 +1012,44 @@ def check_relative_volume_location_in_world_coordinate_system(
...
+    skip_question: bool
```
NicolasGensollen, line 1029:
```suggestion
    skip_question: bool, optional
```

clinica/iotools/utils/data_handling.py
```diff
@@ -1010,29 +1012,44 @@ def check_relative_volume_location_in_world_coordinate_system(
...
+        if True, assume answer is yes
```
NicolasGensollen, line 1030:
```suggestion
        if True, assume answer is yes. Default=False.
```

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+    skip_question: bool = False,
+) -> bool:
+    """Check if images are centered around the origin of the world coordinate
+
+    Parameters
+    ----
+    nifti_input: PathLike
+        Path to nifti file
+
+    bids_dir: str
```
NicolasGensollen, line 21:
```suggestion
    bids_dir: PathLike
```

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+        path to bids directory associated with this check
+
+    modality: str
```
NicolasGensollen, line 24:
```suggestion
    modality: str, optional
```

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+        the modality of the image
```
NicolasGensollen, line 25:
```suggestion
        the modality of the image. Default="t1w".
```

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+
+    skip_question: bool
```
NicolasGensollen, line 27:
```suggestion
    skip_question: bool, optional
```

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+        if True, assume answer is yes
```
NicolasGensollen, line 28:
```suggestion
        if True, assume answer is yes. Default=False.
```

clinica/pydra/t1_volume/tissue_segmentation/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+from clinica.pydra.t1_volume.utils import initialize_tissues_spm_segment
+
+
+@clinica_io
+def t1volume_tissue_segmentation(name: str = "t1volume") -> Workflow:
+    """Workflow for tissue segmentation, bias correction and spatial normalization
+
+    Parameters
+    ----------
+    name : str
```
NicolasGensollen, line 19:
```suggestion
    name : str, optional
```

clinica/pydra/t1_volume/tissue_segmentation/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+        name of pipeline
```
NicolasGensollen, line 20:
```suggestion
        name of pipeline. Default="t1volume".
```

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,72 @@
...
+def initialize_tissues_spm_segment(
+    tissue_probability_map: Union[PathLike, None] = None
+) -> tuple:
+
+    """Prepare data structure for SPM segment interface
+
+    Parameters
+    ----------
+    tissue_probability_map: PathLike, optional
+        Path to the nifti tissue probability map
```
NicolasGensollen, line 14:
```suggestion
        Path to the nifti tissue probability map. Default=None.
```

NicolasGensollen (commented):
Some simple suggestions for the docstrings.

clinica/pydra/t1_volume/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+    ------
+    If volume is not centered on origin of the world coordinate system
+    """
+    from clinica.iotools.utils.data_handling import (
+        check_volume_location_in_world_coordinate_system,
+    )
+
+    nifti_list = [str(nifti_input)]
+
+    flag_centered = check_volume_location_in_world_coordinate_system(
```
NicolasGensollen, line 45: Why not simply:

```python
return check_volume_location_in.....()
```

?

clinica/pydra/t1_volume/tissue_segmentation/pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+
+    spm_segment = Nipype1Task(
+        name="SpmSegmentation",
+        interface=spm.NewSegment(),
+    )
+
+    spm_segment.inputs.write_deformation_fields = [True, True]
+
+    tissue_tuples = initialize_tissues_spm_segment()
+    spm_segment.inputs.tissues = tissue_tuples
```
NicolasGensollen, line 58:
```suggestion
    spm_segment.inputs.tissues = initialize_tissues_spm_segment()
```

clinica/pydra/t1_volume/tissue_segmentation/pipeline.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    spm_segment = Nipype1Task(
+        name="SpmSegmentation",
+        interface=spm.NewSegment(),
+    )
+
+    spm_segment.inputs.write_deformation_fields = [True, True]
+
+    tissue_tuples = initialize_tissues_spm_segment()
```
NicolasGensollen, line 57:
```suggestion
```

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,72 @@
+from os import PathLike
+from typing import Tuple, Union
+
+
+def initialize_tissues_spm_segment(
+    tissue_probability_map: Union[PathLike, None] = None
+) -> tuple:
+
```
NicolasGensollen, line 8:
```suggestion
```

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        "tissues" data structure for SPMSegment
+    """
+
+    from clinica.pipelines.t1_volume_tissue_segmentation.t1_volume_tissue_segmentation_utils import (
+        get_tissue_tuples,
+    )
+    from clinica.utils.spm import get_tpm
+
+    parameters = {}
+    parameters.setdefault("tissue_classes", [1, 2, 3])
```
NicolasGensollen, line 28: Why don't you set the values directly since you know the dictionary is empty? i.e:

```python
parameters["tissue_classes"] = [1, 2, 3]
...
```

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    parameters.setdefault("dartel_tissues", [1, 2, 3])
+    parameters.setdefault("save_warped_unmodulated", True)
+    parameters.setdefault("save_warped_modulated", False)
+
+    if tissue_probability_map:
```
NicolasGensollen, line 33: Maybe something like this ?

```python
parameters["tissue_probability_maps"] = tissue_probability_map if tissue_probability_map else get_tpm()
```

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        parameters.setdefault("tissue_probability_maps", tissue_probability_map)
+    else:
+        parameters.setdefault("tissue_probability_maps", get_tpm())
+
+    tissue_tuples = get_tissue_tuples(
+        parameters["tissue_probability_maps"],
```
NicolasGensollen, line 39: I don't understand why you need to go through the parameters dictionary here? Why not just return the tuple with the values you set as defaults?

clinica/pydra/t1_volume/tissue_segmentation/cli.py
```diff
@@ -0,0 +1,28 @@
...
+
+pipeline_name = "pydra-t1vol-ts"
+
+
+@click.command(name=pipeline_name, hidden=True)
+@cli_param.argument.bids_directory
+@cli_param.argument.caps_directory
+def cli(
+    bids_directory: str,
+    caps_directory: str,
```
NicolasGensollen, line 15: We are losing all the options from the current CLI here, right? (i.e. things like `tissue_classes`, `dartel_tissues`, `tissue_probability_maps`...). You could build the parameters dictionary here, and pass it to the workflow such that the provided values will be used. Then, using the `set_default` method like you did makes perfect sense. WDYT?

NicolasGensollen (commented):
Looks good, thanks @omar-rifai !
I made a few suggestions, mostly on formatting.
I have one concern about the parameters of the CLI though (see comment below).

clinica/pydra/t1_volume/tissue_segmentation/cli.py
omar-rifai: Right. I wanted to start with an implementation in its simplest form without the argument because the idea was to have argument construction and "Clinica integration" left for another decorator. 

That being said I'll add the arguments to the CLI in the meantime as you suggested.

clinica/pydra/t1_volume/utils.py
ghisvail: Which is:

`parameters["tissue_probability_maps"] = tissue_probability_map or get_tpm()`

NicolasGensollen: Thanks ! Wrote this one too fast... 😅

omar-rifai:
@NicolasGensollen, added the pipeline specific arguments as suggested.

clinica/pydra/t1_volume/utils.py
```diff
@@ -0,0 +1,62 @@
...
+        parameters["tissue_classes"],
+        parameters["dartel_tissues"],
+        parameters["save_warped_unmodulated"],
+        parameters["save_warped_modulated"],
+    )
+    return tissue_tuples
+
+
+def init_input_node(bids_name: str) -> Tuple[str, str]:
```
NicolasGensollen, line 38: I might be missing something, but I don't see where you use this function?

NicolasGensollen (commented):
Thanks for addressing my points @omar-rifai !
This LGTM. Just a small suggestion and a question below.

clinica/pydra/t1_volume/utils.py
omar-rifai: You are right, the purpose of the node in the original pipeline was to print informative messages about the subjects to be executed. 

I suppose if we want to add something like this in Pydra we ought to pass by the engine input node instead of doing it pipeline by pipeline. Removed.

omar-rifai:
Hello @NicolasGensollen, please merge if the latest modifications address your comments to be able to rebase in #739. Thanks !

NicolasGensollen (approved):

PR #727, opened 2022-08-11
Fix warnings in `merge-tsv`

14thibea:
`merge-tsv` raises a lot of warning from pandas which prevents other messages from being seen.
This PR solves these issues.

### DIFF
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -659,7 +659,7 @@ def compute_missing_mods(bids_dir, out_dir, output_prefix=""):
                             row_to_append_df[m] = pd.Series("0")
                             mmt.add_missing_mod(ses, m)
 
-            missing_mods_df = missing_mods_df.append(row_to_append_df)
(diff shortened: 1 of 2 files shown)


ghisvail:
> raises a lot of warning

Just for additional context, these warnings result from the [deprecation](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.4.0.html#whatsnew-140-deprecations-frame-series-append) of `pandas.DataFrame.append` in favour of `pandas.concat` since Pandas v1.4.

ghisvail:
@14thibea what's the issue with the index though?

14thibea:
It is not possible to concat a DataFrame with single index with a multi-indexed DataFrame.
And I was trying to concat `ses_df` (indexed by  (participant_id, session_id)) with the content associated with each atlas which is a row indexed by 0.

14thibea:
also as `merge-tsv` do not write the numbered index when using a BIDS only, I removed it when a CAPS must also be merged (this is why the test is not working, in ref data there is still this useless column "Unnamed: 0")

omar-rifai:
Thanks @14thibea.

PR #729, opened 2022-08-23
Pydra implementation of PETVolume

NicolasGensollen:
This is a draft for the Pydra implementation of the current PETVolume pipeline.
WIP....


clinica/pydra/pet_volume/pipeline.py
```diff
@@ -0,0 +1,305 @@
...
+    from clinica.utils.atlas import PET_VOLUME_ATLASES
+    from clinica.utils.group import check_group_label
+
+    parameters.setdefault("group_label", None)
+    check_group_label(parameters["group_label"])
+    if "acq_label" not in parameters.keys():
+        raise KeyError("Missing compulsory acq_label key in pipeline parameter.")
+    parameters.setdefault("pvc_psf_tsv", None)
+    parameters["apply_pvc"] = parameters["pvc_psf_tsv"] is not None
+    parameters.setdefault("mask_tissues", [1, 2, 3])
```
omar-rifai, line 41: Isn't it redundant to set the defaults here and in the cli function signature?

clinica/pydra/interfaces.py
```diff
@@ -3,9 +3,59 @@
...
+    def _list_outputs(self):
+        from clinica.utils.inputs import clinica_file_reader
+        from clinica.utils.participant import get_subject_session_list
+
+        sessions, subjects = get_subject_session_list(
+            self.inputs.base_dir,
+            is_bids_dir=False,
+        )
+        return {
+            k: clinica_file_reader(
```
omar-rifai, line 46: There should be a distinction between files read by `clinica_file_reader` and those read by `clinica_group_reader`(templates for example)

clinica/pydra/pet_volume/pipeline.py
```diff
@@ -0,0 +1,306 @@
...
+            ("_graph_checksums", Any),
+            ("T1w", str, {"mandatory": True}),
+            ("pet", str, {"mandatory": True}),
+            (
+                "mask_tissues",
+                dict,
+                {"tissue_number": 1, "modulation": False},
+                {"mandatory": True},
+            ),
+            ("flow_fields", dict, {"group_label": "group"}, {"mandatory": True}),
```
omar-rifai, line 80: The name of the `group_label` should not be hardcoded but instead be a user input which depends on the name provided when running `t1_volume`

omar-rifai (commented):
Just a couple of comments from issues I encountered while trying to run the code. Maybe you already figured them out.

clinica/pydra/interfaces.py
NicolasGensollen: Yes, I'm working on that atm. Thanks!

clinica/pydra/pet_volume/pipeline.py
NicolasGensollen: Yes, I fixed that already locally but didn't commit it yet.

omar-rifai:
Hi @NicolasGensollen , is this ready for review?

NicolasGensollen:
> Hi @NicolasGensollen , is this ready for review?

Yes, I think it's ready

omar-rifai:
Hello @NicolasGensollen, I'm running into an error when trying to test the pipeline. `'int' object is not subscriptable` on line 132 in `pipeline.py`. 

fwhm seems to be of type `tuple` instead of list of list.

NicolasGensollen:
@omar-rifai I finally managed to fix the issues and made the pipeline run on the CI data.

I'm not sure why you're getting an error with `fwhm`. I don't think it's a huge deal and will take a closer look.

In the mean time, if you want to experiment with it, here's how you can currently run it:

In the PETVolume folder of the CI data, run the following script (or use the equivalent CLI if you prefer):

```python
import os
from pathlib import Path
import clinica.pydra.engine_utils as pydra_utils
import clinica.pydra.pet_volume.pipeline as pydra_pet_volume

currdir = os.path.dirname(os.path.abspath(__file__))

bids_dir = Path(currdir) / "bids"
caps_dir = Path(currdir) / "caps"

parameters = {
    "group_label": "UnitTest",
    "acq_label": "fdg",
    "suvr_reference_region": "pons",
    "pvc_psf_tsv": Path(currdir) / "pvc_psf.tsv",
    "mask_tissues": (1, 2, 3),
    "mask_threshold": 0.3,
    "pvc_mask_tissues": (1, 2, 3),
    "smooth": [6, 8],
    "skip_question": False,
}

pipeline = pydra_pet_volume.build_core_workflow(
    name="pet-volume-pydra",
    input_dir=bids_dir,
    output_dir=caps_dir,
    parameters=parameters,
)

results = pydra_utils.run(pipeline)
```

where `pvc_psf.tsv` is a file, following the conventions defined in the Clinica's docs, I added with random values for `psf_x`, `psf_y`, and `psf_z` for testing purposes.

Let me know if you're able to make it run.

omar-rifai:
Hello @NicolasGensollen, thanksfor the default parameters. I tried running again but I still have the same issue with fwhm. I think there is definitely something to look into there as I can see that:

1- In `pipelines/cli_param/options.py`the default is set to `[8],`
2- The default in the PR (in `pet_volume_cli.py`) is a tuple (8,8) 
3- When invoking the command line with pet-volume, it asks for an `int`: "-s, --smooth INTEGER          Specify the different isomorphic FWHM in millimeters to smooth the image.  [default: 8]"
4- In  `build_smoothing_workflow.py` I'm getting a value of `(8)` by default (if I don't set anything when invoking) but it expects a List of Lists of ints.

NicolasGensollen:
> Hello @NicolasGensollen, thanksfor the default parameters. I tried running again but I still have the same issue with fwhm. I think there is definitely something to look into there as I can see that:
> 
> * In `pipelines/cli_param/options.py`the default is set to [8],
> * The default in the PR (in `pet_volume_cli.py`) is a tuple (8,8)
> * When invoking the command line with pet-volume, it asks for an int: "-s, --smooth INTEGER          Specify the different isomorphic FWHM in millimeters to smooth the image.  [default: 8]"
> * In  `build_smoothing_workflow.py` I'm getting a value of (8) by default (which corresponds to the value in `options.py`) but it expects a List of Lists of ints.

Hi @omar-rifai 

I changed the default value of this PR (from `(8., 8.)` to `[8.]`). I'm not sure why it was still a length 2 tuple here..

The function `_sanitize_fwhm()` is responsible for parsing the input into a list of lists of floats which is then provided to the `Smooth` interface. I'm basically allowing specifying the kernel in three different ways:

- A scalar:

```python
>>> _sanitize_fwhm(3.0)
[[3.0, 3.0, 3.0]]
```

- A list of floats:

```python
>>> _sanitize_fwhm([3.0, 2.0])
[[3.0, 3.0, 3.0], [2.0, 2.0, 2.0]]
```

- A list of lists of floats

```python
>>> _sanitize_fwhm([[3.0, 2.0, 1.0], [2.0, 2.0, 1.0]])
[[3.0, 2.0, 1.0], [2.0, 2.0, 1.0]]
```

Could you try running this again and see if it works for you ?

clinica/pydra/pet_volume/pet_volume_cli.py
```diff
@@ -0,0 +1,115 @@
...
+    bids_directory: str,
+    caps_directory: str,
+    group_label: str,
+    acq_label: str,
+    suvr_reference_region: Optional[str] = None,
+    pvc_psf_tsv: Optional[str] = None,
+    mask_tissues: List[int] = (1, 2, 3),
+    mask_threshold: float = 0.3,
+    pvc_mask_tissues: List[int] = (1, 2, 3),
+    smooth: List[float] = [8.0],
```
MatthieuJoulot, line 68: On the nipype version of the pipeline, the value of `smooth` is `List[int]= (8, 8)`. Is this change right. TBH, i do not know what it does

MatthieuJoulot: And if it is a single value, is it normal to use a list ?

clinica/pydra/pet_volume/pet_volume_cli.py
```diff
@@ -0,0 +1,115 @@
...
+        "pvc_mask_tissues": pvc_mask_tissues,
+        "smooth": smooth,
+        "skip_question": yes,
+    }
+    pipeline = pydra_pet_volume.build_core_workflow(
+        name="pet-volume-pydra",
+        input_dir=bids_directory,
+        output_dir=caps_directory,
+        # tsv_file=subjects_sessions_tsv,
+        # base_dir=working_directory,
```
MatthieuJoulot, line 108: I guess you commented these two lines for testing ? It seems like they shouldn't be commented.

NicolasGensollen:

> And our favorite subject: unit test 😉!! It feels like we could have some for some of the functions in `clinica/pydra/pet_volume/tasks.py`.Maybe the functions are too complex to have them ? WDYT ?

This made me realize that I originally duplicated functions from `pet_volume_utils` to improve their documentation and turn them into Pydra tasks. This appeared as if I had implemented these functions and it is clearly not ideal as it is duplicating code between engines...

I added my documentation and refactoring changes to the original (Nipype engine) functions, and defined the Pydra tasks as simple wrappers that import and call the corresponding functions.
 
Now, I agree that, in the long run, we should have unit tests for these functions. But, as they were not implemented in this PR, I would prefer to do this later as part of our continuous effort of adding unit tests to the whole code base.

I think you can try this PR again. I've tried it with both engines and it seems to run fine.

clinica/pipelines/pet_volume/pet_volume_utils.py
```diff
@@ -23,35 +23,105 @@ def init_input_node(pet_nii):
...
 
+
+def _load_tissues(tissues: list):
+    """Aggregates the image data contained in the tissue images provided.
+
+    Parameters
+    ----------
+    tissues : List[PurePath]
+        List of tissue images to aggregate.
+
```
MatthieuJoulot, line 41: The input parameter tissue is of type `list` here you've put `List[PurePath]`

clinica/pipelines/pet_volume/pet_volume_utils.py
```diff
@@ -70,48 +140,81 @@ def apply_binary_mask(image, binary_mask):
...
+    pvc_method : str
+        Name of the PVC method. This will be concatenated
+        with the `pet_image` filename.
+
+    Returns
+    -------
+    pet_pvc_path : str
+        Name for the PET PVC interface.
+    """
```
MatthieuJoulot, line 195: Looks like this function could have an example ! (since it's not dealing with images)

clinica/pydra/pet_volume/tasks.py
```diff
@@ -0,0 +1,110 @@
+import typing as ty
+from pathlib import PurePath
+
+from pydra.mark import annotate, task
+
```
MatthieuJoulot, line 5: More of an open question here, and not a suggestion/comment. The docstring is very light for the functions in this file. Is it intended as we are only creating a task out of a function, which should be documented ? (and may very well not be at this time).

MatthieuJoulot (changes requested):
Just a few questions and comments.
I also tested both pydra-pet-volume and pet-volume, both of them work !

clinica/pipelines/pet_volume/pet_volume_utils.py
NicolasGensollen: Thanks! Addressed in [6833496](https://github.com/aramis-lab/clinica/pull/729/commits/6833496c4a5bf53e3b252598d4adf1c474756960)

clinica/pipelines/pet_volume/pet_volume_utils.py
NicolasGensollen: Never hurts indeed ! I added an example in [2bc4ed7](https://github.com/aramis-lab/clinica/pull/729/commits/2bc4ed7b2292f2ffc83c164d3b9cd393ab623b10)

clinica/pydra/pet_volume/tasks.py
NicolasGensollen: These functions are not supposed to implement anything, they are simple wrappers in order to turn the original functions into Pydra tasks.
Because of this, the doctstrings would basically be copy pasted which I don't think is a good solution because they will be harder to maintain and they will diverge at some point as people forget to maintain them in synchrony.

In [63fc0b1](https://github.com/aramis-lab/clinica/pull/729/commits/63fc0b18545707f962524b79b7a30405dc219b8d), I proposed to add a note encouraging the user to refer to the original function documentation if needed.
Let me know what you think of this possible solution.

MatthieuJoulot: > These functions are not supposed to implement anything, they are simple wrappers in order to turn the original functions into Pydra tasks.
Because of this, the doctstrings would basically be copy pasted which I don't think is a good solution because they will be harder to maintain and they will diverge at some point as people forget to maintain them in synchrony.

Yup makes sense. If i understand correctly, we'd have to move the documentation once we move on from nipype (which is the reason for which you've 'only' wrapped the function ?)

> Let me know what you think of this possible solution.

At least for know, It looks like a good solution ! Maybe even after we've moved on from nipype :)

NicolasGensollen: >If i understand correctly, we'd have to move the documentation once we move on from nipype (which is the reason for which you've 'only' wrapped the function ?)

Either move the docs or keep the function implementation separated from the Pydra task. The point is to avoid duplication (of the code and the docs).

MatthieuJoulot (approved):

NicolasGensollen:
Thanks for the reviews @MatthieuJoulot @omar-rifai ! 🙏 
Merging...

PR #731, opened 2022-08-31
CAPS Reader for Pydra Engine

NicolasGensollen:
This PR proposes a possible implementation for the CAPS reader to be used with the new Pydra engine. Originally implemented as part of #729 but moved in a separate PR for clarity.

This is very drafty and not tested, but should give a decent idea of the proposed architecture.
Comments and feedbacks are more than welcome as there is clearly room for improvement ! ;)

## Example

For PETVolume, this is more or less what the input specs would look like:

```python
input_spec = pydra.specs.SpecInfo(
     name="Input",
     fields=[
         ("t1W", str, {"mandatory": True}),
         ("pet", str, {"mandatory": True}),
         (
             "mask_tissues",
             dict,
             {"tissue_number": 1, "modulation": False},
             {"mandatory": True},
         ),
         ("flow_fields", dict, {"group_label": "group"}, {"mandatory": True}),
         ("dartel_template", dict, {"group_label": "group"}, {"mandatory": True}),
         ("pvc_mask_tissues", dict, {"tissue_number": 1}, {"mandatory": False}),
     ],
     bases=(pydra.specs.BaseSpec,),
 )

wf = Workflow(
     name,
     input_spec=["foo"],  # # this should be input_spec but error...
)
wf.input_spec = input_spec  # FIXME
```

### DIFF
--- a/clinica/pydra/engine.py
+++ b/clinica/pydra/engine.py
@@ -60,20 +60,27 @@ def build_input_workflow(pipeline: Workflow, core_workflow: Workflow) -> str:
     caps_dict_inputs = pu.list_dict_in_fields(core_workflow)
     caps_query_dict = pu.caps_query(caps_dict_inputs)
 
-    input_workflow = Workflow(name="input_workflow", input_spec=["input_dir"])
-
-    input_workflow.inputs.input_dir = pipeline.lzin.input_dir
-
+    input_workflows = dict()
     for reader, query in zip(
         [bids_reader, caps_reader], [bids_query_dict, caps_query_dict]
     ):
+        name = reader.__name__
+        input_workflow = Workflow(
+            name=f"input_workflow_{name}",
+            input_spec=["input_dir"],
+        )
+        input_workflow.inputs.input_dir = pipeline.lzin.input_dir
         input_workflow = add_input_task(input_workflow, reader, query)
         pipeline.add(input_workflow)
+        input_workflows[name] = input_workflow
 
     # connect input workflow to core workflow
+    for field, _ in bids_query_dict.items():
+        read_data = getattr(input_workflows["bids_reader"].lzout, field)
+        setattr(core_workflow.inputs, field, read_data)
 
-    for field in list_core_inputs:
-        read_data = getattr(input_workflow.lzout, field)
+    for field, _ in caps_query_dict.items():
+        read_data = getattr(input_workflows["caps_reader"].lzout, field)
         setattr(core_workflow.inputs, field, read_data)
 
     return pipeline
--- a/clinica/pydra/engine.py
+++ b/clinica/pydra/engine.py
@@ -1,10 +1,11 @@
 import functools
 from os import PathLike
+from typing import Callable
 
 from pydra import Workflow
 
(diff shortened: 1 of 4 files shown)


omar-rifai:
Hello @NicolasGensollen, if you want to update this PR with the latest changes in #729 we can merge this as a first iteration on the reader.

NicolasGensollen:
@omar-rifai I rebased and picked from #729 the work specific to the `CAPSReader`

omar-rifai:
> @omar-rifai I rebased and picked from #729 the work specific to the `CAPSReader`

Great thanks, I'll try to have a look this afternoon for a review .

NicolasGensollen:
> > @omar-rifai I rebased and picked from #729 the work specific to the `CAPSReader`
> 
> Great thanks, I'll try to have a look this afternoon for a review .

Great thanks ! 👍
I'll leave it like this for a first round, but I realize that it'd already benefit from some refactoring...
It also needs some tests of course ! ;)

clinica/pydra/engine.py
```diff
@@ -19,8 +20,9 @@ def run_wrapper(
 
         pipeline = Workflow(
             name=name,
-            input_spec=["input_dir"],
-            input_dir=input_dir,
+            input_spec=["bids_dir", "caps_dir"],
+            bids_dir=input_dir,
```
omar-rifai, line 24: I suppose it's better to keep `bids_dir` instead of `input_dir` as the inputs also come from the CAPS

clinica/pydra/engine.py
```diff
@@ -52,52 +54,91 @@ def build_input_workflow(pipeline: Workflow, core_workflow: Workflow) -> str:
     """
 
     field = ""
+    readers = []
+    queries = []
 
     list_core_inputs = pu.list_in_fields(core_workflow)
```
omar-rifai, line 60: don't all the fields get added with both calls `list_in_fields` and `list_dict_in_fields`? The condition for adding the elements doesn't seem restrictive

clinica/pydra/engine.py
```diff
@@ -52,52 +54,91 @@ def build_input_workflow(pipeline: Workflow, core_workflow: Workflow) -> str:
...
+            except AttributeError:
+                raise AttributeError(
+                    "Workflow has no 'caps_dir' input. Please verify your input specifications."
+                )
+        input_workflow = add_input_task(input_workflow, reader, query)
+        pipeline.add(input_workflow)
+        input_workflows[name] = input_workflow
 
     # connect input workflow to core workflow
+    for field, _ in bids_query_dict.items():
```
omar-rifai, line 98: Maybe I'm missing something but can't we add these in inside the previous loop? That way we can propbably avoid maintaining the reader list and input_workflows dictionary?

omar-rifai (commented):
Hi @NicolasGensollen, just a couple of comments for readibility. If this works for #729, we can merge and iteratita on later PR for improvements.

omar-rifai:
Will merge this initial implementation and we'll iterate as needed in further pipelines.

PR #737, opened 2022-09-05
Enh pydra run error msg

omar-rifai:
Small enhancement to the pydra run function to print error message. For debugging purpose.

### DIFF
--- a/clinica/pydra/engine_utils.py
+++ b/clinica/pydra/engine_utils.py
@@ -47,20 +47,28 @@ def bids_query(keys: list) -> dict:
 
     return {key: bids_keys_available[key] for key in keys}
 
-
 def run(wf: Workflow) -> None:
+    import re
 
-    """Execute a Pydra workflow
+    try:
+        with Submitter(plugin="cf") as submitter:
+            submitter(wf)
+    except Exception as e:
+        path = re.search("\/.*\.pklz", str(e))
+        if path:
+            path = path.group(0)
+            read_error(path)
+        else:
+            print("Exception:", e)
+        return str(e)
 
-    Parameters
-    ----------
-        wf : Workflow
-            The workflow to execute
-    """
+    results = wf.result(return_inputs=False)
+    return results
 
-    with Submitter(plugin="cf") as submitter:
-        submitter(wf)
 
-    results = wf.result(return_inputs=True)
-    # @TODO: decide where to store results
-    print(results)
+def read_error(path):
+    import cloudpickle as cp
+
+    with open(path, "rb") as fp:
+        err = cp.load(fp)
+        print(err["error message"])
--- a/clinica/pydra/engine_utils.py
+++ b/clinica/pydra/engine_utils.py
@@ -47,6 +47,7 @@ def bids_query(keys: list) -> dict:
 
     return {key: bids_keys_available[key] for key in keys}
 
+
 def run(wf: Workflow) -> None:
     import re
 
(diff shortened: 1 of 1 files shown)


clinica/pydra/engine_utils.py
```diff
@@ -49,18 +49,27 @@ def bids_query(keys: list) -> dict:
 
 
 def run(wf: Workflow) -> None:
-
-    """Execute a Pydra workflow
```
NicolasGensollen, line 53: Why did you remove the docstring?

clinica/pydra/engine_utils.py
```diff
@@ -49,18 +49,27 @@ def bids_query(keys: list) -> dict:
...
+    results = wf.result(return_inputs=False)
+    return results
+
+
+def read_error(path):
+    import cloudpickle as cp
+
+    with open(path, "rb") as fp:
+        err = cp.load(fp)
+        print(err["error message"])
```
NicolasGensollen, line 75: I would either return the error, or change the name of this function to `print_error`. WDYT?

NicolasGensollen (commented):
Thanks @omar-rifai ! This will make debugging easier indeed ! 👍 
Couple suggestions below.

omar-rifai:
Thanks @NicolasGensollen, I was thinking this could be temporary for debugging purposes but might as well have the right decorations. 

Should be good now.

clinica/pydra/engine_utils.py
```diff
@@ -48,19 +52,53 @@ def bids_query(keys: list) -> dict:
...
-    with Submitter(plugin="cf") as submitter:
-        submitter(wf)
+    import cloudpickle as cp
 
-    results = wf.result(return_inputs=True)
-    # @TODO: decide where to store results
-    print(results)
+    with open(path, "rb") as fp:
+        err = cp.load(fp)
+        return err["error message"]
```
NicolasGensollen, line 104:
```suggestion
    return err["error message"]
```

clinica/pydra/engine_utils.py
```diff
@@ -48,19 +52,53 @@ def bids_query(keys: list) -> dict:
...
+        with Submitter(plugin="cf") as submitter:
+            submitter(wf)
+    except Exception as e:
+        path = re.search("\/.*\.pklz", str(e))
+        if path:
+            path = path.group(0)
+            print(read_error(path))
+        else:
+            print("Exception:", e)
+        return str(e)
```
NicolasGensollen, line 80: I would raise the error instead of printing and returning it. This way we avoid the weird return type and make sure the execution stops. WDYT?

omar-rifai: I ended up doing a return directly to avoid the traceback messages. Seems to be the most readable version.

NicolasGensollen: @omar-rifai I was thinking about something along those lines (I haven't tested, there are probably some mistakes to fix and some changes to make...):

```python
class WorkflowExecutionError(Exception):
    def __init__(self, message="Error running the Workflow."):
        import re
        self.message = message
        path = re.search("\/.*\.pklz", self.message)
        if path:
            self.message = read_error(path.group(0))
        super().__init__(self.message)

def read_error(path: PathLike) -> str:
    with open(path, "rb") as fp:
        err = cp.load(fp)
    return err["error message"][:-1]

try:
    with Submitter(plugin="cf") as submitter:
        submitter(wf)
except Exception as e:
    raise WorkflowExecutionError(str(e))
```

omar-rifai: As discussed offline, we can leave this suggestion for a release version and keep the current version for debugging.

NicolasGensollen (approved):
Let's try it like this for debugging.

PR #741, opened 2022-09-08
Update bids readme 

### LINKED ISSUE #721: default readme in BIDS dataset after conversion
hey

We (BIDS maintainers) are trying to see if we can improve the quality of the average README in BIDS dataset. Those tend to be pretty thin and are missing some information we would like to see there even if it can be formalized "yet" into the BIDS specification.

Hoping to leverage the bids converters to help to do.

So I am going around the different BIDS converters repo to see if we can start improving the  default README they put in a the BIDS dataset.

There is template that is provided in the BIDS starter kit: https://github.com/bids-standard/bids-starter-kit/blob/main/templates/README.MD

Some possible options:
1. use this template is
2. use a subset of it
3. add a link to the template README
4. something else (a mix of 2 and 3?)

What do you think? I can try to open a PR if you want.

Also let us know if you think that the template README in the starter kit could / should be improved .

Feel free to disregard if you are already using some version of the template README.

MatthieuJoulot:
Following the issue #721, this PR increases the amount of information written inside the README. On top of the information related to Clinica, information related to the dataset are written as well.

Closes #721


omar-rifai:
Thanks @MatthieuJoulot, there are some issue with the converters for OASIS3, NIFD, and UKB which should be fixed. Try running the tests locally for debugging.

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,72 @@
...
+    """Model representing a BIDS ReadMe.
+
+    See
+    """
+
+    class DatasetType(str, Enum):
+        raw = "raw"
+        derivative = "derivative"
+
+    name: str
```
omar-rifai, line 22: The class attributes should reflect what is needed for the BIDSReadme

MatthieuJoulot: Done !

omar-rifai: Currently it still doesn't really. I still see a `bids_name` and `bids_version` which make sense for ther `dataset_description.json` but are not attributes of the readme.

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,57 @@
...
+
+    See
+    """
+
+    name: str
+    link: str
+    desciption: str
+
+    def write(self, to: IO[str]):
+        import clinica
```
NicolasGensollen, line 23:
```suggestion
        from importlib.metadata import version
```

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,57 @@
...
+
+        to.write(
+            f"This BIDS directory was generated with Clinica v{clinica.__version__}.\n"
```
NicolasGensollen, line 26:
```suggestion
            f"This BIDS directory was generated with Clinica v{version('clinica')}.\n"
```

clinica/iotools/bids_utils.py
```diff
@@ -509,7 +515,8 @@ def write_modality_agnostic_files(
         bids_version: BIDS version if different from the version supported by Clinica.
     """
     _write_bids_dataset_description(study_name, bids_dir, bids_version)
-    _write_readme(bids_dir)
+    # _write_readme(bids_dir)
```
NicolasGensollen, line 518:
```suggestion
```

clinica/iotools/bids_utils.py
```diff
@@ -467,15 +467,20 @@ def _write_bids_dataset_description(
...
-def _write_readme(bids_dir: Union[str, Path]) -> None:
-    """Write `README`file at the root of the BIDS directory."""
-    import clinica
+def _write_readme(
+    study_name: str,
+    data_dict: dict,
+    bids_dir: Union[str, Path],
+    bids_version: Optional[str] = None,
+) -> None:
+    """Write `dataset_description.json` at the root of the BIDS directory."""
```
NicolasGensollen, line 476: I don't understand this docstring. Isn't this function writing the README?

clinica/iotools/bids_utils.py
```diff
@@ -467,15 +467,20 @@ def _write_bids_dataset_description(
...
 
+    bids_desc = BIDSReadme(
+        name=study_name, link=data_dict["link"], desciption=data_dict["desc"]
+    )
     with open(Path(bids_dir) / "README", "w") as f:
-        f.write(
-            f"This BIDS directory was generated with Clinica v{clinica.__version__}.\n"
-            f"More information on https://www.clinica.run\n"
-        )
+        bids_desc.write(to=f)
```
NicolasGensollen, line 483: This feels a little strange. I'd expect a `write` method which would take a file name as input and the method would handle the file opening and closing. WDYT?

NicolasGensollen (commented):
Thanks for working on this @MatthieuJoulot !
I made a quick first pass and some suggestions.

clinica/iotools/bids_utils.py
ghisvail: Traditionnally, `read` / `write` are for buffers, `load` / `save` for files. Agreed, the scientific community mixed both, hence the potential for confusion.

Having `write` handle opening and closing the file improves usage convenience in exchange for additionnal checks (is the path valid and writeable), resource management and reduced testability. If `write` only accepts buffers, then responsibilities are lifted to the caller, but the implementation can be easility tested in memory.

The other alternative, which most libraries do, is to accept both (a buffer or a path). Since this code is only used internally by Clinica, I'd favour robustness and testability other convenience, but opinions may vary 👍

NicolasGensollen: I'm not really convinced by the argument that this is only used internally. AFAICT the `BIDSReadme` class is public as well as its `write` method. Even if most users won't have to use this class, some future contributors might.
I'm fine with both approaches, but as you wrote, most scientific libraries (Pandas, Numpy...) traditionally accept buffer and path with loading/reading/saving/writing methods.
If we decide to implement an uncommon design by only supporting buffers, I think we should at least document it better as most users and contributors will probably expect the usual buffer+path behavior.

ghisvail: > I'm not really convinced by the argument that this is only used internally.

I meant Clinica is not used as a library, so none of these interfaces are for public consumption. Unlike say the `numpy` and `pandas` loaders and writers.

> If we decide to implement an uncommon design by only supporting buffers.

I would not label this as uncommon. `json.load` from the standard library works on file objects only.

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,57 @@
...
+converter = make_converter()
+
+converter.register_unstructure_hook(
+    BIDSReadme,
+    make_dict_unstructure_fn(
+        BIDSReadme,
+        converter,
+        **{a.name: override(rename=_rename(a.name)) for a in fields(BIDSReadme)},
+    ),
+)
```
ghisvail, line 57: None of this should be required actually. It was useful for the dataset description class to implement some additional (de)serialization logic.

clinica/iotools/bids_utils.py
```diff
@@ -467,15 +467,20 @@ def _write_bids_dataset_description(
...
+def _write_readme(
+    study_name: str,
+    data_dict: dict,
+    bids_dir: Union[str, Path],
+    bids_version: Optional[str] = None,
+) -> None:
+    """Write `README` at the root of the BIDS directory."""
+    from clinica.iotools.bids_readme import BIDSReadme
 
+    bids_desc = BIDSReadme(
```
ghisvail, line 479: Might as well use the same name. It's a readme, let the variable named readme too.

```suggestion
    bids_readme = BIDSReadme(
```

clinica/iotools/bids_utils.py
```diff
@@ -496,6 +501,7 @@ def _write_bidsignore(bids_dir: Union[str, Path]) -> None:
 
 def write_modality_agnostic_files(
     study_name: str,
+    data_dict: dict,
```
ghisvail, line 504: We should find a more explicit name than `data_dict`. I am usually not a big fan of puting `dict` or `list` suffixes to variable names because they are redundant with the type.

Instead, I'd propose something like `readme_data` or `readme_content`.

clinica/iotools/bids_utils.py
```diff
@@ -467,15 +467,20 @@ def _write_bids_dataset_description(
...
 
 
-def _write_readme(bids_dir: Union[str, Path]) -> None:
-    """Write `README`file at the root of the BIDS directory."""
-    import clinica
+def _write_readme(
+    study_name: str,
+    data_dict: dict,
+    bids_dir: Union[str, Path],
+    bids_version: Optional[str] = None,
```
ghisvail, line 474: This is not used and should be removed.

ghisvail: Unless we want to specify the minimum BIDS version of the dataset in the readme maybe?

clinica/iotools/bids_utils.py
ghisvail: From PyCharm: "Missing parameter data_dict in docstring"

test/unittests/iotools/test_bids_utils.py
```diff
@@ -122,9 +128,35 @@ def test_write_modality_agnostic_files(tmp_path):
...
+):
+    """Test function `_write_bids_readme`.
+
+    .. note::
+        Tested independantly for convenience since it takes
+        a different set of input parameters.
+
+    """
+    from clinica.iotools.bids_utils import _write_readme
+
```
ghisvail, line 156: Already imported at the top of the module

```suggestion
```

test/unittests/iotools/test_bids_utils.py
```diff
@@ -122,9 +128,35 @@ def test_write_modality_agnostic_files(tmp_path):
...
+    data_dict = {"link": "", "desc": ""}
```
ghisvail, line 157: I don't think testing empty values is right. Since we need to ensure the templating of the readme works, we need to put distinctive values for the fields.

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,57 @@
+from enum import Enum
+from typing import IO
+
+from attrs import define, fields
+from cattr.gen import make_dict_unstructure_fn, override
+from cattr.preconf.json import make_converter
+
+BIDS_VERSION = "1.7.0"
```
ghisvail, line 8: Not needed, unused.

```suggestion
```

ghisvail (changes requested):
Good first iteration, a few comments and suggestions for next.

clinica/iotools/bids_utils.py
MatthieuJoulot: Alright, I'll keep it that way for now.

clinica/iotools/bids_utils.py
MatthieuJoulot: It is called by the dataset description though

ghisvail: `bids_version` is not used in this function

ghisvail (approved):

clinica/iotools/bids_readme.py
```diff
@@ -0,0 +1,28 @@
+from typing import IO
+
+from attrs import define
+
+
+@define
+class BIDSReadme:
+    """Model representing a BIDS ReadMe.
+
+    See
```
NicolasGensollen, line 10: The docstring is incomplete

NicolasGensollen (approved):
LGTM (besides the incomplete docstring).

PR #742, opened 2022-09-08
Enh remove split from engine

omar-rifai:
After discussion with @NicolasGensollen, it seems easier to handle the splits and combines directly inside the core workflows instead of the readers. This PR proposes to update the engine to reflect this proposal.


NicolasGensollen (approved):
Thanks @omar-rifai !
I think this is a slightly better solution than what I did in #729 (writing the split key in a new attribute and split in the engine). So let's go with this solution.

PR #744, opened 2022-09-09
Fix vox_to_world_space_method_1

### LINKED ISSUE #743: T1Volume doesn't work with OASIS3 data
Hi ! :)

**Describe the bug**

I was trying to run T1Volume pipeline with OASIS3 data and it is not working (but was working with data from ADNI). The problem comes from this function : 

Capture d’écran 2022-09-08 à 17 01 10

**To Reproduce**
Run T1Volume with data from OASIS3

**Expected behavior**
Crash in vox_to_world_space_method_1()

**Screenshots**
Capture d’écran 2022-09-08 à 16 56 24

Good luck :)

MatthieuJoulot:
As pointed out in #743, `t1-volume` doesn't work with data from `OASIS-3`. It seems that the function `vox_to_world_space_method_1()` was missing some brackets (which are present in vox_to_world_space_method_2 and 3).

Closes #743.

### DIFF
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -1258,7 +1258,7 @@ def vox_to_world_space_method_1(
     import numpy as np
 
     return np.array(coordinates_vol) * np.array(
-        header["pixdim"][1], header["pixdim"][2], header["pixdim"][3]
+        [header["pixdim"][1], header["pixdim"][2], header["pixdim"][3]]
     )
 
 


NicolasGensollen (approved):
Thanks for the fix @MatthieuJoulot !
It would be great to have some unitTests to catch these... ;)

omar-rifai:
Thanks @MatthieuJoulot. If you can please confirm that there is no regression with the data from `data_ci` by running the test manually then i'll merge.

MatthieuJoulot:
> Thanks @MatthieuJoulot. If you can please confirm that there is no regression with the data from `data_ci` by running the test manually then i'll merge.

I have not found any regression with the data from the data_ci.

omar-rifai:
sorry I merged a bit too fast and missed @NicolasGensollen's comment. If you can add unit tests for these functions it would be great.

PR #746, opened 2022-09-12
add unit test for vox_to_world_space_method_1

MatthieuJoulot:


test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
...
+
+    from clinica.iotools.utils.data_handling import vox_to_world_space_method_1
+
+    vol = 100 + 10 * np.random.randn(5, 5, 2, 100)
+    img = Nifti1Image(vol, np.eye(4))
+    nii_header = img.get_header()
+    coords = np.array([0.0, 0.0, 0.0])
+    vox_to_world_space_method_1(coordinates_vol=coords, header=nii_header)
+    assert np.array_equal(
+        vox_to_world_space_method_1(coordinates_vol=coords, header=nii_header), coords
```
omar-rifai, line 22: You ought to use the return value from the previous call otherwise you are calling the function twice for no reason.

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+from string import Template
+from typing import Union
+
+import pytest
+
+
+def test_vox_to_world_space():
+    """Test function `vox_to_world_space_method_1`."""
```
omar-rifai, line 9: Maybe add a comment to explicit what we are trying to test (i.e that the return data structure is of the right format)

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
```
NicolasGensollen, line 1: Unless I am missing something, I don't think you need any of these imports.

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
...
+    from nibabel.nifti1 import Nifti1Image
+
+    from clinica.iotools.utils.data_handling import vox_to_world_space_method_1
+
+    vol = 100 + 10 * np.random.randn(5, 5, 2, 100)
+    img = Nifti1Image(vol, np.eye(4))
+    nii_header = img.get_header()
+    coords = np.array([0.0, 0.0, 0.0])
+    vox_to_world_space_method_1(coordinates_vol=coords, header=nii_header)
+    assert np.array_equal(
```
NicolasGensollen, line 21: You can use `assert_array_equal` from `numpy` for this

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_vox_to_world_space():
+    """Test function `vox_to_world_space_method_1`."""
+
+    import numpy as np
+    from nibabel.nifti1 import Nifti1Image
+
+    from clinica.iotools.utils.data_handling import vox_to_world_space_method_1
+
+    vol = 100 + 10 * np.random.randn(5, 5, 2, 100)
```
NicolasGensollen, line 16: I would fix the random seed here, WDYT?

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
...
+    img = Nifti1Image(vol, np.eye(4))
+    nii_header = img.get_header()
```
NicolasGensollen, line 18:
```suggestion
```

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,23 @@
...
+    coords = np.array([0.0, 0.0, 0.0])
+    vox_to_world_space_method_1(coordinates_vol=coords, header=nii_header)
```
NicolasGensollen, line 20:
```suggestion
    vox_to_world_space_method_1(coordinates_vol=coords, header=img.get_header())
```

NicolasGensollen (commented):
Thanks @MatthieuJoulot ! 👍 
I made some suggestions below.

test/unittests/iotools/utils/test_data_handling.py
MatthieuJoulot: Removed the call line 20 !

test/unittests/iotools/utils/test_data_handling.py
MatthieuJoulot: I've changed it, I don't know if it is what you expected, so tell me :)

test/unittests/iotools/utils/test_data_handling.py
```diff
@@ -0,0 +1,20 @@
...
+    .. note::
+        We want to verify, that the function does run with it's normal parameters
+        and that it returns what we expect it to return."""
+
+    import numpy as np
+    from nibabel.nifti1 import Nifti1Image
+
+    from clinica.iotools.utils.data_handling import vox_to_world_space_method_1
+
+    vol = np.ndarray(shape=(5, 5, 2, 100))
```
NicolasGensollen, line 13: After looking at the code in Clinica, I don't think this data is testing our use case correctly.
If you create a `Nifti1Image` like this, its header will have an `sform_code` larger than zero and `method_3` will be used instead:

https://github.com/aramis-lab/clinica/blob/cf5775924179a6e72ca340f33caeb2faf881dd35/clinica/iotools/utils/data_handling.py#L1206-L1209

I think we should test in a way similar to this instead:

```python
from nibabel.nifti1 import Nifti1Header

head = Nifti1Header()
assert not head["qform_code"] > 0
assert head["sform_code"] == 0
assert np.all(head["pixdim"] == 1)
center_coordinates = np.array([.5] * 3)
assert_array_equal(
    vox_to_world_space_method_1(coordinates_vol=center_coordinates, header=head),
    center_coordinates
)
head["pixdim"][1] = 2
assert_array_equal(
    vox_to_world_space_method_1(coordinates_vol=center_coordinates, header=head),
    np.array([1. , 0.5, 0.5])
)
```

WDYT?

NicolasGensollen (commented):
Thanks @MatthieuJoulot ! 👍
After a closer look, I think we need better data for this test to make sense (see my comment below). I made a quick suggestion but feel free to improve and fix potential mistakes!

test/unittests/iotools/utils/test_data_handling.py
MatthieuJoulot: It looks better than what i'd done. And it works so it is fine by me ! I'll implement it that way :)

NicolasGensollen (approved):

PR #747, opened 2022-09-13
Remove dependency on nipy for tests

### LINKED ISSUE #734: Consider getting rid of nipy as a testing dependency
Clinica currently depends on `nipy` for testing, transitively through the `nipype.algorithms.metrics.Similairy`. `nipy` fails to install on Python 3.10 with Poetry 1.2, forcing us to pin the minimum version of Poetry to 1.1.x for the moment. I forwarded the issue upstream and got an immediate reply.

Poetry advised to forward the issue to `nipy` and advise them to transition their build system to something [PEP-518](https://peps.python.org/pep-0518) compliant. Unlike standard `setup.py` scripts however, nipy's make heavy use of `numpy.distutils` specific features (which is deprecated since version 1.23 and is planned for removal in Python 3.12) which are pretty complex, if not impossible to transition to plain `setuptools`. It is similar in its structure to what the build system used to look like in `scipy`, which chose to transition to `meson` instead. I don't think `nipy` will have the manpower nor the will to operate such transition any time soon.

In my opinion, the path to least resistance sounds to just vendor or implement our own similarity function and just drop the burden of depending on `nipy`. We hardly use anything in it, yet this dependency is going to become an annoyance for supporting future versions of Python.

ghisvail:
Closes #734


ghisvail:
Ready for review, requires running non-regression tests to validate, and maybe adjust, threshold values.

NicolasGensollen:
Just realized I forgot to make a review of this one. Doing it right now...

NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #752, opened 2022-09-20
[REF] Refactor BIDS and CAPS reader configurations

NicolasGensollen:
This PR proposes to refactor the BIDS (see #658) and CAPS (see #731) reader configurations. More precisely, this PR attempts to:

- propose a uniform format for BIDS and CAPS queries which used to be specified as strings and dictionaries respectively. Now, everything is specified as dictionaries.
- factorize the code since both readers now expect similar data formats

Another reason, which originally motivated this refactoring is that we will pretty soon have to provide more complex BIDS input specifications than simple strings (as it is currently the case for `t1-linear` with `T1W`). 

For example, in `PETVolume` (see #729), we need to read PET scans as inputs, but the user specifies through the CLI the tracer of interest. This means that we need to be able to dynamically pass this information to the querying engine, and not rely on static predefined queries, otherwise we will end up pretty soon with un-maintainable query registries like this:

```python
bids_default_queries = {
        "T1w": {"datatype": "anat", "suffix": "T1w", "extension": [".nii.gz"]},
        "pet": {"datatype": "pet", "suffix": "pet", "extension": [".nii.gz"]},
        "pet_fdg": {"datatype": "pet", "suffix": "pet", "tracer": "fdg", "extension": [".nii.gz"]},
        "pet_av45": {"datatype": "pet", "suffix": "pet", "tracer": "av45", "extension": [".nii.gz"]},
     ...
    }
```

### DIFF
--- a/clinica/pydra/engine.py
+++ b/clinica/pydra/engine.py
@@ -71,9 +71,9 @@ def add_input_reading_task(
     query = query_maker(pu.list_workflow_inputs(core_workflow))
     if len(query) == 0:
         return pipeline
-    input_dir = "bids_dir" if "bids" in reader.name else "caps_dir"
+    input_dir = "bids_dir" if "bids" in reader.__name__ else "caps_dir"
     input_workflow = Workflow(
-        name=f"input_workflow_{reader.name}",
+        name=f"input_workflow_{reader.__name__}",
         input_spec=["input_dir"],
     )
     try:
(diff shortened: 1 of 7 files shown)


NicolasGensollen:
52e0bd1 simplifies the input specifications (no need to specify the `_graph_checksums` field manually anymore, see https://github.com/nipype/pydra/pull/584). This requires the very latest Pydra though....

omar-rifai:
Thanks @NicolasGensollen ! The code LGTM. If no pydra release is planned we can add the dev dependency and merge.

ghisvail:
> Thanks @NicolasGensollen ! The code LGTM. If no pydra release is planned we can add the dev dependency and merge.

We could add it now, check that CI passes, and later update `pyproject.toml` to use `pydra = "^0.20"` if the release happens soon enough.

omar-rifai:
Ok fair enough, I'll merge.

PR #754, opened 2022-09-26
Use cache feature of actions/setup-python@v4

ghisvail:
Recent versions of [actions/setup-python](https://github.com/actions/setup-python/releases/tag/v3.1.0) now support caching `poetry` dependencies.

This PR makes use of this new caching feature which enables further simplification of our CI scripts and make targets and removal of our custom cache management step.


NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #756, opened 2022-09-29
Pydra implementation of PetLinear

NicolasGensollen:
This PR proposes a Pydra implementation of PETLinear.


clinica/pydra/pet_linear/pet_linear_cli.py
```diff
@@ -0,0 +1,103 @@
...
+        Whether to also return the uncropped image or not. Default=False.
+
+    save_pet_in_t1w_space : bool, optional
+        ???. Default=False.
+
+    subjects_sessions_tsv : str, optional
+        ???. Default=None.
+
+    working_directory : str, optional
+        ???. Default=None.
```
ghisvail, line 78:
```suggestion
    working_directory : str, optional
        Save intermediate results to this working directory.
```

clinica/pydra/pet_linear/pet_linear_cli.py
```diff
@@ -0,0 +1,103 @@
...
+        This will also be used for writing the outputs.
+
+    acq_label : str
+        Label given to the PET acquisition, specifying the tracer used.
+        Frequently used values are '18FFDG' or '18FAV45'.
+
+    suvr_reference_region : str
+        ???
+        The reference region must be specified to perform intensity normalization.
+        Accepted values include: 'pons', 'cerebellumPons', 'pons2', 'cerebellumPons2'.
```
ghisvail, line 66: Use set notation when parameter has a fixed set of possible values (numpydoc):

```suggestion
    suvr_reference_region : {'pons', 'cerebellumPons', 'pons2', 'cerebellumPons2'}
        Reference region used for intensity normalization.
```

clinica/pydra/pet_linear/pet_linear_cli.py
```diff
@@ -0,0 +1,103 @@
...
+        ???. Default=False.
+
+    subjects_sessions_tsv : str, optional
+        ???. Default=None.
+
+    working_directory : str, optional
+        ???. Default=None.
+
+    n_procs : int, optional
+        Number of processes to use. Default=None.
```
ghisvail, line 81: `Default=None` is redundant with the `optional` qualifier.

```suggestion
        Number of processes to use.
```

ghisvail (changes requested):
Looking good, with some suggestions for improvement that are worth a consideration.

NicolasGensollen:
Thanks for the review and great suggestions @ghisvail  ! 👍

NicolasGensollen:
I fixed some issues and the pipeline is now running with both engines. So this should be ready for another round of reviews.

MatthieuJoulot (approved):
LGTM ! Tested both the nipype and the pydra version, both work !
Only one question on the code !

clinica/pydra/pet_linear/pet_linear_cli.py
```diff
@@ -0,0 +1,100 @@
...
+        "uncropped_image": uncropped_image,
+        "save_PETinT1w": save_pet_in_t1w_space,
+    }
+    pipeline = pydra_pet_linear.build_core_workflow(
+        name="pet-linear-pydra",
+        input_dir=bids_directory,
+        output_dir=caps_directory,
+        # tsv_file=subjects_sessions_tsv,
+        # base_dir=working_directory,
+        parameters=parameters,
```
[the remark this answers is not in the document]
NicolasGensollen: You're right, I removed them. We'll add them back when we have a supporting implementation.

PR #757, opened 2022-09-29
build: Install nipype up to version 1.8.2

ghisvail:
Version 1.8.3 introduced a regression to the DWI bias correction interface.

See https://github.com/nipy/nipype/pull/3523


MatthieuJoulot (approved):
LGTM

PR #759, opened 2022-09-30
[MAINT] Upgrade joblib to `1.2.0`

NicolasGensollen:
Force using joblib >= 1.2.0 as a security issue was recently fixed.
See https://github.com/joblib/joblib/pull/1321 and https://github.com/joblib/joblib/pull/1327


ghisvail:
TIL `joblib` was not even listed in our dependencies although we directly import it. Double kudos.

ghisvail (approved):

PR #760, opened 2022-09-30
[ENH] Improve CAPS reader

NicolasGensollen:
This PR extracts from #729 the work related to the improvements of the CAPS reader which was necessary for making `PETVolume` work.

@omar-rifai let me know if this works for the pipelines you are currently working on.

I'm happy to get some reviews on this one since I believe the code could be improved quite a bit...


omar-rifai:
Great ! I'll try to test this afternoon and get back to you.

NicolasGensollen:
@omar-rifai After merging #752 and rebasing I realized there were a few small issues due to the fact that the `CAPSReader` was trying to do too many things (i.e reading both with `clinica_file_reader` and `clinica_group_reader` depending on the query).

I propose here to split the `CAPSReader` in two: the `CAPSFileReader` which only reads "standard" CAPS files, and the `CAPSGroupReader` which only reads from groups folders. I also created classes for queries instead of relying on dicts of dicts.

Still unsure this is the best implementation, but I think it makes things a bit clearer. I'm open to suggestions though !

ghisvail:
> I propose here to split the `CAPSReader` in two: the `CAPSFileReader` which only reads "standard" CAPS files, and the `CAPSGroupReader` which only reads from groups folders.

Anything development that gets us closer to the Single Responsibility Principle is a good thing 👍

NicolasGensollen:
Anyone's up for doing a review ?

ghisvail:
> Anyone's up for doing a review ?

I'm on it

clinica/pydra/interfaces.py
```diff
@@ -32,35 +36,59 @@ def __init__(self, **kwargs):
...
+                    temp = [list(i) for i in zip(*temp)]  # Transpose
+                output_query[k] = temp
             else:
-                raise ValueError(f"Unknown reader {reader}.")
-        return query
+                output_query[k] = self._execute_single_query(query)
+        return output_query
+
+    @abc.abstractmethod
+    def _execute_single_query(self, query: Dict) -> List[str]:
```
ghisvail, line 66: Just a comment, if you are not going to type the dictionary (or list) further, just use `dict` (or `list`) instead.

clinica/pydra/interfaces.py
```diff
@@ -121,7 +149,15 @@ def caps_reader(query: dict, input_dir: PathLike):
...
     """
-    caps_data_grabber = CAPSDataGrabber(output_query=query)
+    if isinstance(query, CAPSFileQuery):
+        grabber = CAPSFileDataGrabber
+    elif isinstance(query, CAPSGroupQuery):
+        grabber = CAPSGroupDataGrabber
+    else:
+        raise TypeError(
+            f"caps_reader received an unexpected query type {type(query)}. Supported types are: CAPSFileQuery and CAPSGroupQuery."
+        )
```
ghisvail, line 159: line wrap

```suggestion
        raise TypeError(
            f"caps_reader received an unexpected query type {type(query)}. "
            "Supported types are: CAPSFileQuery and CAPSGroupQuery."
        )
```

clinica/utils/input_files.py
```diff
@@ -194,12 +199,54 @@
 # T1-Volume
 
 
+def aggregator(func):
+    """If the decorated function receives iterable arguments,
+    this decorator will call the decorated function for each
+    value in the iterable and aggregate the results in a list.
+    This works only if the iterables provided have the same length.
+    Arguments lefts as non-iterable will be repeated.
```
ghisvail, line 207: I had to read this multiple times to understand what it does. I believe a few exemples would be welcome.

clinica/utils/input_files.py
```diff
@@ -194,12 +199,54 @@
...
+    """
+
+    @functools.wraps(func)
+    def wrapper_aggregator(*args, **kwargs):
+        arg_sizes = [len(arg) for arg in args if isinstance(arg, Iterable)]
+        arg_sizes += [
+            len(arg) for k, arg in kwargs.items() if isinstance(arg, Iterable)
+        ]
+        if len(np.unique(arg_sizes)) > 1:
```
ghisvail, line 216: Sets cost less than `numpy.unique`

```suggestion
        if len(set(arg_sizes)) > 1:
```

clinica/utils/input_files.py
```diff
@@ -194,12 +199,54 @@
...
+            for i in range(len(new_kwargs)):
+                if not isinstance(arg, Iterable):
+                    new_kwargs[i][k] = arg
+                else:
+                    new_kwargs[i][k] = arg[i]
+        if len(new_args) == 0:
+            return [func(**x) for x in new_kwargs]
+        elif len(new_kwargs) == 0:
+            return [func(*x) for x in zip(*new_args)]
+        return [func(*x, **y) for x, y in zip(zip(*new_args), new_kwargs)]
```
ghisvail, line 238: This is painful to read without comments to dissociate the cases covered by this decorator and their intermediate steps.

test/unittests/pydra/test_interfaces.py
```diff
@@ -0,0 +1,44 @@
...
+)
+from clinica.pydra.query import BIDSQuery, CAPSFileQuery, CAPSGroupQuery
+
+
+def test_bids_reader(tmp_path):
+    from nipype.interfaces.io import BIDSDataGrabber
+
+    from clinica.pydra.interfaces import bids_reader
+
+    task = bids_reader(BIDSQuery({}), tmp_path)
```
ghisvail, line 17: I am not sure what a `BIDSQuery({})` represents, tbh. I would expect this to be an invalid value and therefore throw at construction. Assuming the goal is to represent a default `BIDSQuery`, then the latter should rather be constructed with `BIDSQuery()` imo.

test/unittests/pydra/test_interfaces.py
```diff
@@ -0,0 +1,44 @@
...
+    assert task.name == "bids_reader_task"
+    assert task.inputs.base_dir == tmp_path
+    assert isinstance(task._interface, BIDSDataGrabber)
+
+
+@pytest.mark.parametrize(
+    "query,grabber",
+    [
+        (CAPSFileQuery({}), CAPSFileDataGrabber),
+        (CAPSGroupQuery({}), CAPSGroupDataGrabber),
```
ghisvail, line 28: Same comment, this construction with an empty dictionary reads weird. What would be the differences between `CAPSFileQuery()` and `CAPSFileQuery({})`.

ghisvail (changes requested):
I provided a few suggestions where I could. Perhaps, some more explanation in the opening post as to what the situation was, how it was changed and what the benefits are would help.

As it stands right now, I need further guidelines to see where the pieces of improvement are.

test/unittests/pydra/test_interfaces.py
```diff
@@ -0,0 +1,44 @@
...
+    assert task.inputs.base_dir == tmp_path
+    assert isinstance(task._interface, grabber)
+
+
+def test_caps_reader_error(tmp_path):
+    with pytest.raises(
+        TypeError,
+        match="caps_reader received an unexpected",
+    ):
+        caps_reader(BIDSQuery({}), tmp_path)
```
ghisvail, line 44:
```suggestion
        caps_reader(BIDSQuery({}), tmp_path)  # noqa
```

ghisvail:
I am thinking, perhaps what we are doing with BIDS, CAPS and CAPS group queries is a good use case for using [typed dictionaries](https://peps.python.org/pep-0589/).

At the end of the day, the current `*Query` classes only act as proxies to an internal `dict` without specifying its own encapsulation protocol. Since those classes are used to distinguish different *kinds* of queries, but we kind of want to keep access to the same dictionary protocol underneath, I think that's exactly what `TypedDict` + `isinstanceof` can give us with a clearer intent and limited overhead.

omar-rifai (approved):
I kind of agree with Ghislain on the constructor default which could be simplified, but otherwise looks good. Thanks.

clinica/utils/input_files.py
NicolasGensollen: Yes, I'm not proud of this function and I totally get that it is very painful to read.

I initially thought there was a more elegant/clever way to do this but couldn't find one. If you have suggestions/ideas I'll take them.

To try answering your points, I added a few things in [80b8d88](https://github.com/aramis-lab/clinica/pull/760/commits/80b8d88dfdb63936783ac0004838aaf58ebd9e00):

- Basic unit tests for this function
- Examples and basic explanation of why we need this
- A few comments to break and explain the chunks of code

Hopefully this is enough to understand what this decorator does and how to use it.

clinica/utils/input_files.py
NicolasGensollen: See my answer to your other comment below.

NicolasGensollen:
> I am thinking, perhaps what we are doing with BIDS, CAPS and CAPS group queries is a good use case for using [typed dictionaries](https://peps.python.org/pep-0589/).
> 
> At the end of the day, the current `*Query` classes only act as proxies to an internal `dict` without specifying its own encapsulation protocol. Since those classes are used to distinguish different _kinds_ of queries, but we kind of want to keep access to the same dictionary protocol underneath, I think that's exactly what `TypedDict` + `isinstanceof` can give us with a clearer intent and limited overhead.

I see your point but I don't know if we can get the same behavior with `TypedDict` as the one implemented here. I'll have a closer look.

ghisvail:
Another code smell that makes me think the design is not optimal is that the Liskov substitution principle is broken for `CAPSQuery` w.r.t. the base `Query` on the abstract method `combine_queries`. The function signatures should not mismatch.

ghisvail (commented):
Consensus is to merge this as is, and maybe improve on it later. Removing my request for changes to allow a clean merge.

ghisvail:
Ok GitHub won't remove my request for changes for some reasons, but that's ok 😅

PR #763, opened 2022-10-05
[CI] Improve management of locked dependencies

ghisvail:
- Add target to check the consistency of the locked dependencies with the project
- Ensure install is run with a consistent lock file
- Add target to refresh the lock file

⚠️  Requires Poetry version 1.2

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -15,6 +15,7 @@ pipeline {
           environment {
             CONDA_ENV = "$WORKSPACE/env"
             CONDA_HOME = "$HOME/miniconda"
+            PATH = "$HOME/.local/bin:$PATH"
           }
           stages {
             stage('Build environment') {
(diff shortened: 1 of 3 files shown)


ghisvail:
Yes, Jenkins is still stuck with Poetry v1.1 🤦

ghisvail:
Ideally, we'd want Poetry to be available separately from the project's environment like it's done with GitHub CI. The recommended way nowadays is to use `pipx` to install Python tools into their separate environment and avoid interferences. We should probably adopt the same approach for our CI.

ghisvail:
To be rebased onto #765

ghisvail:
> `poetry==1.2.1` is available in the CI agents.

I get `make: poetry: Command not found` after removal of poetry from Conda manifests.

NicolasGensollen (approved):
Great ! Thanks ! 👍

.jenkins/Jenkinsfile
```diff
@@ -6,7 +6,7 @@ pipeline {
   agent none
   stages {
     stage('Checkout') {
-      failFast true
+      //failFast true
```
NicolasGensollen, line 9: Maybe we can undo this before merging

PR #766, opened 2022-10-11
Delete node dwi

MatthieuJoulot:
The preprocessing using T1 for dwi currently has a problem: the amount of temporary data generated is way too important.
Thus, this PR aims at reducing this amount of temporary data, by suppressing it during the processing (when it is not useful anymore)

### DIFF
--- a/clinica/pipelines/cli_param/option.py
+++ b/clinica/pipelines/cli_param/option.py
@@ -190,5 +190,5 @@ delete_cache = option(
     "-dc",
     "--delete_cache",
     is_flag=True,
-    help="Use a lighter version of the pipeline",
+    help="If True,  large intermediary files will be deleted before the end of the pipeline. This option is only implemented in dwi_preprocessing_using_t1 for now.",
 )
--- a/clinica/pipelines/cli_param/option.py
+++ b/clinica/pipelines/cli_param/option.py
@@ -190,5 +190,5 @@ delete_cache = option(
     "-dc",
     "--delete_cache",
     is_flag=True,
-    help="If True,  large intermediary files will be deleted before the end of the pipeline. This option is only implemented in dwi_preprocessing_using_t1 for now.",
+    help="If True, large intermediary files will be deleted before the end of the pipeline.",  # This option is only implemented in dwi_preprocessing_using_t1 for now.
 )
--- a/clinica/pipelines/cli_param/option.py
+++ b/clinica/pipelines/cli_param/option.py
@@ -186,9 +186,9 @@ atlas_path = option(
     help="Compute atlases at the end of the path",
 )
 
-light_version = option(
-    "-li",
-    "--light_version",
+delete_cache = option(
+    "-dc",
+    "--delete_cache",
     is_flag=True,
     help="Use a lighter version of the pipeline",
 )
--- a/clinica/pipelines/cli_param/option.py
+++ b/clinica/pipelines/cli_param/option.py
@@ -185,3 +185,10 @@ atlas_path = option(
     type=click.Path(exists=True, dir_okay=True, resolve_path=True),
     help="Compute atlases at the end of the path",
 )
+
+light_version = option(
+    "-li",
+    "--light_version",
+    is_flag=True,
+    help="Use a lighter version of the pipeline",
+)
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -306,7 +306,7 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
         # Susceptibility distortion correction using T1w image
         sdc = epi_pipeline(
             self.base_dir,
-            self.parameters["light_mode"],
+            self.parameters["delete_cache"],
             name="SusceptibilityDistortionCorrection",
         )
 
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -304,7 +304,11 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
             initrand=self.parameters["initrand"],
         )
         # Susceptibility distortion correction using T1w image
-        sdc = epi_pipeline(self, name="SusceptibilityDistortionCorrection")
+        sdc = epi_pipeline(
+            self.base_dir,
+            self.parameters["light_mode"],
+            name="SusceptibilityDistortionCorrection",
+        )
 
         # Remove bias correction from (Jeurissen et al., 2014)
         bias = npe.Node(mrtrix3.DWIBiasCorrect(use_ants=True), name="RemoveBias")
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -242,6 +242,7 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
         from clinica.utils.dwi import compute_average_b0
 
         from .dwi_preprocessing_using_t1_utils import (
+            delete_apply_transform,
             init_input_node,
             prepare_reference_b0,
             print_end_pipeline,
@@ -304,7 +305,8 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
             initrand=self.parameters["initrand"],
         )
         # Susceptibility distortion correction using T1w image
-        sdc = epi_pipeline(name="SusceptibilityDistortionCorrection")
+        sdc = epi_pipeline(self, name="SusceptibilityDistortionCorrection")
+
         # Remove bias correction from (Jeurissen et al., 2014)
         bias = npe.Node(mrtrix3.DWIBiasCorrect(use_ants=True), name="RemoveBias")
         # Compute b0 mask on corrected avg b0
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -242,7 +242,6 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
         from clinica.utils.dwi import compute_average_b0
 
         from .dwi_preprocessing_using_t1_utils import (
-            delete_apply_transform,
             init_input_node,
             prepare_reference_b0,
             print_end_pipeline,
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -62,6 +62,7 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
 
         self.parameters.setdefault("use_cuda", False)
         self.parameters.setdefault("initrand", False)
+        self.parameters.setdefault("delete_cache", False)
 
     def check_custom_dependencies(self):
         """Check dependencies that can not be listed in the `info.json` file."""
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
@@ -308,3 +308,16 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
     )
 
     return out_reference_b0, out_b0_dwi_merge, out_updated_bval, out_updated_bvec
+
+
+def delete_apply_transform(marker, dir_to_del, base_dir, light):
+    import shutil
+    from pathlib import Path
+
+    print("\n", base_dir, "\n")
+    print("\n", dir_to_del, "\n")
+    if not light:
+        for z in Path(base_dir).rglob(f"*{dir_to_del}*"):
+            print(z)
+            shutil.rmtree(z)
+    return
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
@@ -310,13 +310,29 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
     return out_reference_b0, out_b0_dwi_merge, out_updated_bval, out_updated_bvec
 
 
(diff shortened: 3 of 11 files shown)


clinica/pipelines/cli_param/option.py
```diff
@@ -185,3 +185,10 @@
     type=click.Path(exists=True, dir_okay=True, resolve_path=True),
     help="Compute atlases at the end of the path",
 )
+
+light_version = option(
+    "-li",
+    "--light_version",
+    is_flag=True,
+    help="Use a lighter version of the pipeline",
```
NicolasGensollen, line 193: We can discuss this, but this feels unclear for users.
Maybe something like `--delete_cache`, or `--clean_cache`?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -82,7 +82,7 @@ def eddy_fsl_pipeline(low_bval, use_cuda, initrand, name="eddy_fsl"):
     return wf
 
 
-def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
+def epi_pipeline(self, name="susceptibility_distortion_correction_using_t1"):
```
NicolasGensollen, line 85: You should pass the arguments you actually need here, which seems to be:

- a boolean to know whether you should clean the cache or not, defaulting to "not clean the cache"
- the `base_dir` path if there is no other way to recover it

Please also update the docstring to describe these new parameters.

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,18 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
     )
 
     return out_reference_b0, out_b0_dwi_merge, out_updated_bval, out_updated_bvec
+
+
+def delete_apply_transform(marker, dir_to_del, base_dir, light):
```
NicolasGensollen, line 313: Could you add type hints and a docstring?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,18 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
...
+    import shutil
+    from pathlib import Path
+
+    print("\n", marker, "\n")
```
NicolasGensollen, line 317: Please remove the prints

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,18 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
...
+    print("\n", base_dir, "\n")
+    print("\n", dir_to_del, "\n")
+    if not light:
```
NicolasGensollen, line 320: I don't understand why you need `light`. You can simply add the cleaning node to the pipeline conditionally to the value of the boolean parameter (called `light_version` for now), or am I missing something?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,18 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
...
+        for a in dir_to_del:
+            for z in Path(base_dir).rglob(f"*{Path(Path(marker).parent).parent}*{a}*"):
```
NicolasGensollen, line 322: I think this line can be simplified

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
NicolasGensollen, line 313: Is there something specific to `apply_transform` in this function ?
It looks like it's deleting files in passed directories containing the passed marker, but it is very hard to tell what file structure is expected in line 322 when reading the code. A docstring with examples and some unit tests would be helpful I think.

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Some comments as I was doing a first pass on it.

clinica/pipelines/cli_param/option.py
MatthieuJoulot: Yes, I agree. We should discuss it ! (Not only for the naming, for the logic as well) :)

clinica/pipelines/cli_param/option.py
```diff
@@ -185,3 +185,10 @@
     type=click.Path(exists=True, dir_okay=True, resolve_path=True),
     help="Compute atlases at the end of the path",
 )
+
+delete_cache = option(
+    "-dc",
+    "--delete_cache",
+    is_flag=True,
+    help="Use a lighter version of the pipeline",
```
NicolasGensollen, line 193: Could you add a bit more information in the description?
Maybe something like: "If True, large intermediary files will be deleted before the end of the pipeline. This option is only implemented in dwi_preprocessing_using_t1 for now."

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -91,6 +95,17 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
     function (SyN algorithm). This workflow allows also a coregistration of DWIs with their respective
     baseline T1-weighted structural scans in order to latter combine tracks and cortex parcellation.
 
+    Parameters
+    ----------
+    base_dir: str
+    Working directory, whcih contains all of the intermediary data generated.
```
NicolasGensollen, line 101:
```suggestion
        Working directory, which contains all of the intermediary data generated.
```

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -91,6 +95,17 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
...
+
+    delete_cache: bool
+    If True, part of the temporary data is automatically deleted after usage.
```
NicolasGensollen, line 104:
```suggestion
        If True, part of the temporary data is automatically deleted after usage.
```

Could be worth documenting what data will be deleted by using this option. WDYT?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -91,6 +95,17 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
...
+
+    name: str
```
NicolasGensollen, line 106:
```suggestion
    name: str, optional
```

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -91,6 +95,17 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
...
+    Name of the pipeline.
```
NicolasGensollen, line 107:
```suggestion
        Name of the pipeline.
        Default="susceptibility_distortion_correction_using_t1".
```

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -225,6 +241,23 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
     )
 
     merge = pe.Node(fsl.Merge(dimension="t"), name="MergeDWIs")
+    # Delete the temporary directory that takes too much place
```
NicolasGensollen, line 244: I think it'd be clearer to add and configure this node only if `delete_cache` is True. Then, you could remove `delete_cache` from the signature of `delete_temp_dirs` and simplify the configuration of the node.

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -275,13 +308,16 @@ def epi_pipeline(name="susceptibility_distortion_correction_using_t1"):
             (jacobian, jacmult, [("jacobian_image", "in_file")]),
             (jacmult, thres, [("out_file", "in_file")]),
             (thres, merge, [("out_file", "in_files")]),
+            (merge, delete_warp_field_tmp, [("merged_file", "marker")]),
```
NicolasGensollen, line 311: You'd have to make this connection conditional on `delete_cache` then...

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,34 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
     )
 
     return out_reference_b0, out_b0_dwi_merge, out_updated_bval, out_updated_bvec
+
+
+def delete_temp_dirs(marker: str, dir_to_del: list, base_dir: str, delete_cache: bool):
+    """This function deletes the directory of the given list, if the pipeline is in "light_mode".
```
NicolasGensollen, line 314:
```suggestion
    """This function deletes the directories of the given list.
```

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
```diff
@@ -308,3 +308,34 @@ def prepare_reference_b0(in_dwi, in_bval, in_bvec, low_bval=5, working_directory
...
+
+    Parameters
+    ----------
+    marker: str
```
NicolasGensollen, line 318: Could we use a better name for this parameter? I found it hard to understand what it is from the docstring.

If I understand correctly, it is a path to some file, but we don't really care about the file itself, we just want to extract some part of it to filter the files to be deleted (subject_session ID maybe??).

I would first be clear on what you want to extract and give it an indicative name, then create a specific function

```python
def extract_whatever_you_want_to_extract(file_path: str) -> str:
    """Docstring with examples of extractions..."""
    return (Path(Path(file_path).parent).parent).name
```

and call it when connecting the delete node in the pipeline:

```python
(merge, delete_warp_field_tmp, [(("merged_file", extract_whatever), "marker")]),
```

And then simplify this function. 

Does this make sense?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
I still have some small concerns about the current implementation, especially around the `marker` parameter which hides some complexity behind a not very informative name.
I think we can make the code a bit more clear by passing to your delete function only what it needs (see comments below).

clinica/pipelines/cli_param/option.py
ghisvail: I like this description 👍

ghisvail: > "This option is only implemented in dwi_preprocessing_using_t1 for now."

With this as a comment, rather than in the description. Users don't care about this.

MatthieuJoulot: Sure !

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py
MatthieuJoulot: So it is used for two things:
 - one: to check that we are deleting the files related to the right subject-session, since the path contained in `marker`, contains the randomly generated name of the the folder of the right subject-session.
 - two: to be sure that we are past the point the files generated are useful to the pipeline and that we can delete them. In that way it is a checkpoint. 
 
For these reasons I chose to name it `marker` as it marks the location of a folder, and a "timepoint" in the pipeline.
Then again, i completely agree that this naming is far from clear to the reader.

I does make sense, will do something somewhat like you are suggesting to improve clarity !

MatthieuJoulot: @NicolasGensollen I have done some changes going your way. WDYT ?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
I think this is much easier to understand now. 👍 
I made some small suggestions to improve the documentation.
As a final step, I think we should add some unit tests for the two new functions that you have implemented (`extract_sub_ses_folder_names` for which you already have an example so should be straightforward, and `delete_temp_dirs`)

test/unittests/pipelines/dwi_preprocessing_using_t1/test_dwi_preprocessing_using_t1_utils.py
```diff
@@ -0,0 +1,59 @@
...
+            ),
+            ["node_1_name", "node_2_name"],
+            [False, False],
+        ),
+    ],
+)
+def test_delete_temp_dirs(checkpoint, dir_to_del, expected):
+
+    base_path = os.getcwd()
+    dir_1 = Path(base_path) / Path(
```
NicolasGensollen, line 46: I think you can factorize a bit. Maybe something like this:

```python
dirs = [tmp_path / Path(f"wd/.../{name}") for name in ["node_1_name", "node_2_name"]]
for dir in dirs:
    if not dir.is_dir():
        os.makedirs(dir)
....
assert [dir.is_dir() for dir in dirs] == expected
```

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
The test looks good, I just think we can factorize a bit and use `tmp_path` (see comments below). Could you add a simple test for `extract_sub_ses_folder_name` too ?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
```diff
@@ -304,7 +304,12 @@ def build_core_nodes(self):
             initrand=self.parameters["initrand"],
         )
         # Susceptibility distortion correction using T1w image
-        sdc = epi_pipeline(name="SusceptibilityDistortionCorrection")
+        sdc = epi_pipeline(
+            self.base_dir,
+            self.parameters["delete_cache"],
```
NicolasGensollen, line 309: Right now this line is breaking the instantiation tests (`KeyError` for `delete_cache`). I think we should add a default setting in `check_pipeline_parameters` (line 47), and maybe set `delete_cache` in the test.

NicolasGensollen (commented):
Thanks for addressing my points @MatthieuJoulot !
This LGTM minus the current test failure due to the new parameter introduced in this PR (see my message below).

NicolasGensollen:
@ghisvail let me know if you're planning to review this one when you're back (you're listed in the reviewers). Otherwise I'll merge it.

ghisvail:
> @ghisvail let me know if you're planning to review this one when you're back (you're listed in the reviewers). Otherwise I'll merge it.

Go for it

NicolasGensollen:
Merging then !

PR #767, opened 2022-10-12
CI: Automate updates of GitHub Actions

ghisvail:
I have done these updates manually before. Apparently, they are quite easy to automate with `dependabot`.


ghisvail:
According to the documentation, we could also automate security updates for our locked dependencies using this file.

NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #769, opened 2022-10-17
[UKB2BIDS] Drop directories labeled as `unusable`

MatthieuJoulot:
This PR drops all the unusable folder, which will therefore not be converted.


NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Do you think additional safety checks are needed to ensure the `source_filename` column does not contain unexpected things (nans for instance...) which could make your filter fail ?

MatthieuJoulot:
> Do you think additional safety checks are needed to ensure the `source_filename` column does not contain unexpected things (nans for instance...) which could make your filter fail ?
Hey @NicolasGensollen ! I don't think anything else is necessary. The only caveat I know of is the `unusable` folders.
Since the `source_filename` exists only if an image was found, there shouldn't be any nans or empty filenames.

NicolasGensollen (approved):
Fair enough, thanks for the explanations @MatthieuJoulot !

ghisvail (approved):

PR #770, opened 2022-10-18
Add out_file parameter to DWIBiasCorrect

MatthieuJoulot:
DWIPreprocessingUsingT1 crashes following changes made in nipype as detailled in this PR [#3523](https://github.com/nipy/nipype/pull/3523).
After testing, giving the `out_file` to the DWIBiasCorrect interfaces enables the pipeline not to crash.

### DIFF
--- a/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
+++ b/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_pipeline.py
@@ -306,7 +306,13 @@ class DwiPreprocessingUsingT1(cpe.Pipeline):
         # Susceptibility distortion correction using T1w image
         sdc = epi_pipeline(name="SusceptibilityDistortionCorrection")
         # Remove bias correction from (Jeurissen et al., 2014)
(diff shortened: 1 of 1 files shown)


ghisvail (approved):

ghisvail:
Shoot, I did not see you requested @NicolasGensollen input too 🤦

ghisvail:
For additional context, it looks like the regression in `nipype` still affects us despite the proposed fix. Upstream hinted this might require further investigation so we are not out of the woods yet. Best is to just set up the naming ourselves for now.

MatthieuJoulot:
> Shoot, I did not see you requested @NicolasGensollen input too 🤦

Yep, I always request both of you 😉

> LGTM. I'm just wondering where this filename comes from. 🤔
Did you chose it @MatthieuJoulot ? Do other output file names start with an upper case letter ?

This filename is the one that was generated before the regression. It is an intermediary result so i just took the name as it was since it woudn't affect the final name for the output files.

NicolasGensollen:
> @NicolasGensollen this filename is the one that was generated before the regression. It is an intermediary result so i just took the name as it was since it woudn't affect the final name for the output files.

OK, great ! 👍

Issue #771, opened 2022-10-18
Logging breaks when launched from a directory with no write privileges


danieltudosiu:
**Describe the bug**
Given an adequately configured GPU cluster where you can't run as root in a Docker container neither can you write on at the following location "/" clinica adni-to-bids breaks due to permission issues.

**To Reproduce**

Dockerfile:
```
FROM continuumio/miniconda3

ARG USER_ID
ARG GROUP_ID
ARG USER
RUN addgroup --gid $GROUP_ID $USER
RUN adduser --disabled-password --gecos '' --uid $USER_ID --gid $GROUP_ID $USER


RUN apt-get update && apt-get install -y build-essential

RUN conda install python==3.9
RUN conda install -c conda-forge dcm2niix
RUN pip install clinica==0.7.2
```
Run script
```

runai submit clinica-adni-to-bids \
  --image 10.202.67.207:5000/danieltudosiu:clinica_0_7_2 \
  --volume /nfs/project/AMIGO/ADNI_MRI/metadata/:/clinica_data_directory/ \
  --volume /nfs/project/AMIGO/ADNI_MRI/sourcedata/ADNI/:/dataset_directory/ \
  --volume /nfs/project/AMIGO/ADNI_MRI/rawdata/:/bids_directory/ \
  --large-shm \
  --host-network \
  --backoff-limit 0 \
  --gpu 0 \
  --cpu 8 \
  --project danieltudosiu \
  --command \
  -- clinica convert adni-to-bids --working_directory /bids_directory --n_procs 16 /dataset_directory /clinica_data_directory /bids_directory
```

Just run it in a Docker container as a non-root user and you should have the same issue.

**Expected behavior**
Be allowed to select the logging directory.

NicolasGensollen:
Hi @danieltudosiu 

Thanks for reporting. Could you also share the error you get? By reading your message, it is not obvious that it is related to log file writing permissions.

danieltudosiu:
@ghisvail  @NicolasGensollen  the reason that I pushed the change directly to the logger that is given the Python is because that's the logger that errors out for me by giving the following trace:
```
Traceback (most recent call last):
  File "/opt/conda/bin/clinica", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.9/site-packages/clinica/cmdline.py", line 89, in main
    cli()
  File "/opt/conda/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
[...]

PR #772, opened 2022-10-18
Added the option to chose the logging file path

danieltudosiu:
This is my solution to #771 issue.

### DIFF
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -24,7 +24,9 @@ CONTEXT_SETTINGS = dict(
 )
 
 
-def setup_logging(verbose: bool = False) -> None:
+def setup_logging(
+    verbose: bool = False, logging_file_path: str = "/pypeline.log"
+) -> None:
     """Setup Clinica's logging facilities.
 
     Args:
@@ -39,6 +41,13 @@ def setup_logging(verbose: bool = False) -> None:
     from colorlog import ColoredFormatter, StreamHandler
 
     logging_level = "DEBUG" if verbose else "INFO"
+    logging.basicConfig(
+        filename=logging_file_path,
+        filemode="a",
+        format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
+        datefmt="%H:%M:%S",
+        level=logging.DEBUG,
+    )
 
     # Root logger configuration.
     root_logger = logging.getLogger()
@@ -62,7 +71,7 @@ def setup_logging(verbose: bool = False) -> None:
         {
             "logging": {"log_directory": os.getcwd(), "log_to_file": True},
             "execution": {"check_version": False},
-        },
+        }
     )
     nipype.logging.update_logging(nipype.config)
     # Disable nipype logging to console.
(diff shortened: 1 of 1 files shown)


clinica/cmdline.py
```diff
@@ -62,7 +71,7 @@ def setup_logging(verbose: bool = False) -> None:
         {
             "logging": {"log_directory": os.getcwd(), "log_to_file": True},
             "execution": {"check_version": False},
-        },
+        }
```
NicolasGensollen, line 74: Unless I am missing something, the nipype logger will not be affected and still write to the current folder, right ?

clinica/cmdline.py
```diff
@@ -74,8 +83,15 @@ def setup_logging(verbose: bool = False) -> None:
...
 @click.version_option(version=clinica.__version__)
 @click.option("-v", "--verbose", is_flag=True, help="Increase logging verbosity.")
-def cli(verbose: bool) -> None:
-    setup_logging(verbose=verbose)
+@click.option(
+    "-lfp",
+    "--logging_file_path",
+    type=str,
+    help="The path to the logging file path.",
+    default="/pypeline.log",
```
NicolasGensollen, line 91: I am not sure I understand the default value here. Unless I am missing something, this will cause permission issues for most users running Clinica without root privileges

NicolasGensollen (commented):
Thanks for contributing @danieltudosiu !
I have a couple questions related to the proposed fix.
But more generally, I think we first need to be on the same page with the error that you're getting (see my comment in the issue).
At this point, I don't have a clear idea about this error and I don't understand how this PR is fixing your issue. I tried it on a remote machine, and the default values are causing permission errors.

clinica/cmdline.py
```diff
@@ -62,7 +71,7 @@ def setup_logging(verbose: bool = False) -> None:
         {
             "logging": {"log_directory": os.getcwd(), "log_to_file": True},
```
ghisvail, line 72: In my opinion, this should be the only line that requires a modification to support this use case: i.e. passing the `log_directory` by optional argument, still using `os.getcwd()` as default to keep compatibility with old behaviour).

You'd still need to mount an appropriate writable volume to your container to write the log files into and persist them after the container has finished running.

ghisvail (changes requested):

clinica/cmdline.py
ghisvail: We should probably not change the current behaviour, which is to write the log file in the cwd. Sure, it's not always guaranteed to be writeable (as reported in the container's case) but `/pypeline.log` definitely will not for local usage.

Other alternatives that I know of are either to write somewhere in `TMPDIR`, or in a so-called "home" directory under something like `~/.clinica` or [`$XDG_STATE_HOME`](https://specifications.freedesktop.org/basedir-spec/0.8/ar01s03.html) preferably. The former provides better availability guarantees but is less discoverable by users. That latter provides better flexibility but would require more work to manage the appropriate paths.

ghisvail: Another alternative is also to introduce an override via an environment variable, like `CLINICA_LOG_DIR`.

clinica/cmdline.py
ghisvail: I believe so. There is nothing in nipype's [config code](https://github.com/nipy/nipype/blob/master/nipype/utils/config.py) that suggests otherwise.

PR #774, opened 2022-10-21
Uninformative message

MatthieuJoulot:
Make the message regarding the incorrect subjects more informative.

### DIFF
--- a/clinica/pipelines/engine.py
+++ b/clinica/pipelines/engine.py
@@ -748,6 +748,9 @@ class Pipeline(Workflow):
                 for f in listdir(bids_dir)
                 if isdir(join(bids_dir, f)) and f.startswith("sub-")
             ]
+            # print("all sub: ", all_subs)
+            # if self.tsv_file:
+
             cross_subj = []
             long_subj = []
             for sub in all_subs:
@@ -759,6 +762,7 @@ class Pipeline(Workflow):
                     for f in listdir(join(bids_dir, sub))
                     if isdir(join(bids_dir, sub, f))
                 ]
+                # print("folder list: ", folder_list)
                 for fold in folder_list:
                     if not fold.startswith("ses-"):
                         is_cross_sectional = True
@@ -779,7 +783,7 @@ class Pipeline(Workflow):
                 for subj in cross_subj:
                     cprint(
                         msg=(
-                            f"Subject {(sub)} in your BIDS folder does not respect the longitudinal organisation "
+                            f"Subject {(subj)} in your BIDS folder does not respect the longitudinal organisation "
                             "from BIDS specification."
                         ),
                         lvl="warning",
--- a/clinica/pipelines/engine.py
+++ b/clinica/pipelines/engine.py
@@ -772,7 +772,21 @@ class Pipeline(Workflow):
                 cprint(
                     msg=(
                         f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
-                        "from BIDS specification. Clinica does not know how to handle cross sectional dataset, but it "
+                        "from BIDS specification."
+                    ),
+                    lvl="warning",
+                )
+                for subj in cross_subj:
+                    cprint(
+                        msg=(
+                            f"Subject {(sub)} in your BIDS folder does not respect the longitudinal organisation "
+                            "from BIDS specification."
+                        ),
+                        lvl="warning",
+                    )
+                cprint(
+                    msg=(
+                        f"Clinica does not know how to handle cross sectional dataset, but it "
                         "can convert it to a Clinica compliant form (using session ses-M00)"
                     ),
                     lvl="warning",


clinica/pipelines/engine.py
```diff
@@ -772,7 +776,21 @@ def copy2_add_ses(src, dst):
                 cprint(
                     msg=(
                         f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
-                        "from BIDS specification. Clinica does not know how to handle cross sectional dataset, but it "
+                        "from BIDS specification."
+                    ),
+                    lvl="warning",
+                )
+                for subj in cross_subj:
```
NicolasGensollen, line 783: This is generating a warning message for each problematic subject which might be very very verbose if there are lots of them.
It might be more readable to build a bullet list of these subjects and append it to the warning message. WDYT?

clinica/pipelines/engine.py
```diff
@@ -748,6 +748,9 @@ def copy2_add_ses(src, dst):
                 for f in listdir(bids_dir)
                 if isdir(join(bids_dir, f)) and f.startswith("sub-")
             ]
+            # print("all sub: ", all_subs)
+            # if self.tsv_file:
```
NicolasGensollen, line 752: Could you remove the dead code ?

NicolasGensollen (commented):
Thanks for taking care of that @MatthieuJoulot !

clinica/pipelines/engine.py
ghisvail: I agree that spawning dozens of warnings in case the dataset happens to have a lot of these errors would be too verbose. Besides, `print` statements are not cheap regardless of the language.

Now I wonder whether we should limit the upper bound to this list in case there is like 50 or a 100 issues in a large dataset for instance. Perhaps that's a separate concern though.

MatthieuJoulot: yup it makes sense, I have done something, tell me what you think of it :)

NicolasGensollen: I'm not sure to understand why you are not building the full warning message string, and then pass it to `cprint`. Unless I'm missing something, this is still potentially giving the user tons of warnings. 

Maybe we could do something along those lines:

```python
msg = f"{len(cross_subj)} subjects in your BIDS....The subjects concerned are:\n\t- "  # this is fine since you know there's at least one
msg += "\n\t- ".join(cross_subj)
msg += f"\nClinica does not know...."
cprint(msg, lvl="warning")
```

Then, it should be straightforward to bound the list if needed as suggested by @ghisvail

clinica/pipelines/engine.py
```diff
@@ -769,14 +770,16 @@ def copy2_add_ses(src, dst):
...
-                        f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
-                        "from BIDS specification. Clinica does not know how to handle cross sectional dataset, but it "
-                        "can convert it to a Clinica compliant form (using session ses-M00)"
-                    ),
-                    lvl="warning",
+                msg = (
+                    f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
+                    "from BIDS specification.\nThe subjects concerned are:\n\t- "
+                )
+                msg += "\n\t- ".join(cross_subj)
```
NicolasGensollen, line 777: Possible implementation of @ghisvail 's suggestion:

```python
max_subjects_displayed = 50
bound = min(len(cross_subj), max_subjects_displayed)
msg = (
    f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
    f"from BIDS specification.\nThe subjects concerned are (showing only the first {bound}):\n\t- "
)
msg += "\n\t- ".join(cross_subj[:bound])
```

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Below is a possible implementation of @ghisvail's suggestion for cropping the list (no strong opinion on the actual number, but I suppose 50 is large enough...).

clinica/pipelines/engine.py
MatthieuJoulot: Yes that looks like a good thing to do !

clinica/pipelines/engine.py
ghisvail: On principle, we should aim to decouple business logic from error handling logic. In practice, this means this sort of formatting code should belong to some domain exception class (let's say `ClinicaInconsistentDatasetError`) constructed with the relevant context (here a list of subjects) and responsible for constructing the appropriate error message.

Then the callee just needs to initialize the exception class with the collected data (the list of subjects not conforming) and the caller is responsible for catching the exception and forwarding the exception message to the appropriate logging channel. It makes also the error handling code significantly more testable, whereas the current code is not.

Clinica is filled with procedural code doing some sort of clumsy error reporting. Perhaps this PR is the right opportunity to experiment with more Pythonic error handling mechanisms, considering how localized it is.

NicolasGensollen: @ghisvail I agree this would be a good thing to do indeed.
I don't think this is too much work, so we should probably do it in this PR.

@MatthieuJoulot I think we need to create a `ClinicaInconsistentDatasetError` class in `clinica/utils/exceptions.py` and simply move all this formatting logic inside such that the engine would look like:

```python
if len(cross_subj) > 0:
    raise ClinicaInconsistentDatasetError(cross_subs)
```

MatthieuJoulot: @NicolasGensollen @ghisvail this may be a naive comment, but won't raising an error kill the process ? which we wouldn't want to because the user may want to create an alternative dataset.

ghisvail: > but won't raising an error kill the process ?

If you don't catch it, indeed.

That's why you'd need to surround the call to the function with a `try / except` block catching `ClinicaInconsistentDatasetError` and forwarding the error message to `cprint` with warning level.

MatthieuJoulot: Okay ! Works in part, but since `check_not_cross_sectionnal` is a very long function with lots of stuff inside (bit of mess), currently it would still break stuff. I'm going to look for way to isolate the specific part we are interested in.

MatthieuJoulot: Okay @ghisvail @NicolasGensollen I created a class and tried to handle the `try / except` in a smart way. WDYT ?

clinica/pipelines/engine.py
```diff
@@ -607,8 +608,31 @@ def check_not_cross_sectional(self):
         from os import listdir
         from os.path import abspath, basename, dirname, isdir, join
 
+        from clinica.utils.exceptions import ClinicaInconsistentDatasetError
         from clinica.utils.stream import cprint
 
+        def detect_cross(all_subs, bids_dir, cross_subj, long_subj):
```
NicolasGensollen, line 614: I would extract this function out of the `Pipeline` class, add type hints and a proper docstring. 
We can discuss this, but I think this function shouldn't raise/print anything, but return the list of cross sectional and longitudinal subjects.
Maybe the name can be changed to `detect_cross_sectional_and_longitudinal_subjects` or something like that. 
I would also remove the last two input parameters to avoid unnecessary mutations.

clinica/pipelines/engine.py
```diff
@@ -607,8 +608,31 @@ def check_not_cross_sectional(self):
...
+                    for f in listdir(join(bids_dir, sub))
+                    if isdir(join(bids_dir, sub, f))
+                ]
+                for fold in folder_list:
+                    if not fold.startswith("ses-"):
+                        is_cross_sectional = True
+                if is_cross_sectional:
+                    cross_subj.append(sub)
+                else:
+                    long_subj.append(sub)
```
NicolasGensollen, line 630: We can simplify this a little bit:

```python
for sub in all_subs:
    folder_list = ...
    if any([fold.startswith("ses-") for fold in folder_list]):
        cross_subj.append(sub)
    else:
        long_subj.append(sub)
```

clinica/pipelines/engine.py
```diff
@@ -607,8 +608,31 @@ def check_not_cross_sectional(self):
...
+
+            # The following code is run if cross sectional subjects have been found
+            if len(cross_subj) > 0:
```
NicolasGensollen, line 633: Don't raise but return `cross_subj` and `long_subj`

clinica/pipelines/engine.py
```diff
@@ -748,35 +772,12 @@ def copy2_add_ses(src, dst):
                 for f in listdir(bids_dir)
                 if isdir(join(bids_dir, f)) and f.startswith("sub-")
             ]
+
```
NicolasGensollen, line 775: Remove the two empty list creations below

clinica/pipelines/engine.py
```diff
@@ -748,35 +772,12 @@ def copy2_add_ses(src, dst):
...
-                    msg=(
-                        f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
-                        "from BIDS specification. Clinica does not know how to handle cross sectional dataset, but it "
-                        "can convert it to a Clinica compliant form (using session ses-M00)"
-                    ),
-                    lvl="warning",
-                )
+            try:
+                detect_cross(all_subs, bids_dir, cross_subj, long_subj)
+            except ClinicaInconsistentDatasetError:
```
NicolasGensollen, line 780: Replace with:

```python
cross_subj, long_subj = whatever_name_you_chose(all_subs, bids_dir)
try:
    _check_cross_subj(cross_subj)
except ClinicaInconsistentDatasetError as e:
    cprint(e, lvl="warning")
```

The check function would look like:

```python
def _check_cross_long_subj(cross_subj: list) -> None:
    if len(cross_subj) > 0:
        raise ClinicaInconsistentDatasetError(cross_subj)
```

and unindent the rest

clinica/utils/exceptions.py
```diff
@@ -19,3 +19,23 @@ class ClinicaCAPSError(ClinicaException):
...
+        msg = (
+            f"{len(cross_subj)} subjects in your BIDS folder did not respect the longitudinal organisation "
+            f"from BIDS specification.\nThe subjects concerned are (showing only the first {bound}):\n\t- "
+        )
+        msg += "\n\t- ".join(cross_subj[:bound])
+        msg += (
+            f"\nClinica does not know how to handle cross sectional dataset, but it "
+            "can convert it to a Clinica compliant form (using session ses-M00)"
+        )
+        cprint(msg, lvl="warning")
```
NicolasGensollen, line 41: If you print the error as a warning here, it makes the error class behave in a very unexpected way. You simply need to format the message and call the parent constructor here.

clinica/utils/exceptions.py
```diff
@@ -19,3 +19,23 @@ class ClinicaCAPSError(ClinicaException):
 
 class ClinicaParserError(ClinicaException):
     """Base class for parser errors."""
+
+
+class ClinicaInconsistentDatasetError(ClinicaException):
+    """Base class for inconsistent datasets errors."""
+
+    def __init__(self, cross_subj):
```
NicolasGensollen, line 27:
```suggestion
    def __init__(self, cross_subj: list):
```

NicolasGensollen (commented):
Thanks for looking into this @MatthieuJoulot !
I think your suggestion of refactoring the `check_not_cross_sectional` method is good as it is very long and difficult to read.
I made some suggestions below (I haven't tested the code suggestions I made, so you might have to adapt/correct...).

clinica/pipelines/engine.py
MatthieuJoulot: So, this actually doesn't work, because of it is reversed, it is actually the other way around.
Would something like this work ? It seems to work on my end.
```python
for sub in all_subs:
    folder_list = ...
    if not all([fold.startswith("ses-") for fold in folder_list]):
        cross_subj.append(sub)
    else:
        long_subj.append(sub)
```

MatthieuJoulot: So, this actually doesn't work, because of it is reversed, it is actually the other way around.
Would something like this work ? It seems to work on my end.
```python
for sub in all_subs:
    folder_list = ...
    if not all([fold.startswith("ses-") for fold in folder_list]):
        cross_subj.append(sub)
    else:
        long_subj.append(sub)
```

NicolasGensollen: Yes ! 👍 Sorry for the wrong suggestion, I wrote this a bit too quickly...

clinica/utils/filemanip.py
```diff
@@ -207,3 +207,40 @@ def extract_metadata_from_json(json_file, list_keys):
         file.close()
 
     return list_values
+
+
+def detect_cross_sectional_and_longitudinal_subjects(
+    all_subs: list, bids_dir: str
+) -> list:
```
NicolasGensollen, line 214:
```suggestion
) -> Tuple[list, list]:
```

clinica/utils/filemanip.py
```diff
@@ -207,3 +207,40 @@ def extract_metadata_from_json(json_file, list_keys):
...
+    """This function detects whether a subject as a longitudinal or cross sectionnal structure.
```
NicolasGensollen, line 215:
```suggestion
    """This function detects whether a subject has a longitudinal or cross sectional structure.
```

clinica/utils/filemanip.py
```diff
@@ -207,3 +207,40 @@ def extract_metadata_from_json(json_file, list_keys):
...
+    If it has a mixed structure, it will be considered cross sectionnal.
+
+    Parameters
+    ----------
+    all_subs: list
+        List containing all the names of the subjects.
+
+    bids_dir: str
+        Path to the bids directory.
```
NicolasGensollen, line 224:
```suggestion
        Path to the bids directory.

```

clinica/utils/filemanip.py
```diff
@@ -207,3 +207,40 @@ def extract_metadata_from_json(json_file, list_keys):
...
+    Returns
+    -------
+    cross_subj: list
+        List of all the subject detected with a cross_sectionnal organisation.
```
NicolasGensollen, line 228:
```suggestion
        List of all the subject detected with a cross_sectionnal organisation.

```

clinica/utils/filemanip.py
```diff
@@ -207,3 +207,40 @@ def extract_metadata_from_json(json_file, list_keys):
...
+    long_subj: list
+        List of all the subject detected with a longitudinal organisation.
+    """
```
NicolasGensollen, line 231: I think it'd be great to add an Example section and some unit tests for this function.

NicolasGensollen (commented):
Thanks @MatthieuJoulot !

clinica/utils/filemanip.py
MatthieuJoulot: It's the same thing on my side. Do i need to change something ?

clinica/utils/filemanip.py
MatthieuJoulot: I assumed you wanted me to correct the typo on sectionnal to sectional

clinica/utils/filemanip.py
NicolasGensollen: I added an empty line to separate the sections.

clinica/utils/filemanip.py
NicolasGensollen: I just added an empty line to separate the sections, but you're right about the typo that I missed 👍

NicolasGensollen (approved):

ghisvail:
> @ghisvail let me know if you want to take another look at this, otherwise I'll merge.

I am taking a look 👍

ghisvail (approved):

PR #777, opened 2022-11-07
Allow setting a custom logging directory via an envvar

### LINKED ISSUE #771: Logging breaks when launched from a directory with no write privileges

ghisvail:
Defaults to current working directory, unless CLINICA_LOGGING_DIR is set. This is a stop gap solution for #771 until a proper config structure for customising Clinica is designed.

Closes #771 
Closes #772

### DIFF
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -60,7 +60,10 @@ def setup_logging(verbose: bool = False) -> None:
     nipype.config.enable_debug_mode()
     nipype.config.update_config(
         {
-            "logging": {"log_directory": os.getcwd(), "log_to_file": True},
+            "logging": {
+                "log_directory": os.getenv("CLINICA_LOGGING_DIR", os.getcwd()),
+                "log_to_file": True,
+            },
             "execution": {"check_version": False},
         },
     )
--- a/docs/InteractingWithClinica.md
+++ b/docs/InteractingWithClinica.md
@@ -70,10 +70,8 @@ clinica run modality-pipeline bids_directory caps_directory -tsv my_participants
 "modality" is a prefix that corresponds to the data modality (e.g. T1, DWI, fMRI, PET) or to the category of processing (machine learning, statistics...).
 If you execute `clinica run --help`, you can see the list of `modality-pipeline` available: they correspond to the different pipelines displayed on the [main page of the documentation](..).
 
-<!-- ### clinica visualize
-
 !!! note
-    We are currently rewriting this section. We will update this section ASAP. -->
+    Clinica run logs are written in the current working directory by default. A different directory may be specified by setting the `CLINICA_LOGGING_DIR` environment variable.
 
 ### `clinica convert`
 


NicolasGensollen (commented):
LGTM, I wondering whether we could add some documentation such that users get to learn about this configuration option. I'm not sure what is the best spot for this though...
Maybe here: https://aramislab.paris.inria.fr/clinica/docs/public/latest/Third-party/#environment-variables-setup
or here: https://aramislab.paris.inria.fr/clinica/docs/public/latest/Installation/

ghisvail:
I looked at both locations you suggested but none of them sound appropriate to me. It's basically a runtime modifier, so it does not concern installation, nor third-party integration. It's not even something I would advise *any* users to play with tbh, hence my reluctance to add it to expose it publicly to the docs.

NicolasGensollen:
>It's not even something I would advise any users to play with tbh, hence my reluctance to add it to expose it publicly to the docs.

I'm not sure I'm following you here. Unless I missed something, the environment variable `CLINICA_LOGGING_DIR` is never set by Clinica, so it is the responsibility of the user to set it manually in order to change the default location for writing nipype's logs.

Currently, the only way to learn about the existence of such feature is by looking at the code itself. I strongly doubt many users will get to know about it tbh. I'm not even sure developers will remember this in a couple years...

Here is an example of Scikit-learn's documentation related to their environment variables: https://scikit-learn.org/stable/computing/parallelism.html#configuration-switches

If playing with this configuration option is dangerous for some reasons, we should explain these reasons properly in the docs, or consider not allowing the config option at all. I don't think hiding it in the code is a good solution.

>I looked at both locations you suggested but none of them sound appropriate to me. It's basically a runtime modifier, so it does not concern installation, nor third-party integration. 

Fair enough, but I still think we should find a location for these kind of advanced configurations, especially if we rely more on them in the future. We can always create a new section if needed.

NicolasGensollen (approved):
Thanks for addressing the doc request @ghisvail 
Location looks good to me, at least for now.

PR #778, opened 2022-11-07
[ENH] Make `clinica run` subcommands registration dynamic

NicolasGensollen:
**General description**

This PR proposes to change the way subcommands are registered for `clinica run` and is a first step towards having a `clinica_pipeline` decorator which would require, among other things, a dynamic registration of subcommands.

**The current way**

To add a pipeline command, you need to manually edit the file `clinica/pipelines/cli.py` to import and add the the new subcommand there.

This isn't very dynamic and can be easily forgotten when implementing a new subcommand.
For example, it seems like `clinica run compute-atlas` isn't registered although it is defined here:

https://github.com/aramis-lab/clinica/blob/3386bff1bd2898bec569ff79ab7fb3f0c19800dc/clinica/pipelines/t1_freesurfer_atlas/t1_freesurfer_atlas_cli.py#L7-L19

```
$ clinica run compute-atlas
Usage: clinica run [OPTIONS] COMMAND [ARGS]...
Try 'clinica run -h' for help.

Error: No such command 'compute-atlas'.
```

**The proposed way**

This PR defines folders to look for subcommands: `clinica/pipelines` and `clinica/pydra` for now, and looks for python files defining subcommand (ending with `_cli.py` and defining `pipeline_name`).

This removes the need for manually editing `clinica/pipelines/cli.py`. To add a new subcommand, you just need to create a new folder and define the python CLI file inside.

### DIFF
--- a/clinica/pipelines/cli.py
+++ b/clinica/pipelines/cli.py
@@ -1,87 +1,46 @@
+import os
+from pathlib import Path
+
 import click
 
-from clinica.pydra.t1_linear import t1_linear_cli as pydra_t1_linear_cli
-from clinica.pydra.t1_volume.create_dartel import cli as pydra_t1vol_cd_cli
-from clinica.pydra.t1_volume.tissue_segmentation import cli as pydra_t1vol_ts_cli
-
-from .deeplearning_prepare_data import (
-    deeplearning_prepare_data_cli as deeplearning_prepare_data_cli,
-)
-from .dwi_connectome import dwi_connectome_cli
-from .dwi_dti import dwi_dti_cli
-from .dwi_preprocessing_using_fmap import dwi_preprocessing_using_phasediff_fmap_cli
-from .dwi_preprocessing_using_t1 import dwi_preprocessing_using_t1_cli
-from .machine_learning import classification_cli
-from .machine_learning_spatial_svm import spatial_svm_cli
-from .pet_linear import pet_linear_cli
-from .pet_surface import pet_surface_cli, pet_surface_longitudinal_cli
-from .pet_volume import pet_volume_cli
-from .statistics_surface import statistics_surface_cli
-from .statistics_volume import statistics_volume_cli
-from .statistics_volume_correction import statistics_volume_correction_cli
-from .t1_freesurfer import t1_freesurfer_cli
-from .t1_freesurfer_longitudinal import (
-    t1_freesurfer_longitudinal_cli,
-    t1_freesurfer_longitudinal_correction_cli,
-    t1_freesurfer_template_cli,
-)
-from .t1_linear import flair_linear_cli, t1_linear_cli
-from .t1_volume import t1_volume_cli
-from .t1_volume_create_dartel import t1_volume_create_dartel_cli
-from .t1_volume_dartel2mni import t1_volume_dartel2mni_cli
-from .t1_volume_existing_template import t1_volume_existing_template_cli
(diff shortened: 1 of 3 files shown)


NicolasGensollen:
@ghisvail Thanks for sharing your thoughts on this.

I agree with your three points. I feel like points 1 and 2, related to the deterministic order of the command registration, could be solved by slightly modifying this PR's implementation, but I'd need to investigate further.

For the third point, I relied on click's documentation, slightly modifying this example, which is using `eval()`: https://click.palletsprojects.com/en/8.1.x/commands/#custom-multi-commands

I think we could remove the need for `eval()` but I still haven't looked into it.

> I believe it could be worth brainstorming on possible alternatives? I can look it up.

Absolutely ! By opening this PR, my goal was to propose a first possible working implementation and see what people think in terms of possible directions for improvement.

I can try to address the three points above within this design while you think of possible alternatives.

ghisvail:
I got a bit of thinking about it and may have a simpler alternative to offer.

As it stands today, command registration happens in a central location (`clinica/pipelines/cli.py`). This file is responsible for defining what's gonna become the `run` subgroup *and* to aggregate all pipeline subcommands. It's this mixture of responsibility that makes dynamic registration challenging.

Now, we could split this responsibility by removing the command aggregation part, and have each pipeline CLI module do:

```python
from clinica.pipelines.cli import cli as run_cli

def cli(...):
    [...]

run_cli.add_command(cli)
```

Then, what if this pattern of importing and adding a command were hidden away in a decorator, say the `@clinica_pipeline` you mentioned in your opening post. Then, you'd have achieved dynamic registration without resorting to `eval`, I believe.

NicolasGensollen:
I could be wrong but I think you would then have a "command not found issue" when trying to run the CLI.

I'm thinking this way because the CLI entry point is defined here:

https://github.com/aramis-lab/clinica/blob/e9c3d21397027521f4b71b19b27edfdd3915b5cc/pyproject.toml#L82-L83

and points to `clinica/cmdline.py`, which defines the main commands (`run`, `convert`, etc...), imports `clinica/pipelines/cli.py` :

https://github.com/aramis-lab/clinica/blob/e9c3d21397027521f4b71b19b27edfdd3915b5cc/clinica/cmdline.py#L15

and adds the subcommands:

https://github.com/aramis-lab/clinica/blob/e9c3d21397027521f4b71b19b27edfdd3915b5cc/clinica/cmdline.py#L87

If you remove subcommand registration from `Clinica/pipelines/cli.py`, that is, these lines:

https://github.com/aramis-lab/clinica/blob/e9c3d21397027521f4b71b19b27edfdd3915b5cc/clinica/pipelines/cli.py#L52-L58

in order to have them scattered across pipeline scripts, then I believe these subcommands will not be available.

Does this make sense or I am missing something ?

ghisvail:
A commit is worth a thousand words in this case I think. Here is a wip branch showcasing how dispatching CLI registration works:

https://github.com/ghisvail/clinica/commit/2f03587d699ac2cb678eaa44a1abe6726c1a0a59

ghisvail:
The logic here is for the `run` command, but is applicable to any other command.

NicolasGensollen:
Thanks for the example commit @ghisvail, it makes things much clearer (to me at least).
I played a little bit with the suggested approach and I think it's an interesting alternative to what's currently proposed in this PR. 
I'll think a little bit more about it and we can discuss this sometimes this week.

PR #780, opened 2022-11-07
Center all modalities if no modality is specified

### LINKED ISSUE #779: center-nifti modality option isn't optional
**Describe the bug**
When lauching the commands center-nifti, if the modality is not indicated, the command doesn't  center the niftis and just goes onto the second part of the algorithm, which consists in copying the rest of the data to the centered-bids

**To Reproduce**
Launch center-nifti without -m/--modality option

**Expected behavior**
If the option is not specified, every nifti should be centered

ghisvail:
- Fixes centering bypassed entirely if no modality specified (#779)
- Add typing annotations to `center_all_nifti`
- Rewrite docstring in numpydoc style

Closes #779

### DIFF
--- a/clinica/iotools/utils/data_handling.py
+++ b/clinica/iotools/utils/data_handling.py
@@ -1,6 +1,7 @@
 """Data handling scripts."""
 
 from os import PathLike
+from typing import Iterable, List
 
 import click
 from nibabel.nifti1 import Nifti1Header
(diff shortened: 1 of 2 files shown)


MatthieuJoulot (commented):
LGTM :) I also tested it and it behaves as we expect it to.
Is there any way to get some unit test as well ?

ghisvail:
>  Is there any way to get some unit test as well?

Good point, the regression tests needed an update. It's done.

PR #781, opened 2022-11-07
[TEST, DOC] Add tests and docstrings to functions from `filemanip.py`

NicolasGensollen:
This PR proposes to improve the `clinica.utils.filemanip` module by:

- Adding type hints when possible
- Adding docstrings following the Numpydoc's conventions
- Refactoring some code
- Adding unit tests for all functions

Also, the function `unzip_nii` was implemented in a very different way from `zip_nii` and was offering a different API. This PR also proposes to change this and have the same implementation and API for both functions.

### DIFF
--- a/clinica/utils/filemanip.py
+++ b/clinica/utils/filemanip.py
@@ -1,7 +1,7 @@
 import os
 import re
 from pathlib import Path, PurePath
-from typing import List, Tuple, Union
+from typing import Callable, List, Optional, Tuple, Union
 
 import pandas as pd
 from nipype.utils.filemanip import split_filename
@@ -9,9 +9,50 @@ from nipype.utils.filemanip import split_filename
 from clinica.utils.exceptions import ClinicaException
 
 
+def _zip_unzip_nii(
+    in_file: Union[PurePath, List[PurePath]],
+    func: Callable,
+    same_dir: bool = False,
+) -> Union[PurePath, List[PurePath]]:
+    import gzip
+    import operator
+    import shutil
+    from os import getcwd
+    from os.path import abspath, join
+
+    from traits.trait_base import _Undefined
+
+    zipping = func.__name__ == "zip_nii"
+
+    if (in_file is None) or isinstance(in_file, _Undefined):
+        return None
+
+    if isinstance(in_file, list):
+        return [func(f, same_dir) for f in in_file]
+
+    op = operator.eq if zipping else operator.ne
+    if op(in_file.suffix, ".gz"):
+        return in_file
+
+    if not in_file.exists():
+        raise FileNotFoundError(f"File {in_file} does not exist.")
+
+    orig_dir, base, ext = split_filename(str(in_file))
+    new_ext = ext + ".gz" if zipping else ext[:-3]
+    out_file = abspath(join(orig_dir if same_dir else getcwd(), base + new_ext))
+
+    outer = open if zipping else gzip.open
+    inner = gzip.open if zipping else open
+    with outer(in_file, "rb") as f_in:
+        with inner(out_file, "wb") as f_out:
+            shutil.copyfileobj(f_in, f_out)
+
+    return Path(out_file)
+
+
 def zip_nii(
     in_file: Union[PurePath, List[PurePath]], same_dir: bool = False
-) -> PurePath:
+) -> Union[PurePath, List[PurePath]]:
     """Compress the provided file(s).
 
     Parameters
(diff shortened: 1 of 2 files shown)


clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    import operator
     import shutil
     from os import getcwd
     from os.path import abspath, join
+    from pathlib import Path
 
     from nipype.utils.filemanip import split_filename
     from traits.trait_base import _Undefined
 
+    zipping = func.__name__ == "zip_nii"
```
ghisvail, line 15: Could be simplified by just passing a `compress: bool` parameter instead no? Passing an entire function just to match on its name sounds like an odd pattern to me.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    is returned.
+
+    Raises
+    ------
+    FileNotFoundError
+        If the provided file does not exist.
+    """
+    from clinica.utils.filemanip import _zip_unzip_nii  # noqa
+
+    return _zip_unzip_nii(in_file, zip_nii, same_dir)
```
ghisvail, line 84: I would be in favour to pass `compress=True` as parameter instead of the parent function.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    to the provided file is returned.
+
+    Raises
+    ------
+    FileNotFoundError
+        If the provided file does not exist.
+    """
+    from clinica.utils.filemanip import _zip_unzip_nii  # noqa
+
+    return _zip_unzip_nii(in_file, unzip_nii, same_dir)
```
ghisvail, line 129: Same comment with `compress=False`.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
     from clinica.utils.stream import cprint
 
-    assert len(participant_ids) == len(session_ids)
+    if len(participant_ids) != len(session_ids):
+        raise ValueError(
+            "The number of participant IDs is not equal to the number of session IDs."
+        )
 
+    out_folder = Path(out_folder)
     os.makedirs(out_folder, exist_ok=True)
```
ghisvail, line 173: Using [Path.mkdir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)

```suggestion
    out_folder.mkdir(parents=True, exist_ok=True)
```

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
     except Exception as e:
         cprint(msg=f"Impossible to save {out_file} with pandas", lvl="error")
         raise e
 
 
+def _raise_non_bids_or_caps_compliant_filename(filename: str) -> None:
+    raise ValueError(
+        f"Input filename {filename} is not in a BIDS or CAPS compliant format."
+        " It does not contain the subject and session information."
+    )
```
ghisvail, line 193: I looked at usages of this function and found it used twice, in L232 and L304. In both, the pattern is the same, i.e. a regex match on the path raising an exception if nothing found. I believe it's the whole thing that should be factored out, not just the raising an exception part.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    >>> get_filename_no_ext("foo.nii.gz")
+    'foo'
+    >>> get_filename_no_ext("sub-01/ses-M000/sub-01_ses-M000.tar.gz")
+    'sub-01_ses-M000'
+    """
     from nipype.utils.filemanip import split_filename
 
     _, filename_no_ext, _ = split_filename(filename)
 
     return filename_no_ext
```
ghisvail, line 262: We can get rid of `nipype` for this now that we have got `pathlib`.

```suggestion
    from pathlib import PurePath
    
    stem = PurePath(filename).stem
    while('.' in stem):
        stem = PurePath(stem).stem

    return stem
```

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
 
     m = re.search(r"(sub-[a-zA-Z0-9]+)/(ses-[a-zA-Z0-9]+)", bids_or_caps_file)
-
     if not m:
-        raise ValueError(
-            f"Input filename {bids_or_caps_file} is not in a BIDS or CAPS compliant format."
-            " It does not contain the subject and session information."
-        )
-
+        _raise_non_bids_or_caps_compliant_filename(bids_or_caps_file)
```
ghisvail, line 232: duplicated

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
 
-    id_bids_or_caps_files = [
-        re.search(r"(sub-[a-zA-Z0-9]+)_(ses-[a-zA-Z0-9]+)", file).group()
-        for file in bids_or_caps_files
-    ]
+    id_bids_or_caps_files = []
+    for f in bids_or_caps_files:
+        m = re.search(r"(sub-[a-zA-Z0-9]+)_(ses-[a-zA-Z0-9]+)", f)
+        if not m:
+            _raise_non_bids_or_caps_compliant_filename(f)
```
ghisvail, line 304: duplicated

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    with open(filename, "r") as log_file:
+        for line in log_file:
+            if re.match("(.*)crashfile:(.*)", line):
+                crash_files.append(line.replace("\t crashfile:", "").replace("\n", ""))
 
     return crash_files
 
 
-def read_participant_tsv(tsv_file):
+def read_participant_tsv(tsv_file: str) -> Tuple[List[str], List[str]]:
```
ghisvail, line 377: Just a side note, we could ensure both lists are of the same length by returning a `Iterable[Tuple[str, str]]` instead of a tuple of iterables.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    from pathlib import Path
 
     import pandas as pd
 
     from clinica.utils.exceptions import ClinicaException
 
-    if not os.path.isfile(tsv_file):
+    tsv_file = Path(tsv_file)
+    if not tsv_file.is_file():
         raise ClinicaException(
```
ghisvail, line 417: It is more Pythonic to do:

```python
try:
    df = pd.read_csv(...)
except ... :
     ...
```

than testing the file.

Actually, the file could exist without permissions to read it.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
-    try:
-        with open(json_file, "r") as file:
-            data = json.load(file)
-            for key in list_keys:
-                list_values.append(data[key])
-    except EnvironmentError:
-        raise EnvironmentError(
-            f"[Error] Clinica could not open the following JSON file: {json_file}"
+    json_file = Path(json_file)
+    if not json_file.exists():
```
ghisvail, line 458: Same comment about Pythonic file reads.

ghisvail (changes requested):
Looking good, with some requests for improvement and suggestions.

Another thing to remember is that these `zip_nii` and `unzip_nii` utilities should have been replaced by proper calls to nipype's GZip / Gunzip interfaces. We'd have to get rid of them when porting our code to `pydra`. I think they were mostly useful for SPM-based pipelines.

clinica/utils/filemanip.py
NicolasGensollen: Indeed, this is much better! Don't know why I did that... 🤔 
Implemented in [db32e53](https://github.com/aramis-lab/clinica/pull/781/commits/db32e53f4bd42edeae52719c3f8151cbe7b223cd)

clinica/utils/filemanip.py
NicolasGensollen: Actually, the pattern is slightly different, but I agree it is still worth a factorization at the cost of a small additional argument to this function. I also changed the name since it needs to return the match to the caller and not only raise.
Implemented in [7cb8c5a](https://github.com/aramis-lab/clinica/pull/781/commits/7cb8c5a4a578a0ba7ea0376176e8c151a7730198)

clinica/utils/filemanip.py
NicolasGensollen: I agree => implemented in [e263a47](https://github.com/aramis-lab/clinica/pull/781/commits/e263a475637ffb76b31a835d72afe2c1977b5b2b)

clinica/utils/filemanip.py
NicolasGensollen: implemented in b6b315a

clinica/utils/filemanip.py
NicolasGensollen: Do you suggest we do this modification in this PR ?
I think it is a good idea but I expect a few changes on code relying on this function. I'm OK for doing it here or in a follow-up PR, as you prefer.

clinica/utils/filemanip.py
```diff
@@ -1,209 +1,476 @@
...
+    If the provided file is already zipped (that is, its extension
+    is '.gz'), then nothing is done and the path to the provided file
+    is returned.
+
+    Raises
+    ------
+    FileNotFoundError
+        If the provided file does not exist.
+    """
+    from clinica.utils.filemanip import _zip_unzip_nii  # noqa
```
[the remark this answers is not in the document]
NicolasGensollen: If you don't import `_zip_unzip_nii`, Nipype will complain that the function isn't self contained when using it in a Pipeline.

clinica/utils/filemanip.py
ghisvail: Yes, we would need to trace where `read_participant_tsv` is used and where its output are tested. I agree that this should be the scope of another PR.

ghisvail (approved):

PR #782, opened 2022-11-08
[MAINT] Upgrade to Nibabel 3.2.2 and Nilearn 0.9.2

NicolasGensollen:


pyproject.toml
```diff
@@ -28,15 +28,15 @@ classifiers = [
...
+nibabel = "^3.2.2"
 nipype = ">=1.7.1,<1.8.3"
 argcomplete = "^1.9.4"
 pandas = "^1.4"
 jinja2 = "^3"
 xvfbwrapper = "*"
 numpy = "^1.17"
 scikit-learn = "^1.0"
-nilearn = {version = "^0.7.0", extras = ["plotting"]}
+nilearn = {version = "^0.9.0", extras = ["plotting"]}
```
ghisvail, line 39: The `plotting` extras have been officially [introduced](https://github.com/nilearn/nilearn/commit/ae1fe0ab68d2e88a5957ba7ab610cb742f6aebbc) for version 0.9.1.

```suggestion
nilearn = {version = "^0.9.1", extras = ["plotting"]}
```

pyproject.toml
```diff
@@ -28,15 +28,15 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
-nibabel = "^2.3.3"
+nibabel = "^3.2.2"
```
ghisvail, line 31: Could we aim for the latest version (`4.0.2`)? Or is there a limiting factor somewhere?

ghisvail (approved):

pyproject.toml
NicolasGensollen: Looks like some of the Nipype's interfaces we depend on need `trackviz` which has been removed in Nibble `4.0.0`...

NicolasGensollen: See https://github.com/nipy/nipype/issues/3486

ghisvail: Are we actually using the interface Chris mentions in the issue? From the conversation, it looks like it's a very niche use case and the problematic import has been encapsulated deeper since version `1.8.3`.

PR #787, opened 2022-11-15
[ENH] Make `clinica run` subcommands registration dynamic (second implementation)

NicolasGensollen:
This PR proposes a possible alternative implementation of #778 as suggested by @ghisvail in #778.

Closes #778

### DIFF
--- a/clinica/__init__.py
+++ b/clinica/__init__.py
@@ -3,3 +3,5 @@ from importlib.metadata import version
 __all__ = ["__version__"]
 
 __version__ = version("clinica")
+
+from . import pipelines, pydra
--- a/clinica/__init__.py
+++ b/clinica/__init__.py
@@ -3,5 +3,3 @@ from importlib.metadata import version
 __all__ = ["__version__"]
 
 __version__ = version("clinica")
-
-from . import pipelines, pydra
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -12,7 +12,6 @@ import clinica
 from clinica.engine.template import cli as generate_cli
 from clinica.iotools.converters.cli import cli as convert_cli
 from clinica.iotools.utils.cli import cli as iotools_cli
-from clinica.pipelines.cli import cli as run_cli
 from clinica.utils.exceptions import ClinicaException
 from clinica.utils.stream import cprint
 
@@ -84,7 +83,6 @@ def cli(verbose: bool) -> None:
 cli.add_command(convert_cli)
 cli.add_command(generate_cli)
 cli.add_command(iotools_cli)
-cli.add_command(run_cli)
 
 
 def main() -> None:
--- a/clinica/cmdline.py
+++ b/clinica/cmdline.py
@@ -12,6 +12,7 @@ import clinica
 from clinica.engine.template import cli as generate_cli
 from clinica.iotools.converters.cli import cli as convert_cli
 from clinica.iotools.utils.cli import cli as iotools_cli
+from clinica.pipelines.cli import cli as run_cli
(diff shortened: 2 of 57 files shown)


clinica/pipelines/__init__.py
```diff
@@ -1 +1,31 @@
...
+    t1_freesurfer_template_cli,
+)
+from .t1_linear import flair_linear_cli, t1_linear_cli
+from .t1_volume import t1_volume_cli
+from .t1_volume_create_dartel import t1_volume_create_dartel_cli
+from .t1_volume_dartel2mni import t1_volume_dartel2mni_cli
+from .t1_volume_existing_template import t1_volume_existing_template_cli
+from .t1_volume_parcellation import t1_volume_parcellation_cli
+from .t1_volume_register_dartel import t1_volume_register_dartel_cli
+from .t1_volume_tissue_segmentation import t1_volume_tissue_segmentation_cli
```
ghisvail, line 31: Looks to me like we are exposing a lot of symbols for nothing. In my opinion, it should be possible to do:

`clinica/pipelines/__init__.py`
```python
from . import (
    deeplearning_prepare_data,
    dwi_connectome,
    [...],
    t1_volume_register_dartel,
    t1_volume_tissue_segmentation,
)
```

and have individual submodules be responsible for importing their own CLI package(s) like:

`clinica/pipelines/t1_linear/__init__.py`
```python
import flair_linear_cli
import t1_linear_cli
```

which would actually run the CLI registration code.

clinica/cmdline.py
```diff
@@ -84,7 +83,6 @@ def cli(verbose: bool) -> None:
 cli.add_command(convert_cli)
 cli.add_command(generate_cli)
 cli.add_command(iotools_cli)
-cli.add_command(run_cli)
```
ghisvail, line 87: That could probably stay as is, no? I used it as an example, but I think the dynamic registration pattern should start at the `clinica/pipelines/cli.py` level. WDYT?

clinica/__init__.py
```diff
@@ -3,3 +3,5 @@
 __all__ = ["__version__"]
 
 __version__ = version("clinica")
+
+from . import pipelines, pydra
```
ghisvail, line 7: Why do we need this?

ghisvail (commented):
I think it looks 🆗 modulo some comments

clinica/pydra/t1_linear/t1_linear_cli.py
```diff
@@ -3,11 +3,12 @@
 import clinica.pydra.engine_utils as pydra_utils
 import clinica.pydra.t1_linear.t1_linear as pydra_t1_linear
 from clinica.pipelines import cli_param
+from clinica.pipelines.cli import cli as run_cli
 
 pipeline_name = "pydra-t1-linear"
 
 
-@click.command(name=pipeline_name, hidden=True)
+@click.command(name=pipeline_name, hidden=False)
```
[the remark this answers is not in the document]
NicolasGensollen: I turned them on to make some checks while developing, I forgot to hide them back.
Thanks for catching that !
Should be fixed in [5dea4c9](https://github.com/aramis-lab/clinica/pull/787/commits/5dea4c9c4802b2ca5c344fa635df9fef5e07c3d7)

clinica/pipelines/__init__.py
NicolasGensollen: Good suggestion !
I re-organized the imports in [ffc5ff0](https://github.com/aramis-lab/clinica/pull/787/commits/ffc5ff0b16c2b5a1a2d77036eaa83deb1202cd1a)

ghisvail (approved):
LGTM

PR #788, opened 2022-11-16
[REF, DOC, TEST] Refactor `clinica.utils.freesurfer`

NicolasGensollen:
This PR proposes to improve the `clinica.utils.freesurfer` module by:

- Adding type hints
- Adding or completing docstrings
- Formatting all docstrings to Numpydoc
- Refactoring code to avoid duplication and antipatterns
- Making sure all functions fit on the screen
- Adding unit tests (WIP...)

The function `generate_regional_measures_alt` is a copy paste of `generate_regional_measures` with very little changes. This PR proposes to remove it and enable the user to provide a list of atlases (which seems to be the only important difference between the two functions).
This would need some proper testing to check that freesurfer pipelines work fine (WIP...).


clinica/utils/freesurfer.py
```diff
@@ -146,61 +41,182 @@ def generate_regional_measures_alt(
...
+            "Image %s does not contain FreeSurfer segmentation"
+            % prefix.replace("_", " | ")
         )
+    output_dir = Path(output_dir) if output_dir else seg_path / "regional_measures"
+    try:
+        os.makedirs(output_dir)
+    except OSError as exception:
+        # if dest_dir exists, go on, if its other error, raise
+        if exception.errno != errno.EEXIST:
+            raise
```
ghisvail, line 186: Using `exist_ok`

```suggestion
    os.makedirs(output_dir, exist_ok=True)
```

clinica/utils/freesurfer.py
```diff
@@ -146,61 +41,182 @@ def generate_regional_measures_alt(
...
+
+    atlases = atlases or ["desikan", "destrieux", "ba"]
+    prefix = _get_prefix(subject_id)
+    seg_path = Path(os.path.expanduser(segmentation_path))
+    stats_folder = seg_path / subject_id / "stats"
+
+    if not stats_folder.is_dir():
+        raise IOError(
+            "Image %s does not contain FreeSurfer segmentation"
+            % prefix.replace("_", " | ")
```
ghisvail, line 178: I think [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) would be more appropriate.

```suggestion
        raise FileNotFoundError(
            "Image %s does not contain FreeSurfer segmentation"
            % prefix.replace("_", " | ")
```

clinica/utils/freesurfer.py
```diff
@@ -218,289 +234,129 @@ def extract_image_id_from_longitudinal_segmentation(freesurfer_id):
...
 
-    Args:
-        stats_filename (string): path to the .stats file
-        info_type (string): 'volume', 'thickness' or 'area'
+    Parameters
+    ----------
+    stats_filename : PurePath
+        Path to the .stats file.
+
+    info_type : str
```
ghisvail, line 261: Same comment as above. Screams `enum` to me.

clinica/utils/freesurfer.py
```diff
@@ -218,289 +234,129 @@ def extract_image_id_from_longitudinal_segmentation(freesurfer_id):
...
     voxel_size = f.header.get_zooms()
     t1_size = f.header.get_data_shape()
-    if (
-        (voxel_size[0] * t1_size[0] > 256)
-        or (voxel_size[1] * t1_size[1] > 256)
-        or (voxel_size[2] * t1_size[2] > 256)
-    ):
-        # cprint(f"Setting MRI Convert to crop images to 256 FOV for {in_t1w} file.")
+    optional_flag = ""
+    if any([v * t > 256 for v, t in zip(voxel_size, t1_size)]):
```
ghisvail, line 359: We could drop `optional_flag`.

```python
if any([v * t > 256 for v, t in zip(voxel_size, t1_size)]):
    return " ".join([recon_all_args, "-cw256"])
else:
    return recon_all_args
```

ghisvail (changes requested):
LGTM, with some suggestions for additional type safety and better error reporting.

Feel free to discard what you don't like 👍

ghisvail (approved):
Changed from request changes to approve.

clinica/utils/freesurfer.py
NicolasGensollen: Indeed, thanks !

clinica/utils/freesurfer.py
```diff
@@ -1,70 +1,21 @@
...
-        try:
-            os.makedirs(output_dir)
-        except OSError as exception:
-            # if dest_dir exists, go on, if its other error, raise
-            if exception.errno != errno.EEXIST:
-                raise
-
-        # Generate TSV files for parcellation files
-        #
+def _read_stats_file(stats_filename: PurePath, column_type: str) -> pd.DataFrame:
```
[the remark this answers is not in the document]
NicolasGensollen: I agree this is cleaner. 👍 
I added you suggestion for both column types and info types.

clinica/utils/freesurfer.py
```diff
@@ -218,289 +234,129 @@ def extract_image_id_from_longitudinal_segmentation(freesurfer_id):
...
+    scalar_list : List[float]
+        list of values corresponding to the keys.
     """
     import warnings
 
-    import pandas
-
     try:
-        data = pandas.DataFrame({"label_name": name_list, "label_value": scalar_list})
+        data = pd.DataFrame({"label_name": name_list, "label_value": scalar_list})
```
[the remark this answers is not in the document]
NicolasGensollen: Hmm, I agree but at the same time it seems like this is this function's only purpose to receive two lists as inputs, which kind of makes sense if you look at how it is called (on lists built by concatenating lists over a for loop with no checks on lengths).
If you change the signature, then the input that makes the most sense to me would be a dictionary. But then the whole function is useless since you can simply call `pd.DataFrame(dico)`.
I'm not even sure adding the ValueError as I did makes sense, since these kind of errors were only reported as warnings before...
Will look into it further next week...

NicolasGensollen:
Thanks for the review @ghisvail ! 👍 
I'm keeping this as a draft since it's still lacking some tests in my opinion.
I'll try to finish this PR next week.

NicolasGensollen:
I believe this is ready for another round of reviews.

clinica/utils/freesurfer.py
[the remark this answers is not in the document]
NicolasGensollen: I deleted this function. As suspected, it wasn't needed.
Most functions in this module were doing a lot of unnecessary conversions between dictionaries, pandas dataframes, tuples of lists...
All this conversion logic should have been removed now in order to rely as much as possible on DataFrames, which hopefully makes things clearer.

clinica/utils/freesurfer.py
```diff
@@ -146,61 +56,217 @@ def generate_regional_measures_alt(
...
+        raise FileNotFoundError(f"Stats file {stats_filename} could not be found.")
+
+    return pd.read_csv(
+        stats_filename,
+        names=columns,
+        comment="#",
+        header=None,
+        delimiter="\s+",
+        dtype=str,
+    )
```
ghisvail, line 74: A bit more Pythonic:

```suggestion
    try:
        return pd.read_csv(
            stats_filename,
            names=columns,
            comment="#",
            header=None,
            delimiter="\s+",
            dtype=str,
        )
    except FileNotFoundError:
        raise FileNotFoundError(f"Stats file {stats_filename} could not be found.")
```

clinica/utils/freesurfer.py
```diff
@@ -218,289 +284,169 @@ def extract_image_id_from_longitudinal_segmentation(freesurfer_id):
...
-        raise exception
+    Returns
+    -------
+    secondary_stats_dict : pd.DataFrame
+        DataFrame containing the regions of the brain (in a column named 'label_name'),
+        and the corresponding volume/thickness/area values depending on the
+        input info type (in a column named 'label_value').
+    """
+    if info_type == InfoType.MEANCURV:
+        return pd.DataFrame()
```
ghisvail, line 406: The docstring says the function returns a dataframe with a certain shape (`label_name` and `label_value` columns), yet it can also return an empty dataframe not respecting this shape.

I wonder whether we would not be better off returning `None` and make the signature `-> Optional[pd.DataFrame]`. Unless there is an implementation detail that justifies it of course.

clinica/utils/freesurfer.py
```diff
@@ -218,289 +284,169 @@ def extract_image_id_from_longitudinal_segmentation(freesurfer_id):
...
+    ]
+    regions = [r[0] for r in regions_stats if r[0] != ""]
+    values = [r[1] for r in regions_stats if r[1] != ""]
+    if len(regions) != len(values):
+        raise ValueError(
+            f"In stats file {stats_filename}, the extracted secondary stats "
+            f"have different lengths {len(regions)} vs. {len(values)}. "
+            "Please check the lines starting with '# Measure' in this file."
+        )
+    return pd.DataFrame({"label_name": regions, "label_value": values})
```
ghisvail, line 424: The whole thing could be simplified to this:

```suggestion
    return pd.DataFrame.from_records(regions_stats, columns={"label_name", "label_value"})
```

This logic looks very unsafe to me, as we can literally shift one column with respect to the other accidentally.

```python
    regions = [r[0] for r in regions_stats if r[0] != ""]
    values = [r[1] for r in regions_stats if r[1] != ""]
``` 

In my opinion, `_extract_region_and_stat_value` should be responsible for filtering lines containing empty values and return proper records.

ghisvail: I suspect that `_extract_region_and_stat_value` can only return `("", "")` and not `("", "right_val")` or `("left_val", "")`. But the code should prevent invalid states be computed in the first place, hence why splitting the columns and filtering them individually is wrong in my opinion.

Instead, `_extract_region_and_stat_value` should return records without empty values, which would require no further splitting or filtering.

ghisvail (changes requested):
We are getting there, some suggestions to improve robustness.

clinica/utils/freesurfer.py
NicolasGensollen: Thanks for the suggestion @ghisvail ! Using `from_records` is much better indeed ! 👍 
For the second point, I also agree. I made `_extract_region_and_stat_value` return `None` if nothing could be extracted. The list of tuples then need to be cleaned from `Nones` to be a valid input to the data frame constructor. Hope this answers your concerns.

clinica/utils/freesurfer.py
NicolasGensollen: I followed your suggestion to return `None`. To be honest, I'm not really sure to understand why mean curvature is considered as a special case since nothing is ever done for this type of information. I kind of feel why we should totally get rid of it... 🤔

ghisvail: I agree. I guess we'll have it figured out when converting it to Pydra.

ghisvail (approved):
Approved with a more Pythonic suggestion. Thanks for taking care of this chuky set of legacy code.

ghisvail (approved):

PR #790, opened 2022-11-17
[ENH] `clinica_pipeline` decorator

NicolasGensollen:
Following #787 which made `clinica run` subcommand registration dynamic, this PR proposes a first very simple implementation of the `clinica_pipeline` decorator.
For now, it only takes care about subcommand registration.


ghisvail (approved):

ghisvail (approved):
That's a very good start I think 👍

ghisvail:
I wonder whether we could go the extra mile and automatically handle all the `click` sugaring stuff when defining a pipeline. It could start with like `@clinica_pipeline(name)` propagating the name parameter to `click.command(name=name)` for instance. Just a random thought.

NicolasGensollen:
> I wonder whether we could go the extra mile and automatically handle all the `click` sugaring stuff when defining a pipeline. It could start with like `@clinica_pipeline(name)` propagating the name parameter to `click.command(name=name)` for instance. Just a random thought.

Yes, definitely something to investigate.

Issue #786, opened 2022-11-10
error converting AIBL to BIDS


JiyangJiang:
Hi Clinica group,

I downloaded all AIBL MRI and PET (I clicked pre/post-processing tickboxes) from IDA, and clinical data from AIBL website.

I was trying to convert AIBL data to BIDS, but came across the following error. Any advice is much appreciated!

Many thanks,
Jiyang

+++++++++++++++++++++++++++++++++++++++++++++++++++

......
2022-11-10 21:30:22,062:INFO:Scans information for sub-AIBL1010 ses-M54 not found.
2022-11-10 21:30:22,063:INFO:Scans information for sub-AIBL1010 ses-M36 not found.
2022-11-10 21:30:22,064:INFO:Scans information for sub-AIBL108 ses-M18 not found.
2022-11-10 21:30:22,065:INFO:Scans information for sub-AIBL108 ses-M00 not found.
2022-11-10 21:30:22,066:INFO:Scans information for sub-AIBL108 ses-M54 not found.
2022-11-10 21:30:22,066:INFO:Scans information for sub-AIBL108 ses-M36 not found.
2022-11-10 21:30:22,067:INFO:Scans information for sub-AIBL1534 ses-M18 not found.
2022-11-10 21:30:22,068:INFO:Scans information for sub-AIBL1534 ses-M00 not found.
2022-11-10 21:30:22,069:INFO:Scans information for sub-AIBL1534 ses-M54 not found.
2022-11-10 21:30:22,069:INFO:Scans information for sub-AIBL1534 ses-M36 not found.
2022-11-10 21:30:22,070:INFO:Scans information for sub-AIBL1357 ses-M18 not found.
2022-11-10 21:30:22,071:INFO:Scans information for sub-AIBL1357 ses-M00 not found.
2022-11-10 21:30:22,072:INFO:Scans information for sub-AIBL1357 ses-M54 not found.
2022-11-10 21:30:22,073:INFO:Scans information for sub-AIBL1357 ses-M36 not found.
Traceback (most recent call last):
  File "/home/jiyang/miniconda3/envs/clinicaEnv/bin/clinica", line 8, in <module>
    sys.exit(main())
  File "/home/jiyang/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/clinica/cmdline.py", line 89, in main
    cli()
  File "/home/jiyang/miniconda3/envs/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
[...]

PR #799, opened 2022-11-22
FIX: Only list first-level non-hidden directories

ghisvail:
Found when trying to reproduce #786 with a local AIBL dataset containing a collection and associated XML metadata.

Without this fix, the converter fails with the following error:

```
ValueError: invalid literal for int() with base 10: 'AIBL_1262_PET-CTAC_SUM_S236830_I453888.xml'
```

Resulting from the fact that the `listdir_nohidden` utility function will return first-level directories **and files**, unlike what its name and docstring suggest 🤦 

There is a lot of horrible code in this converter so I went for the immediate fix, which is to replace the implementation with one based on `pathlib` and use `Path.is_dir` to filter first-level files from what's being returned. I also took this opportunity to update the docstring and add tests.

### DIFF
--- a/clinica/iotools/converters/aibl_to_bids/aibl_utils.py
+++ b/clinica/iotools/converters/aibl_to_bids/aibl_utils.py
@@ -1,22 +1,30 @@
 """Utils to convert AIBL dataset in BIDS."""
 from functools import partial
-from typing import Sequence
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
LGTM, thanks for the fix and the non regression test !

PR #813, opened 2022-11-24
[MAINT] Put a lower bound on `pydra-nipype1` version

NicolasGensollen:
Make sure we install `pydra-nipype1 >= 0.2` since we need Omar's fix to run our Pydra-based pipelines.


ghisvail (approved):
Absolutely, nice catch!

PR #814, opened 2022-11-24
[FIX] Add pydra register dartel to cli and fix a couple issues

NicolasGensollen:
Follow-up PR of #768 : 

- Add subcommand to the CLI
- Improve docstring
- Fix `task_prepare_dartel_input_images`

I tried it with these fixes and it seems to work as expected.

### DIFF
--- a/clinica/pydra/t1_volume/register_dartel/cli.py
+++ b/clinica/pydra/t1_volume/register_dartel/cli.py
@@ -26,26 +26,36 @@ def cli(
     group_label: str,
     dartel_tissues: List[int] = (1, 2, 3),
 ) -> None:
-    """Inter-subject registration using Dartel (using an existing Dartel template).
+    """Inter-subject registration using an existing Dartel template (Pydra engine).
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


clinica/pydra/t1_volume/register_dartel/cli.py
```diff
@@ -24,26 +26,36 @@ def cli(
...
+        Path to the CAPS directory.
 
     group_label : str
-        User-defined identifier to target a specific group of subjects. For this pipeline, it is associated to the
-        DARTEL template that you had created when running the t1-volume pipeline.
+        User-defined identifier to target a specific group of subjects.
+        For this pipeline, it is associated to the DARTEL template that was created
+        when running the t1-volume pipeline.
+
+    dartel_tissues : List[int], optional
```
ghisvail, line 44: FYI, `numpydoc` recommends to use more litteral type names like `list of int` rather than raw Python annotations like `List[int]`.

ghisvail (approved):
LGTM

PR #822, opened 2022-12-01
FIX: Test instantiation for all Pydra pipelines

ghisvail:
Fixes the following warning:

`RuntimeWarning: The empty option group "Common pipelines options" was found (line 56) for "cli". The group will not be added.`

Also add  instantiation tests for all Pydra pipelines (they are hidden but they can still be tested).

### DIFF
--- a/clinica/pydra/t1_volume/create_dartel/__init__.py
+++ b/clinica/pydra/t1_volume/create_dartel/__init__.py
@@ -1,0 +1,1 @@
+from . import cli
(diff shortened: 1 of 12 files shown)


ghisvail:
Jenkins docs will fail because of its fragility.

clinica/pydra/t1_volume/create_dartel/__init__.py
```diff
@@ -0,0 +1 @@
+from . import cli
```
NicolasGensollen, line 1: I'm wondering whether we should homogenize the naming of the cli files... Some pydra pipelines (like t1-linear or pet-volume use the connection `name_of_the_pipeline_cli.py`, while some use just `cli.py`).

ghisvail:
I agree the naming convention should be harmonised. I did not want to rename too much stuff to keep the PR light.

NicolasGensollen:
> I agree the naming convention should be harmonised. I did not want to rename too much stuff to keep the PR light.

Let's do it in another PR then 👍

PR #832, opened 2022-12-09
[FIX] Expose `fwhm` to `t1_volume_template_tpm_in_mni`

### LINKED ISSUE #830: FWHM seems unused in statistics-volume pipeline with t1-volume data
In statistics-volume pipeline, [FWHM can be set and is 8 by default](https://github.com/aramis-lab/clinica/blob/02aaa90321ec853fb09ea2f965b1b338921a1845/clinica/pipelines/statistics_volume/statistics_volume_cli.py#L28).

But during data loading, if `t1-volume` is used as `orig_input_data_volume`, then `t1_volume_template_tpm_in_mni` is [called](https://github.com/aramis-lab/clinica/blob/02aaa90321ec853fb09ea2f965b1b338921a1845/clinica/pipelines/statistics_volume/statistics_volume_pipeline.py#L119), that loads [non smoothed data](https://github.com/aramis-lab/clinica/blob/02aaa90321ec853fb09ea2f965b1b338921a1845/clinica/utils/input_files.py#L417).

This is different from `pet-volume` pipeline where the `fwhm` is used to define the [`information_dict`](https://github.com/aramis-lab/clinica/blob/02aaa90321ec853fb09ea2f965b1b338921a1845/clinica/pipelines/statistics_volume/statistics_volume_pipeline.py#L115).

In the output files though, `fwhm` is set to 8 while it seems that the volumes used to compute the statistics were not smoothed.

Am I missing something here ?

NicolasGensollen:
Fixes #830 

This PR proposes a possible fix for #830 by adding a new `fwhm` argument to `t1_volume_template_tpm_in_mni` in order to query files obtained with smoothing, similar to what is done for pet.

@AudreyDuran I still need to test it but feel free to take a look and check if this patch fixes your problem.

### DIFF
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -401,24 +401,49 @@ def t1_volume_native_tpm_in_mni(tissue_number, modulation):
     }
 
 
-def t1_volume_template_tpm_in_mni(group_label, tissue_number, modulation):
+def t1_volume_template_tpm_in_mni(group_label, tissue_number, modulation, fwhm=None):
+    """Build the dictionary required by clinica_file_reader to get the tissue
+    probability maps based on group template in MNI space.
+
+    Parameters
+    ----------
+    group_label : str
+        Label used for the group of interest.
+
+    tissue_number : int
+        An integer defining the tissue of interest.
+
+    modulation : {"on", "off"}
+        Whether modulation is on or off.
+
+    fwhm : int, optional
+        The smoothing kernel in millimeters.
+
+    Returns
+    -------
+    dict :
+        Information dict to be passed to clinica_file_reader.
+    """
     import os
 
     from .spm import INDEX_TISSUE_MAP
 
     pattern_modulation = "on" if modulation else "off"
     description_modulation = "with" if modulation else "without"
+    fwhm_key_value = f"_fwhm-{fwhm}mm" if fwhm else ""
+    fwhm_description = f"with {fwhm}mm smoothing" if fwhm else "with no smoothing"
+
     return {
         "pattern": os.path.join(
             "t1",
             "spm",
             "dartel",
             f"group-{group_label}",
-            f"*_T1w_segm-{INDEX_TISSUE_MAP[tissue_number]}_space-Ixi549Space_modulated-{pattern_modulation}_probability.nii*",
+            f"*_T1w_segm-{INDEX_TISSUE_MAP[tissue_number]}_space-Ixi549Space_modulated-{pattern_modulation}{fwhm_key_value}_probability.nii*",
         ),
         "description": (
             f"Tissue probability map {INDEX_TISSUE_MAP[tissue_number]} based "
-            f"on {group_label} template in MNI space (Ixi549) {description_modulation} modulation."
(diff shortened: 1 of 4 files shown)


AudreyDuran:
LGTM thanks 👍

clinica/utils/input_files.py
```diff
@@ -401,24 +401,49 @@ def t1_volume_native_tpm_in_mni(tissue_number, modulation):
     }
 
 
-def t1_volume_template_tpm_in_mni(group_label, tissue_number, modulation):
+def t1_volume_template_tpm_in_mni(group_label, tissue_number, modulation, fwhm=None):
```
AudreyDuran, line 404: Just to be sure, can `fwhm` be set to None in click (since the default is 8) ?

NicolasGensollen: I implemented it this way for backward compatibility concerns.

I was however wondering the same thing. Indeed, when you look at the code of Dartel2MNI, especially here:

https://github.com/aramis-lab/clinica/blob/02aaa90321ec853fb09ea2f965b1b338921a1845/clinica/pipelines/t1_volume_dartel2mni/t1_volume_dartel2mni_pipeline.py#L265-L268

It feels like you *could* intentionally ask for no smoothing by not setting `fwhm` in the CLI, but this is not what's currently happening since there is a default value implemented both in the CLI and in the `check_parameters` function. Unless I missed something, there is currently no way (available to the user) to have the smoothing parameter equal to None at this point.

I agree that this is very confusing. I think that if smoothing is required, then we should remove the condition in the workflow, otherwise, we should have no smoothing if the user didn't pass explicitly a value for fwhm in the CLI. The docs could also be clarified to reflect and stress this point.

NicolasGensollen: I agree. Based on the docs, I'd expect the stats to be computed on smoothed volumes only if I provide a value for `fwhm` in the CLI. If nothing is provided for `fwhm`, it feels very weird to have the stats computed on 8mm-smoothed images.

NicolasGensollen: After looking into it, I reverted the modifications I initially made to `t1_volume_parcellation`, leaving the current behavior in place. I only changed the `statistics_volume` pipeline to use the `fwhm` value provided (which should match the behavior for pet).

clinica/utils/input_files.py
```diff
@@ -401,24 +401,49 @@ def t1_volume_native_tpm_in_mni(tissue_number, modulation):
...
+        Label used for the group of interest.
+
+    tissue_number : int
+        An integer defining the tissue of interest.
+
+    modulation : {"on", "off"}
+        Whether modulation is on or off.
+
+    fwhm : int, optional
+        The smoothing kernel in millimeters.
```
ghisvail, line 420: Can FWHM be a float?

ghisvail: We would not want it based on our naming convention, but I am asking if it makes sense from a domain perspective.

ghisvail (approved):

clinica/utils/input_files.py
NicolasGensollen: I left them as integers because this is how they are represented in Clinica, but I believe they could (and should) be floats.

PR #834, opened 2022-12-12
FIX: FutureWarning in pandas.concat

ghisvail:
Fix warning `FutureWarning: In a future version of pandas all arguments of concat except for the argument 'objs' will be keyword-only.` reported in CI with the latest version of `pandas`. The `axis` parameter will become keyword-only and we are using the default value anyway. Better just remove it.


MatthieuJoulot (approved):
Thanks @ghisvail, LGTM ! 
Tested on `ukb-to-bids`as well, still works the same :)

PR #837, opened 2022-12-12
[TEST, CI] Refactor non regression tests

NicolasGensollen:
This PR proposes to split non-regression tests into three classes: fast, normal, and slow, with the following goals in mind:

- Slow tests (marked with `@pytest.mark.slow`) will continue to run as before, once a week over the weekend.
- Fast tests (marked with `@pytest.mark.fast`) will be added to the pipeline running on each commit. We should keep the running time below 20 minutes for this to make sense.
- Normal tests and fast tests will be run once a day (probably over night ??). I think that a running time of a few hours should be fine.

Atm I am guilty of a nasty copy-paste of the non regression's Jenkins file. I have a limited knowledge of them but I'd be surprised if no factorization was possible...


ghisvail:
I'd suggest to use the `.Jenkinfile` suffix which is detected as groovy syntax by popular IDEs.

For instance:

```
.jenkins/
    Jenkinsfile
    nonregression-fast.Jenkinsfile
    nonregression-slow.Jenkinsfile
```

NicolasGensollen:
> I'd suggest to use the `.Jenkinfile` suffix which is detected as groovy syntax by popular IDEs.
> 
> For instance:
> 
> ```
> .jenkins/
>     Jenkinsfile
>     nonregression-fast.Jenkinsfile
>     nonregression-slow.Jenkinsfile
>     docs.Jenkinsfile
> ```
> 
> EDIT: Added the docs pipeline.

Indeed, good suggestion.

NicolasGensollen:
> > Atm I am guilty of a nasty copy-paste of the non regression's Jenkins file. I have a limited knowledge of them but I'd be surprised if no factorization was possible.
> 
> It may require diving into Jenkins support for [shared libraries](https://www.jenkins.io/doc/book/pipeline/shared-libraries/).

Thanks for the link @ghisvail ! 👍 
I'll take a closer look, but I think the refactoring of the Jenkins files could be the subject of another PR.

clinica/utils/filemanip.py
```diff
@@ -16,7 +15,7 @@ def _zip_unzip_nii(in_file: str, same_dir: bool, compress: bool):
     if (in_file is None) or isinstance(in_file, _Undefined):
         return None
 
-    if isinstance(in_file, Iterable):
+    if isinstance(in_file, (list, tuple)):
         return [_zip_unzip_nii(f, same_dir, compress) for f in in_file]
 
     in_file = Path(in_file)
```
ghisvail, line 21: We might be able to simplify the whole None, str, and sequence-type handling with:

```python
    try:
        # Assuming in_file is path-like.
        in_file = Path(in_file)
    except TypeError:
        try:
            # Assuming in_file is a sequence type.
            return [_zip_unzip_nii(f, same_dir, compress) for f in in_file]
        except TypeError:
            # All other cases.
            return None
```

test/nonregression/testing_tools.py
```diff
@@ -8,6 +8,20 @@
 import pandas as pd
 
 
+def configure_paths(
+    base_dir: PathLike,
+    tmp_path: PathLike,
+    name: str,
+) -> Tuple[PathLike, PathLike, PathLike]:
```
ghisvail, line 15: You are gonna have a type alignment error since `os.PathLike` does not implement path concat with `/`. You'd need to either use `Path` or `PurePath` all the way, or do explicit conversion within the function.

```suggestion
    base_dir: Path,
    tmp_path: Path,
    name: str,
) -> Tuple[Path, Path, Path]:
```

Considering this is just a utility function, `os.PathLike` is not much interesting to be honest.

ghisvail (approved):
LGTM with a few non-blocking suggestions.

test/nonregression/testing_tools.py
NicolasGensollen: Thanks! 👍  I saw it but forgot to fix it before asking for a review.

clinica/utils/filemanip.py
NicolasGensollen: Thanks! Much better indeed 👍

ghisvail:
The fast group of non-reg tests took a bit less than 18 minutes.

NicolasGensollen:
> The fast group of non-reg tests took a bit less than 18 minutes.

We can remove tests from this group if you think this is too long (or add some tests if you think it is acceptable to wait a little longer)

PR #838, opened 2022-12-12
MNT: Upgrade lock file to version 2

ghisvail:
Poetry 1.3 introduced a new format for its lock file. Since Poetry has been upgraded on conda-forge and on GitHub CI, now is a good time to upgrade our lock file to minimize the potential for conflicts in future merge requests.


.jenkins/doc_build
```diff
@@ -18,7 +18,7 @@ pipeline {
       steps {
         sh '''
           source "${CONDA_HOME}/etc/profile.d/conda.sh"
-          conda create -y -p "${CONDA_ENV}" python=3.9
+          conda create -y -p "${CONDA_ENV}" python=3.9 poetry>=1.3
```
NicolasGensollen, line 21: I believe we installed poetry through pipx on all the CI machines. I think we need to manually upgrade on all of them for this PR to work, right ?

ghisvail: I thought the Conda version would take priority over the one installed through `pipx`, but maybe not 🤔

ghisvail: I have reverted the modifications on the Jenkins CI files.

ghisvail:
Requires upgrading Poetry on Jenkins to version 1.2.2 minimum, which supports reading the new version of the lock file.

NicolasGensollen (approved):
Upgrade is done.

PR #841, opened 2022-12-14
[CI] Use shared libs for Jenkins pipelines

NicolasGensollen:
Investigate the usage of shared libraries to refactor our Jenkins pipelines.

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -1,3 +1,5 @@
+@Library('clinica-jenkins-sharedlibrary') _
+
 pipeline {
   options {
     timeout(time: 2, unit: 'HOURS')
@@ -46,6 +48,7 @@ pipeline {
                   clinica --help
                   conda list
                 '''
+		helloWorldExternal(name: "Nicolas", dayOfWeek:"Saturday")
               }
             }
             stage('Test instantiation') {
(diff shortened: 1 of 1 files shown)


NicolasGensollen:
Alright, it seems like this is working as expected.

As suggested by @ghisvail I had a look at [shared libraries](https://www.jenkins.io/doc/book/pipeline/shared-libraries/) for improving our Jenkins pipelines.

The main objectif was to avoid all the copy-pasting of small scripts which is far from coding best practices and was already resulting in inconsistencies between steps.

I have created a library called `clinica-jenkins-sharedlib` under my GitHub profile : https://github.com/NicolasGensollen/clinica-jenkins-sharedlib

This library proposes a first simple implementation of common tasks we use in our Jenkins pipelines.

For now, I only changed one JenkinsFile but the same modifications will apply to the other files we have.

I think the next steps could be:

- [ ] Someone (@ghisvail ?) review both this PR and the shared library
- [ ] If we agree that this is an improvement, I'll transfer ownership of the shared library to the AramisLab organization.
- [ ] Once that's done, I'll update the Jenkins configuration to point to the new location and rerun the CI to make sure everything is running as it should
- [ ] Merge this PR and update other Jenkins files (in other PRs ?)
- [ ] Investigate if the library could be used by other similar projects like ClinicaDL

ghisvail:
Closing this PR, since we got rid of Jenkins in #971

PR #842, opened 2022-12-14
[UKB2BIDS] Correct the file that is used for FLAIR

MatthieuJoulot:
The current implentation takes as input the `T2_FLAIR.nii.gz`. This image is registered on the T1 image and it is therefore not the one we wish to convert to BIDS, as it is not the rawest image, which would be `T2_FLAIR_orig_defaced.nii.gz`.


ghisvail (approved):
LGTM. Thanks for your clear explanation as to why this change was needed.

PR #853, opened 2022-12-21
[CI] Clean typos in Jenkins files

NicolasGensollen:
When triggering the slow non regression tests, I get an immediate `unexpected char: 0xFFFF @ line 493, column 1` which means that the file couldn't be parsed successfully. I had a look but couldn't find any obvious syntax mismatch for the moment. This PR just fixes a few small typos I found while reading through.


ghisvail:
I provided #854 to help. There is probably a mixup of tabs and spaces somewhere.

NicolasGensollen:
Looks like #854 fixed the issue with the Jenkinsfile ! Thanks @ghisvail ! 👍 
Closing this then...

Issue #827, opened 2022-12-05
run statistics-volume-correction raises "AssertionError: pattern argument cannot start with char: /"


AudreyDuran:
**Describe the bug**
Running `statistics-volume-correction` raises an `AssertionError`.

**To Reproduce**
1. run `statistics-volume` pipeline ; 
2. run the volume-correction pipeline: 
`clinica run statistics-volume-correction clinica_CAPS clinica_CAPS/groups/group-treatedvsplacebo/statistics_volume/group_comparison_measure-graymatter/group-treatedvsplacebo_treated-lt-placebo_measure-graymatter_fwhm-8_TStatistics.nii 3.1 5.273 6.059 25 16`

**Expected behavior**
The pipeline should run without errors

**Screenshots**

Raises the following error:
```bash
Traceback (most recent call last):
  File "/workspace/clinicaEnv/bin/clinica", line 8, in <module>
    sys.exit(main())
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/clinica/cmdline.py", line 89, in main
    cli()
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/clinica/pipelines/statistics_volume_correction/statistics_volume_correction_cli.py", line 83, in cli
    else pipeline.run()
  File "/workspace/clinicaEnv/lib/python3.8/site-packages/clinica/pipelines/engine.py", line 261, in run
    self.build()
[...]

PR #858, opened 2022-12-28
[FIX] Fix `statistics-volume-correction` call to `_check_information()`

AudreyDuran:
Closes issue https://github.com/aramis-lab/clinica/issues/827

### DIFF
--- a/clinica/pipelines/statistics_volume_correction/statistics_volume_correction_cli.py
+++ b/clinica/pipelines/statistics_volume_correction/statistics_volume_correction_cli.py
@@ -11,7 +11,7 @@ pipeline_name = "statistics-volume-correction"
 @clinica_pipeline
 @click.command(name=pipeline_name)
 @cli_param.argument.caps_directory
-@click.argument("t_map", type=click.Path(exists=True, resolve_path=True))
+@click.argument("t_map", type=str)
 @click.argument("height_threshold", type=float)
 @click.argument("FWEp", type=float)
 @click.argument("FDRp", type=float)


NicolasGensollen (commented):
Hi @AudreyDuran 
Thanks for looking into this.
I'm probably missing something but I don't understand why you end up having this error. What happens when you run the following command ?

```
$ clinica run statistics-volume-correction clinica_CAPS group-treatedvsplacebo_treated-lt-placebo_measure-graymatter_fwhm-8_TStatistics.nii 3.1 5.273 6.059 25 16
```

The group reader should find the file by building the following pattern:

https://github.com/aramis-lab/clinica/blob/7034cd696c5dd143def9709014b62b3ee0b8e83f/clinica/utils/inputs.py#L665-L668

AudreyDuran:
If I don't give the path to the Tmap, I get the following error (file not found).
image

So it should be a path and not only a filename given in input. As a consequence, `check_information` is failing [here](https://github.com/aramis-lab/clinica/blob/444e4a4e711d6a373fae07412fa378da1e356572/clinica/utils/inputs.py#L311) because `pattern` is the absolute path to the Tmap and thus begins with a `/`.

NicolasGensollen:
You should not pass the full path to the Tmap, only the filename, as done here, in the tests, for example:

https://github.com/aramis-lab/clinica/blob/7034cd696c5dd143def9709014b62b3ee0b8e83f/test/nonregression/pipelines/test_run_pipelines_stats.py#L200-L202

The full path to the caps folder gets added in the group reader, and everything in between the root of the CAPS folder and the filename shouldn't matter in theory. So, I don't know why you get the "path does not exist" error at the moment. I'll take a closer look tomorrow.

NicolasGensollen:
Hi @AudreyDuran 
I had a look and I was able to reproduce your error. The problem is in the parametrization of the CLI and not in the pattern nor the group reader's logic. More precisely, the `t_map` argument of the CLI is defined as being a path to file that must exist : 

https://github.com/aramis-lab/clinica/blob/7034cd696c5dd143def9709014b62b3ee0b8e83f/clinica/pipelines/statistics_volume_correction/statistics_volume_correction_cli.py#L14

First of all, this is inconsistent with the type hint of the function which says it should be a string. Besides, it is inconvenient because we are already specifying the path to the CAPS folder.

Moreover, if you respect the CLI and provide the full path to the tmap file, you indeed end up raising the error with the leading "\\" but for good reasons because the pattern isn't what you expect.

Note that this error is not caught in the test I linked above because the tmap file name isn't passed through the CLI, but directly as a parameter.

I think the correct fix is to change the CLI parametrization to accept a simple string. That is, make the following change:

```
-@click.argument("t_map", type=click.Path(exists=True, resolve_path=True))
+@click.argument("t_map", type=str)
```

If you agree with this proposition, could you update the fix in this PR ?

AudreyDuran:
Hi @NicolasGensollen, yes indeed, thanks for the review ! The MR is ready

NicolasGensollen (approved):
LGTM, thanks @AudreyDuran !

PR #861, opened 2023-01-03
MNT: Enable install and testing for Python 3.11

ghisvail:
Requires wheels for `h5py`, see https://github.com/h5py/h5py/issues/2146

~~I thought VTK would be the only obstacle, I was wrong.~~

We now have wheels for Python 3.11.


NicolasGensollen (approved):
Great! Thanks @ghisvail !

PR #863, opened 2023-01-04
MNT: Bump pydra-bids to version 0.0.6

ghisvail:
Required to allow co-install with the latest version of `pydra-freesurfer`. I'll relax the versioning constraints on `pydra` for both `pydra-bids` and `pydra-freesurfer` to avoid painful lock step updates like this.


ghisvail:
My bad, `pydra-bids` needs a new release first.

PR #867, opened 2023-01-16
[CI, TEST] Run SPM instantiation tests in same process

NicolasGensollen:
Closes #859 

Alternative possible solution for the intermittent errors we currently have in instantiation tests. The error happens when running tests in parallel, for pipelines relying on SPM.

After investigation, I'm still unsure why this happens and why it seems to happen more frequently on `clinica-ubuntu-3`... I tried with SPM standalone and without and I get the same results.

In #859, I was proposing to run the instantiation tests in a single process to avoid these failures.
This PR proposes instead to keep the parallelism at the pytest session level, but group the tests using SPM in a single test function. 
It also proposes to clean the names of the test functions (remove all capital letters which are not PEP8 compliant), and explicitly skip a test which was commented (I don't know why, so the skip reason I've put isn't super informative...).

### DIFF
--- a/test/instantiation/test_instantiate_all_pipelines.py
+++ b/test/instantiation/test_instantiate_all_pipelines.py
@@ -4,21 +4,16 @@ import warnings
 from os import fspath
 from pathlib import Path
 
-# Small unit tests for all pipelines
-##
-# test if instantiation and building of workflows is working
-
+import pytest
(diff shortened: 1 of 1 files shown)


MatthieuJoulot (approved):
LGTM !

PR #869, opened 2023-01-18
[REF, TEST] Refactor and test pipeline handling

NicolasGensollen:
The file `Clinica/iotools/utils/pipeline_handling.py` contains some functions used to extract information from a CAPS folder depending on the pipeline. This information is then merged with previously obtained information from BIDS folder. This is primarily used by the `merge-tsv` iotools.

Currently, there is one function per supported pipeline (`dwi_dti`, `pet_volume`, `t1_freesurfer_longitudinal`, `t1_freesurfer`, and `t1_volume`). 
Each function duplicates the whole CAPS folder parsing logic, which is basically a bunch of imbricated `for` loops with some file name parsing and dataframe manipulations. This results in a lot of copy pasted code and hides the specificities of each pipeline.

There is an exception for the pipelines `t1_volume` and `pet_volume` which already share a common implementation. However, it seems to me that this factorization can be done for all supported pipelines.

This PR proposes to refactor these functions to rely on a common parsing logic (implemented in `_extract_metrics_from_pipeline()`) and put the pipeline-specific parsing logic in smaller functions (`_get_label_list()`, `_skip_atlas()`, `_get_mod_path()`, and `_get_atlas_name()`).
It also proposes to add some unit tests for these functions.

This is still WIP as I need to add more tests and make sure I didn't break anything while refactoring.


MatthieuJoulot (changes requested):
First review ! Looking good so far

NicolasGensollen:
Thanks for the review @MatthieuJoulot ! 👍 
I've done some refactoring of the tests as you suggested.

MatthieuJoulot (approved):
LGTM !

PR #872, opened 2023-01-23
[FIX] Handle 4D images with dummy fourth dimension

### LINKED ISSUE #871: [PET] Improve how to 4D images with 1 volume are handled
Hi!

Following Discussion https://github.com/aramis-lab/clinica/discussions/870, it turns out that 4D images with 1 volume are not taken into account in PET pipelines:

- `pet-surface*`:
https://github.com/aramis-lab/clinica/blob/4e5f5d93f7629d726e4efc6a70ee16c603d62dab/clinica/pipelines/pet_surface/pet_surface_utils.py#L949-L952

- `pet-volume*`:
https://github.com/aramis-lab/clinica/blob/4e5f5d93f7629d726e4efc6a70ee16c603d62dab/clinica/pipelines/pet_volume/pet_volume_utils.py#L13-L16

A solution would be:
```python
    if len(img.shape) == 4:
        if img.shape[3] > 1:
            error_msg = (
                f"Clinica does not handle 4D volumes for {image_id.replace('_', ' | ')}"
            )

```
(I have a doubt but a oneliner `if len(img.shape) == 4 and img.shape[3] > 1:` could also work)

Best,
Alex

NicolasGensollen:
Closes #871

### DIFF
--- a/clinica/pipelines/pet_volume/pet_volume_utils.py
+++ b/clinica/pipelines/pet_volume/pet_volume_utils.py
@@ -1,9 +1,9 @@
 def init_input_node(pet_nii):
-    from clinica.utils.filemanip import check_img_3d, get_subject_id
+    from clinica.utils.filemanip import get_subject_id, load_img_3d
     from clinica.utils.ux import print_begin_image
 
     image_id = get_subject_id(pet_nii)
-    pet_nii = check_img_3d(pet_nii)
+    pet_nii = load_img_3d(pet_nii)
     print_begin_image(image_id)
 
     return pet_nii
--- a/clinica/pipelines/pet_volume/pet_volume_utils.py
+++ b/clinica/pipelines/pet_volume/pet_volume_utils.py
@@ -1,9 +1,15 @@
 def init_input_node(pet_nii):
-    from clinica.utils.filemanip import get_subject_id, load_img_3d
+    from clinica.utils.filemanip import get_subject_id, load_volume
+    from clinica.utils.stream import cprint
     from clinica.utils.ux import print_begin_image
 
     image_id = get_subject_id(pet_nii)
-    load_img_3d(pet_nii)
+    try:
+        load_volume(pet_nii)
+    except ValueError as e:
+        error_msg = f"Clinica could not load volumes for {image_id.replace('_', ' | ')}. {str(e)}"
+        cprint(error_msg, lvl="error")
+        raise ValueError(error_msg)
     print_begin_image(image_id)
 
     return pet_nii
--- a/clinica/pipelines/pet_volume/pet_volume_utils.py
+++ b/clinica/pipelines/pet_volume/pet_volume_utils.py
@@ -1,23 +1,9 @@
 def init_input_node(pet_nii):
-    import nibabel as nib
-
-    from clinica.utils.filemanip import get_subject_id
-    from clinica.utils.stream import cprint
+    from clinica.utils.filemanip import check_img_3d, get_subject_id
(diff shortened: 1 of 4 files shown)


clinica/pipelines/pet_volume/pet_volume_utils.py
```diff
@@ -1,23 +1,9 @@
...
-    img = nib.load(pet_nii)
-    if len(img.shape) == 4:
-        error_msg = (
-            f"Clinica does not handle 4D volumes for {image_id.replace('_', ' | ')}"
-        )
-        cprint(error_msg, lvl="error")
-        raise NotImplementedError(error_msg)
-
-    # Print begin message
+    pet_nii = load_img_3d(pet_nii)
```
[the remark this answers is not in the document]
NicolasGensollen: For some reason, I thought the function was returning `img`, and not the input image path.
So this will most likely break as the next node will try to unzip an instance of a `Nifti1Image`.
I'm going to remove the mutation of `pet_nii` to keep the previous behavior.
Thanks for catching this !

test/unittests/utils/test_filemanip.py
```diff
@@ -179,6 +181,47 @@ def test_zip_unzip_nii(tmp_path):
...
+def test_load_img_3d(tmp_path, case, test_image):
+    from clinica.utils.filemanip import load_img_3d
+
+    with pytest.raises(
+        FileNotFoundError,
+        match="No such file or no access: 'foo'",
+    ):
+        load_img_3d("foo")
+    (tmp_path / "sub-01").mkdir()
+    (tmp_path / "sub-01" / "ses-M000").mkdir()
```
ghisvail, line 208:
```suggestion
    (tmp_path / "sub-01" / "ses-M000").mkdir(parents=True)
```

ghisvail: I wonder whether we really need a BIDS-like layout to test this function, instead of just creating a nifti file with some random name under `tmp_path`?

ghisvail (approved):
Approved with suggestions.

Thanks you for this very welcome piece of refactoring 🙏

test/unittests/utils/test_filemanip.py
NicolasGensollen: I initially thought the same, but you do need a BIDS-like layout because of the call to `get_subject_id` (`image_id = get_subject_id(str(image_path))`), which parses the `image_path` with the underlying assumption that it is a BIDS compliant name.

ghisvail: Then I believe this piece of logic should go elsewhere:

```
       image_id = get_subject_id(str(image_path))
        error_msg = (
            f"Clinica does not handle {dim}D volumes for {image_id.replace('_', ' | ')}"
        )
```

which would make `load_img_3d` more generic (why not `load_volume` actually?).

Imo, there is nothing particulary BIDS specific to this loader apart from the error handling logic, which could be deported upstream (where the BIDS context exists) to reduce coupling.

Also `NotImplementedError` is probably not the exception type we want to raise. It's mostly used for ABCs to signal uncomplete class derivation. Perhaps we should use a good old `ValueError` instead.

NicolasGensollen: I agree with your points. I proposed an implementation in 19a794e
Let me know if this is more or less what you had in mind.

ghisvail: Perfecto 👌

ghisvail (approved):

PR #876, opened 2023-01-27
DOC: Improve getting started section of the readme

ghisvail:
- Promote pipx as the recommended method for install
- Promote conda as means to install Clinica plus dependencies
- Explain why all third-party dependencies cannot be distributed
- Use Python 3.10 for the conda environment (now the default version)


NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #878, opened 2023-01-27
[REF TEST] Refactor and add unit tests for `check_dependency`

NicolasGensollen:
This PR proposes to make a few changes to `clinica/utils/check_dependency.py`:

- Add type hints to functions
- Update existing docstrings to follow Numpydoc conventions
- Add documentations and examples
- Refactor check functions to be partials and not duplicate the checking logic
- Add unit tests
- Remove `check_software_requirements` which was not used

### DIFF
--- a/test/unittests/utils/test_check_dependency.py
+++ b/test/unittests/utils/test_check_dependency.py
@@ -45,12 +45,11 @@ def test_check_software():
         _check_software("foo", binaries=["foo"])
 
     assert _check_software("foo", binaries=["ls"]) is None
-    assert _check_software("foo", binaries=["ls", "cd"]) is None
     with pytest.raises(
         ClinicaMissingDependencyError,
         match="Clinica could not find foo software: the FOO variable is not set.",
     ):
-        _check_software("foo", binaries=["ls", "cd"], env=("FOO", "foo"))
+        _check_software("foo", binaries=["ls"], env=("FOO", "foo"))
 
     os.environ["FOO"] = "foo"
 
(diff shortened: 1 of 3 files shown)


test/unittests/utils/test_check_dependency.py
```diff
@@ -0,0 +1,77 @@
...
+        ClinicaMissingDependencyError,
+        match="The FOO environment variable you gave is not a folder",
+    ):
+        check_environment_variable("FOO", "foo")
+    os.environ["FOO"] = "."
+    assert check_environment_variable("FOO", "foo") == "."
+    os.environ.pop("FOO")
+
+
+def test_check_software():
```
MatthieuJoulot, line 33: It feels like this function could easily be factorized in two functions using pytest. One would test the errors, while the other tests the assertions. WDYT ?

MatthieuJoulot (changes requested):
Just one small change, otherwise this looks good ! 
One minor question: should we do the same thing you're doing for FSL for Freesurfer, and check that the version is recent enough ? it would make sense to have only versions 6.0 and above (since that is what we handle with `freesurfe-atlas`

test/unittests/utils/test_check_dependency.py
NicolasGensollen: I proposed a possible refactoring in cdf8f93
Let me know if this is more or less what you had in mind.

NicolasGensollen:
> One minor question: should we do the same thing you're doing for FSL for Freesurfer, and check that the version is recent enough ? it would make sense to have only versions 6.0 and above (since that is what we handle with `freesurfe-atlas`

I added a version check in the same spirit in cdf8f93. Let me know if this works for you.

MatthieuJoulot (approved):
Looks good :)

PR #879, opened 2023-01-30
[REF TEST] Refactor `converter_utils`

NicolasGensollen:
This PR proposes to update the utilities defined in `clinica.iotools.converter_utils.py`:

- Add type hints to functions
- Update existing docstrings to follow Numpydoc conventions
- Add documentations
- Refactor the `MissingModsTracker` class
      - Simplify constructor
      - Add safety checks in associated methods
- Refactor `print_statistics`, `print_longitudinal_analysis`, and `print_table`
      - Decouples the summary string generation logic from the file writing.
      - Rename to `write_*` since they write to files and don't print
      - Pass `Path` objects instead of file handles which avoids forgetting to close them
- Remove `has_one_index` which was not used
- Remove `increment_dict` and use a `Counter` object instead
- Add unit tests

### DIFF
--- a/clinica/iotools/converter_utils.py
+++ b/clinica/iotools/converter_utils.py
@@ -304,7 +304,7 @@ def compute_table(mod_dict: dict) -> str:
     table : str
         The summary table as a string.
     """
-    diagnoses = sorted(list(set().union(*mod_dict.values())))
+    diagnoses = sorted(set().union(*mod_dict.values()))
     table = "\t" + "\t| ".join(diagnoses)
     table += "\n" + "-" * 8 * (len(diagnoses) + 2) + "\n"
     for mod_key, mod_value in mod_dict.items():
--- a/clinica/iotools/converter_utils.py
+++ b/clinica/iotools/converter_utils.py
@@ -125,7 +125,7 @@ def write_statistics(
     mmt : MissingModsTracker
         Instance of MissingModsTracker.
     """
-    with open(summary_file, "r") as fp:
+    with open(summary_file, "w") as fp:
         fp.write(compute_statistics(num_subjs, ses_aval, mmt))
 
 
--- a/clinica/iotools/converter_utils.py
+++ b/clinica/iotools/converter_utils.py
@@ -1,4 +1,75 @@
-from typing import List
+from os import PathLike
+from typing import List, Optional
+
+
+class MissingModsTracker:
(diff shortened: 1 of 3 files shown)


clinica/iotools/converter_utils.py
```diff
@@ -30,41 +101,79 @@ def sort_session_list(session_list: List[str]) -> List[str]:
...
-        mmt: object MissingModsTracker
+    Parameters
+    ----------
+    summary_file : PathLike
+        Path of the output file where statistics should be written.
+
+    num_subjs : int
+        Number of subjects.
+
+    ses_aval : list of str
```
ghisvail, line 122: ses_aval -> ses_avail

Let's fix the original typo since we are at it :+1:

clinica/iotools/converter_utils.py
```diff
@@ -76,149 +185,131 @@ def print_statistics(summary_file, num_subjs, ses_aval, mmt):
...
+
+    Parameters
+    ----------
+    bids_dir : PathLike
+        Path to the BIDS directory.
+
+    out_dir : PathLike
+        Path to the output directory of the check-missing-modality pipeline.
 
+    ses_aval : list of str
```
ghisvail, line 242: ses_aval -> ses_avail

clinica/iotools/converter_utils.py
```diff
@@ -76,149 +185,131 @@ def print_statistics(summary_file, num_subjs, ses_aval, mmt):
...
-        Returns:
-             The hash map containing the list of missing files.
-        """
-        return self.missing
+    Returns
+    -------
+    table : str
+        The summary table as a string.
+    """
+    diagnoses = sorted(list(set().union(*mod_dict.values())))
```
ghisvail, line 304: `sorted` works on sets directly:

```suggestion
    diagnoses = sorted(set().union(*mod_dict.values()))
```

ghisvail (approved):
Approved with minor suggestions

clinica/iotools/converter_utils.py
NicolasGensollen: Thanks for catching this ! 👍

clinica/iotools/converter_utils.py
NicolasGensollen: Thanks ! 🙏

PR #886, opened 2023-02-27
[TEST] Add non regression test for B0 flirt pipeline

NicolasGensollen:
This PR proposes to add non regression tests specific to the `b0_flirt` pipeline which relies on FSL to perform rigid registration of the B0 dataset onto the first DWI volume.
The pipeline is defined here:

https://github.com/aramis-lab/clinica/blob/0566ca64de22e5c9ec522cdb0268c1c04538c2c7/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py#L334-L338

and used in the `DWIPreprocessingUsingT1` pipeline:

https://github.com/aramis-lab/clinica/blob/5733d8c1c5817fa8dcff5b4ce38b4c06802f20b6/clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_utils.py#L284-L287

The test is relatively fast and runs in a few minutes (usually 3 to 4 minutes).

This PR is linked to the data PR: https://github.com/aramis-lab/clinica_data_ci/pull/24

PR #892, opened 2023-03-15
[TEST] Add non regression test for Eddy FSL pipeline

NicolasGensollen:
Similar to #886 for `b0_flirt`, this PR proposes to add non regression tests for a sub-workflow of the `DWIPreprocessingUsingT1` workflow: the `EddyFSL` workflow.

It also does the following modifications:

- Move the node responsible for masking B0 within the EddyFSL workflow, which makes the inputs of the later easier to understand. It also connects directly to the `b0_flirt` workflow.
- Enable the user to specify a custom working directory for the EddyFSL workflow where output files should be written.

See https://github.com/aramis-lab/clinica_data_ci/pull/26

It seems to run in about 2.5 hours on the Aramis machines.


MatthieuJoulot (commented):
LGTM ! Cool to test such a time taking pipeline. Will test and approve if the test works !

MatthieuJoulot (approved):
The test works ! It's all good :)

PR #893, opened 2023-03-20
[REF] Use `eddy_fsl_pipeline` in `DWIPreprocessingUsingPhasediff`

NicolasGensollen:
The goal of this PR is to simplify the `DWIPreprocessingUsingPhaseDiff` workflow.

This workflow is already written as a sequence of steps separated by comments. Having these steps implemented in separate workflows would make things much easier to test and understand.

Furthermore, the pipeline relies on `Eddy` in various places but doesn't use the `eddy_fsl_pipeline` defined for the `DWIPreprocessingUsingT1` pipeline. 

This PR thus proposes to:

- extract the first step of the pipeline in its own workflow `compute_reference_b0`
- within this extracted workflow, use the `eddy_fsl_pipeline` instead of re-doing everything from scratch
- replace the third step by another call to `eddy_fsl_pipeline` for the same reasons

Note: To do that, I had to slightly modify the `eddy_fsl_pipeline` from the `DWIPreprocessingUsingT1` workflows module. This makes me think that we should have a module for workflows common to multiple DWI pipelines.

This PR adds a functional test that compares the output of the `compute_reference_b0` workflow against reference values. I have computed these reference values with the pipeline implementation defined on `dev` (without the modifications of this PR) and checked that the tests still pass with the modifications done here.

The added functional test needs the data from this PR: https://github.com/aramis-lab/clinica_data_ci/pull/27

### DIFF
--- a/clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
+++ b/clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
@@ -1,3 +1,8 @@
+from typing import Optional
+
+from nipype.pipeline.engine import Workflow
+
+
 def prepare_phasediff_fmap(name="prepare_phasediff_fmap"):
     """This workflow adapts the fsl_prepare_fieldmap script from FSL for the FSL eddy command.
 
@@ -110,10 +115,10 @@ def compute_reference_b0(
     low_bval: float,
     use_cuda: bool,
     initrand: bool,
-    output_dir=None,
-    name="compute_reference_b0",
-):
-    """Step 1 of the DWI preprocessing using phasediff pipeline.
+    output_dir: Optional[str] = None,
+    name: str = "compute_reference_b0",
+) -> Workflow:
+    """Step 1 of the DWI preprocessing using phase diff pipeline.
 
     Compute the reference b0 (i.e. average b0 with EPI distortions)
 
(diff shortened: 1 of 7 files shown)


clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
```diff
@@ -104,3 +104,145 @@ def prepare_phasediff_fmap(name="prepare_phasediff_fmap"):
...
     return wf
+
+
+def compute_reference_b0(
+    low_bval: float,
+    use_cuda: bool,
+    initrand: bool,
+    output_dir=None,
+    name="compute_reference_b0",
+):
```
MatthieuJoulot, line 115: Could you precise what the typing for what the pipeline returns ?

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
```diff
@@ -1,4 +1,11 @@
-def eddy_fsl_pipeline(low_bval, use_cuda, initrand, name="eddy_fsl"):
+def eddy_fsl_pipeline(
+    low_bval,
+    use_cuda,
+    initrand,
+    image_id: bool = False,
+    field: bool = False,
+    name="eddy_fsl",
+):
```
MatthieuJoulot, line 8: while you're at it, should you precise typing, and give a proper docstring to this function. You may as well want to do the same renaming you've started. If I remember correctly, you weren't fond of the name `low_bval`

test/nonregression/pipelines/test_run_pipelines_dwi.py
```diff
@@ -52,6 +52,83 @@ def test_dwi_preprocessing_using_t1(cmdopt, tmp_path):
...
+    available in the JSON file (also available in the input_dir).
+    In the main pipeline, the input node is performing the extraction of these
+    metadata from the JSON file. Here, we perform this extraction directly with
+    the `extract_metadata_from_json` and `bids_dir_to_fsl_dir`functions.
+
+    The compute_reference_b0 workflow produces two outputs that are compared
+    against their reference values:
+        - The reference B0 volume
+        - The brain mask computed on the B0 volume
+    """
```
MatthieuJoulot, line 83: Same comment as before. Could you get this docstring to be numpydocs compliant ?

clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
```diff
@@ -104,3 +104,144 @@ def prepare_phasediff_fmap(name="prepare_phasediff_fmap"):
...
+        - "image_id": Prefix to be used for output files
+
+    It is parametrized by:
+        - low_bval: float, threshold value to determine the B0 volumes in the DWI image
+        - use_cuda: bool, boolean to indicate whether cuda should be used or not
+        - initrand: bool, ???
+        - output_dir: str, path to output directory. If provided, the pipeline will write
+          its output in this folder.
+        - name: str, name of the pipeline
+    """
```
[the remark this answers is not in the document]
NicolasGensollen: Should have been addressed by [48ba105](https://github.com/aramis-lab/clinica/pull/893/commits/48ba105ce3f9b7d71fb4f8965bb8a1b0ca2e8c3a)
Let me know if this answers your concern or not.

clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
NicolasGensollen: Should have been addressed by [48ba105](https://github.com/aramis-lab/clinica/pull/893/commits/48ba105ce3f9b7d71fb4f8965bb8a1b0ca2e8c3a)
We'll need to check that the return value and the Optional type aren't breaking stuff when used through Nipype...

test/nonregression/pipelines/test_run_pipelines_dwi.py
NicolasGensollen: Hmmm, since this is a test there isn't really a standard docstring format.
I have written a lot of things because these tests are a bit different from the usual non-regression tests in clinica which we traditionally write only for full user-facing pipelines.

The sub-workflows I wish to test through these tests don't have the BIDS reading and CAPS writing capabilities of clinica's user facing pipelines. They simply expect a bunch of file paths and parameters as inputs, they do some computations writing files in temporary locations, and make some of them available as outputs.
By adding a writing node with conditional connections, the pipeline can potentially write to the test's `tmp_path`, and the output can then be compared to the reference.

This doesn't seem trivial to me, hence the verbose docstring. Let me know if it doesn't feel clear to you or if you have ideas to clarify things.

MatthieuJoulot: okay, that sounds good, I wasn't sure about the docstring format since it is, as you mentionned,  a test !

clinica/pipelines/dwi_preprocessing_using_t1/dwi_preprocessing_using_t1_workflows.py
NicolasGensollen: Names were changed in [32051ae](https://github.com/aramis-lab/clinica/pull/893/commits/32051ae4af21f3031ce13ab7ac6593ea5885a62e)

MatthieuJoulot (approved):
LGTM :)

NicolasGensollen:
Merging...

PR #897, opened 2023-03-27
[REF] Extract Step 2 of `DWIPreprocessingUsingPhasediff` into its own workflow

NicolasGensollen:
Similar to #893 for Step 1, this PR proposes to extract the Step 2 of the `DWIPreprocessingUsingPhaseDiff` pipeline into its own workflow: `calibrate_and_register_fmap`
This PR also proposes to add functional tests for the extracted workflow in isolation. 
The added test is classified as "fast" as it is running in about 1 minute. It requires the data I have generated here: https://github.com/aramis-lab/clinica_data_ci/pull/29

### DIFF
--- a/clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
+++ b/clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
@@ -501,16 +501,16 @@ def calibrate_and_register_fmap(
             [("out_file", "in_file")],
         ),
         # Output connections
-        (smooth_calibrated_fmap, outputnode, ["out_file", "smooth_calibrated_fmap"]),
+        (smooth_calibrated_fmap, outputnode, [("out_file", "smooth_calibrated_fmap")]),
         (
             register_bet_magnitude_fmap_onto_b0,
             outputnode,
-            ["out_file", "bet_magnitude_fmap_registered_onto_b0"],
+            [("out_file", "bet_magnitude_fmap_registered_onto_b0")],
         ),
         (
             apply_xfm_on_calibrated_fmap,
             outputnode,
-            ["out_file", "registered_calibrated_fmap"],
+            [("out_file", "registered_calibrated_fmap")],
         ),
     ]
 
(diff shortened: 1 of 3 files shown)


NicolasGensollen:
The Jenkins failure is expected as the data needed for the tests aren't merged and uploaded yet.

clinica/pipelines/dwi_preprocessing_using_fmap/dwi_preprocessing_using_phasediff_fmap_workflows.py
```diff
@@ -275,3 +275,222 @@ def compute_reference_b0(
...
     return wf
+
+
+def calibrate_and_register_fmap(
+    output_dir: Optional[str] = None,
+    name: str = "calibrate_and_register_fmap",
+) -> Workflow:
+    """Step 2 of the DWI preprocessing using phase diff pipeline.
+
+    Calibrate and register the field map.
```
MatthieuJoulot, line 286:
```suggestion
    Calibrates and registers the field map.
```

MatthieuJoulot (changes requested):
Looks good. Just one small suggestion.
I'll test it asap with the data you've generated.

test/nonregression/pipelines/test_run_pipelines_dwi.py
```diff
@@ -175,6 +175,60 @@ def test_prepare_phasediff_fmap(cmdopt, tmp_path):
...
+        [
+            "sub-01_ses-M000_phasediff_rads_unwrapped_radsec_fieldmap_demean_maths_flirt_smooth.nii.gz",
+            "sub-01_ses-M000_magnitude1_corrected_brain_flirt.nii.gz",
+            "sub-01_ses-M000_phasediff_rads_unwrapped_radsec_fieldmap_demean_maths_flirt.nii.gz",
+        ],
+    ):
+        out_file = fspath(tmp_path / "tmp" / folder / filename)
+        ref_file = fspath(ref_dir / folder / filename)
+
+        assert similarity_measure(out_file, ref_file, 0.99)
```
MatthieuJoulot, line 229:
```suggestion
        assert similarity_measure(out_file, ref_file, 0.98)
```
That way we account for the difference in precision between linux and mac. 0.99 doesn't get through mac

MatthieuJoulot (changes requested):

NicolasGensollen:
Tests are passing with the updated data.
Let me know if you think there are still things to change before we merge @MatthieuJoulot !

NicolasGensollen:
Thanks for the reviews @MatthieuJoulot !
Merging...

PR #885, opened 2023-02-27
Pydra implementation of  statistics-volume

MatthieuJoulot:
This PR proposes a Pydra-based implementation of the pipeline statistics-volume.

WIP !

PR #905, opened 2023-04-06
[TEST] run MATLAB based tests in same group

NicolasGensollen:
Follow-up PR of #885 

This PR proposes to run the two statistics tests relying on SPM within the same group to avoid race conditions when running the test suite in parallel.


NicolasGensollen:
This seems to have solved the issue.
Merging to fix CI on dev...

PR #907, opened 2023-04-12
[TEST] Add unit tests for StatisticsVolume utility functions

NicolasGensollen:
Following the refactoring made in #885, this PR proposes to add unit tests for the utility functions of pipeline `StatisticsVolume`.

This requires `pytest-mock` which is added to the dependencies in #883

PR #908, opened 2023-04-12
[INFRA] add `pytest-mock` and `pytest-random-order` to dependencies

NicolasGensollen:
This PR proposes to add `pytest-mock` and `pytest-random-order` to the testing dependencies.
The former is used in #883 and #907 and I'd like to randomize the execution order of unit tests to avoid test pollution (to be done in another PR).


MatthieuJoulot (approved):
All good !

PR #914, opened 2023-05-15
[ADNI] Fix viscode error

MatthieuJoulot:
This PR fixes a viscode error occuring with the new clinical data.

Issue #913, opened 2023-05-15
ValueError: invalid literal for int() with base 10: 'ns1'


mselimata:
**Describe the bug**
Running adni-to-bids pipeline on T1w MRI data of ADNI and it is not resulting the relevant tsv files due to the value error
**To Reproduce**
1. In miniconda python 3.10 environment running clinica convert adni-to-bids /truba/repo/ALLADNI/MRI/ADNI/ /truba/home/meatay/clinical_adni/ /truba/home/meatay/ADNI_BIDS/
2. It creates the nifti folder structure
3. General tsv files like patients and modalities in the general bids folder but not the specific tsv files corresponding to the subjects which has the information about each sessions
4. See error, this is a huge log with many file pathways and the error is in the end of the log file which is easy to omit until the realization of missing subject tsv files.

[32m2023-05-15 02:56:39,553:INFO:Creating modality agnostic files...[0m
[32m2023-05-15 02:56:39,843:INFO:Creating participants.tsv...[0m
/truba/home/meatay/miniconda3/envs/clinica/lib/python3.10/site-packages/clinica/iotools/bids_utils.py:114: DtypeWarning: Columns (19,20,21,104,105,106) have mixed types. Specify dtype option on import or set low_memory=False.
  file_to_read = pd.read_csv(file_to_read_path)
/truba/home/meatay/miniconda3/envs/clinica/lib/python3.10/site-packages/clinica/iotools/bids_utils.py:114: DtypeWarning: Columns (19,20,21,104,105,106) have mixed types. Specify dtype option on import or set low_memory=False.
  file_to_read = pd.read_csv(file_to_read_path)
[32m2023-05-15 02:56:44,174:INFO:The following subjects of dataset directory were not found in your BIDS folder :
070S4719, 098S4095, 003S6158, 005S6393, 153S6633, 153S6694, 114S6758, 016S6799, 012S6837, 305S6879, 021S6940, 027S6961, 021S6978, 341S6960, 341S6981, 024S7025, 341S7018, 037S6991, 035S7031, 035S7019, 037S7012, 011S7089, 053S7086[0m
/truba/home/meatay/miniconda3/envs/clinica/lib/python3.10/site-packages/clinica/iotools/converters/adni_to_bids/adni_utils.py:446: PerformanceWarning: indexing past lexsort depth may impact performance.
  dxsum_df.loc[(alternative_id, "sc"), "DIAGNOSIS"].values[0]
[...]

PR #915, opened 2023-05-15
[ENH] Improve function `viscode_to_session`

NicolasGensollen:
Related to issue #913 and follow-up PR of #914 

This PR proposes some improvements to the function `viscode_to_session` such as to avoid cryptic errors due to invalid conversions to integer.

More precisely:

- Add the possibility to pass custom baseline identifiers which automatically return "ses-M000"
- Improve docs and add examples
- Use regex in order to make sure the conversion to integer will work
- If the input doesn't match the pattern, raise a meaningful error
- Add unit tests for the added functionalities

### DIFF
--- a/clinica/iotools/converter_utils.py
+++ b/clinica/iotools/converter_utils.py
@@ -315,20 +315,61 @@ def compute_table(mod_dict: dict) -> str:
     return table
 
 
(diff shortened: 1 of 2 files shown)


MatthieuJoulot (approved):
LGTM !

NicolasGensollen:

PR #916, opened 2023-05-15
Add random seed for ANTs

14thibea:
Hello all :)

For a project using ANTs we would need to have a deterministic setup. This is the aim of this PR in which a value can be given for the random seed of ANTs, thus leading to the same result (using the same computer).
Before merging it would be worth testing if the results are the same on different computers as well. Also this could be checked in the CI.

Best,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,6 +191,6 @@ Temporary Items
 .apdisk
 
 # Clinica specific files
-
+clinica/resources/masks/*
 crash*
 test/data
--- a/clinica/pipelines/t1_linear/anat_linear_pipeline.py
+++ b/clinica/pipelines/t1_linear/anat_linear_pipeline.py
@@ -222,9 +222,6 @@ class AnatLinear(cpe.Pipeline):
             ]
         )
 
-
-
-
         if not (self.parameters.get("uncropped_image")):
             self.connect(
                 [
@@ -273,10 +270,9 @@ class AnatLinear(cpe.Pipeline):
         ants_registration_node.inputs.fixed_image = self.ref_template
         ants_registration_node.inputs.transform_type = "a"
         ants_registration_node.inputs.dimension = 3
-        if self.parameters.get("random_seed") is not None:
-            ants_registration_node.inputs.random_seed = self.parameters.get(
-                "random_seed"
-            )
+
+        if random_seed := self.parameters.get("random_seed", None):
+            ants_registration_node.inputs.random_seed = random_seed
 
         # 3. Crop image (using nifti). It uses custom interface, from utils file
 
--- a/clinica/pipelines/t1_linear/anat_linear_pipeline.py
+++ b/clinica/pipelines/t1_linear/anat_linear_pipeline.py
@@ -221,10 +221,10 @@ class AnatLinear(cpe.Pipeline):
                 (self.output_node, write_node, [("affine_mat", "@affine_mat")]),
             ]
         )
-    
-                    
-            
-        
+
+
+
+
         if not (self.parameters.get("uncropped_image")):
             self.connect(
                 [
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -191,6 +191,6 @@ Temporary Items
 .apdisk
 
 # Clinica specific files
-
+clinica/resources/masks/*
```
ghisvail, line 194: Do you remember what generates these?

14thibea: Yes I installed clinica locally with `pip install -e .` (it wasn't working with poetry)
And then when I ran the pipeline with my installation the masks were downloaded there.

ghisvail: Ok, you did well to exclude them here then 👍

clinica/pipelines/t1_linear/anat_linear_pipeline.py
```diff
@@ -273,6 +273,10 @@ def build_core_nodes(self):
         ants_registration_node.inputs.fixed_image = self.ref_template
         ants_registration_node.inputs.transform_type = "a"
         ants_registration_node.inputs.dimension = 3
+        if self.parameters.get("random_seed") is not None:
+            ants_registration_node.inputs.random_seed = self.parameters.get(
+                "random_seed"
+            )
```
ghisvail, line 279: For our own records:

- This option was made available from version 2.3.0 onwards. It's probably old enough (June 2018) as of today.

- This option prevents execution with multiple threads. We are mostly interested in patient-level parallel processing so I don't think that's an issue.

docs/Pipelines/T1_Linear.md
```diff
@@ -43,6 +43,9 @@ On default, cropped images (matrix size 169×208×179, 1 mm isotropic voxels) ar
 generated to reduce the computing power required when training deep learning models.
 Use the option `--uncropped_image` if you do not want to crop the image.
 
+It is also possible to obtain a deterministic result by setting the value of the random
+seed used by ANTs with the option `--random_seed`. Default will lead to a non-determinstic result.
```
ghisvail, line 47: Just to be sure 😅 

```suggestion
It is also possible to obtain a deterministic result by setting the value of the random
seed used by ANTs with the option `--random_seed`. Default will lead to a non-determinstic result. This option requires ANTs version `2.3.0` onwards.
```

14thibea: No idea, but it won't hurt to put it anyway (even if it is possible with a previous version)

ghisvail:
@14thibea shall I merge this as is, or are there other things you need to test? Your PR is still at draft stage atm.

14thibea:
Hi @ghisvail 
I am going to run t1-linear in another machine (Linux) to check I have the same result!
If it works I will try to put this on the CI as well :)

sophieloiz:
Hi @ghisvail, from what I understand, this PR addresses the random seed issue just for the t1-linear pipeline. Would it be possible to also add this seed for the flair-linear pipeline ?

ghisvail:
> Hi @ghisvail, from what I understand, this PR addresses the random seed issue just for the t1-linear pipeline.

Actually, both `t1-linear` and `flair-linear` use the same `anat-linear` pipeline underneath. So `flair-linear` will also benefit from the pieces of improvement proposed by @14thibea 👍

sophieloiz:
Perfect, thanks ! Do we need to modify the` flair_linear_cli.py` file to add the `--random_seed parameter ` ?

ghisvail (approved):

ghisvail:
Just a quick heads-up:

- Following Friday's meeting, I added random seed support to `pet-linear`
- I took the liberty to do some refactoring in the process, i.e.:
  - Move the random seed option to `cli_param` as it is now used by more than one pipeline
  - Use assignment expressions since we are on Python 3.8+ and they simplify the code

PR #919, opened 2023-05-17
[PYDRA] Machine learning spatial svm pydra pipeline

MatthieuJoulot:


clinica/pydra/machine_learning_spatial_svm/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+from clinica.pydra.engine import clinica_io
+
+
+def _build_query(parameters: dict) -> dict:
+    input_name = parameters["orig_input_data"].replace("-", "_")
+    if input_name == "pet_volume":
+        query = {}
+    elif input_name == "t1_volume":
+        query = {}
+    return query
```
NicolasGensollen, line 13: This function isn't doing anything atm. Could you move the query building logic in it ?

clinica/pydra/machine_learning_spatial_svm/spatial_svm_cli.py
```diff
@@ -0,0 +1,113 @@
...
+        output_dir=caps_directory,
+        parameters=parameters,
+        # caps_directory=caps_directory,
+        # tsv_file=subjects_sessions_tsv,
+        # base_dir=working_directory,
+        # parameters=parameters,
+        # name=pipeline_name,
+    )
+    pydra_utils.run(pipeline, n_procs)
+    # exec_pipeline = (
```
NicolasGensollen, line 100: to be removed ?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
I made a very quick first pass. Looks mostly good to me.
Could you also add this to the instantiation tests ?

https://github.com/aramis-lab/clinica/blob/45f50bfbdfd6403f56ebd6be98c7aa272d234ad0/test/instantiation/test_cli.py#L104-L109

clinica/pydra/machine_learning_spatial_svm/pipeline.py
MatthieuJoulot: Yes. I thought i would need to build a more complex query.

clinica/pydra/machine_learning_spatial_svm/pipeline.py
```diff
@@ -0,0 +1,122 @@
...
+    Returns
+    -------
+    wf : Workflow
+        The core
+    """
+    from typing import Any
+
+    import clinica.pydra.machine_learning_spatial_svm.tasks as utils
+
+    # query = _build_query(parameters)
```
NicolasGensollen, line 30: I think the idea of using `_build_query` was good. It will make this function a bit easier to read.

I think we can implement it in the same way as:

https://github.com/aramis-lab/clinica/blob/a08d1c830c539ab9ec04a225a3907a8f5da1d044/clinica/pydra/statistics_volume/pipeline.py#L46-L69

clinica/pydra/machine_learning_spatial_svm/spatial_svm_cli.py
```diff
@@ -0,0 +1,98 @@
+from typing import Optional
+
+import click
+
+import clinica.pydra.engine_utils as pydra_utils
+import clinica.pydra.machine_learning_spatial_svm.pipeline as pydra_machine_learning_spatial_svm
+from clinica.pipelines import cli_param
+from clinica.pipelines.engine import clinica_pipeline
+
+pipeline_name = "pydra-machinelearning-prepare-spatial-svm"
```
NicolasGensollen, line 10: Small thing, but given the structure of this name, I would also expect a "-" between "machine" and "learning".

NicolasGensollen (commented):
Just a small suggestion about having the query building logic in a separate function. Otherwise LGTM.

clinica/pydra/machine_learning_spatial_svm/pipeline.py
```diff
@@ -0,0 +1,111 @@
+import pydra
+from pydra import Workflow
+
+from clinica.pydra.engine import clinica_io
+
+
+def _build_query(parameters: dict) -> dict:
+    input_name = parameters["orig_input_data_ml"].replace("-", "_")
```
NicolasGensollen, line 8: We don't need to do that since we're only comparing to raw strings.

```suggestion
```

clinica/pydra/machine_learning_spatial_svm/pipeline.py
```diff
@@ -0,0 +1,111 @@
...
+    if input_name == "t1_volume":
+        query = {
```
NicolasGensollen, line 10:
```suggestion
        return {
```

clinica/pydra/machine_learning_spatial_svm/pipeline.py
```diff
@@ -0,0 +1,111 @@
...
+            "acq_label": parameters["acq_label"],
+            "suvr_reference_region": parameters["suvr_reference_region"],
+            "use_brainmasked_image": False,
+            "use_pvc_data": parameters["use_pvc_data"],
+            "fwhm": 0,
+            "group_label": parameters["group_label"],
+        }
+
+    return query
```
NicolasGensollen, line 25:
```suggestion
```

clinica/pydra/machine_learning_spatial_svm/spatial_svm_cli.py
```diff
@@ -0,0 +1,98 @@
...
+       GROUP_LABEL is an user-defined identifier to target a specific group of subjects.
+
+       The third positional argument can be t1-volume to use tissue maps or pet-volume to use standardized uptake value ratio (SUVR) maps.
+
+    Prerequisites: You need to execute the t1-volume pipeline to run the pipeline on T1-weighted MRI data, and the t1-volume + pet-volume pipelines to apply it to PET data.
+
+    See https://aramislab.paris.inria.fr/clinica/docs/public/latest/Pipelines/MachineLearning_PrepareSVM/"
+    """
+
+    from clinica.utils.exceptions import ClinicaException
```
NicolasGensollen, line 62: I just realized that we should adopt the same kind of logic as the one we used for `statistics_volume`. That is, move all the parameter checking logic in a `_check_pipeline_parameters` function 

https://github.com/aramis-lab/clinica/blob/a08d1c830c539ab9ec04a225a3907a8f5da1d044/clinica/pydra/statistics_volume/pipeline.py#L9-L21

which is then called at the beginning of `build_core_workflow`:

https://github.com/aramis-lab/clinica/blob/a08d1c830c539ab9ec04a225a3907a8f5da1d044/clinica/pydra/statistics_volume/pipeline.py#L102

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Just a few things I forgot to mention in my previous review.

clinica/pydra/machine_learning_spatial_svm/spatial_svm_cli.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if orig_input_data_ml == "pet-volume":
```
NicolasGensollen, line 64: I would also move this in `_check_pipeline_parameters` for the following reasons:

- the code is verifying the consistency of the parameters, so it makes sense to have it inside this function
- atm these checks won't be executed when calling the pipeline without using the CLI


WDYT ?

MatthieuJoulot: yes that makes sense !

MatthieuJoulot: it is actually what we check in `_check_pipeline_parameters`, so i'll be removing this.

NicolasGensollen: Indeed ! Thanks !

PR #920, opened 2023-05-17
[ADNI] Fix uns1 issue

MatthieuJoulot:
Added some more tests, since it seems like the `uns1` problem is general.


NicolasGensollen (approved):
As discussed, let's merge this and make sure it fixes the converter issues before doing a follow-up refactoring of this function to improve code quality.

PR #921, opened 2023-05-17
[TEST] Add missing tests

MatthieuJoulot:
Adding some missing instanciation tests. I do not know if `pydra-freesurfer`works @ghisvail ?


NicolasGensollen (approved):
Great ! Thanks for spotting these @MatthieuJoulot ! 👍

ghisvail (approved):

ghisvail:
> I do not know if pydra-freesurferworks @ghisvail ?

It should. Which interfaces are you planning to use?

PR #923, opened 2023-05-22
[GENFI] Patch conversion issue

MatthieuJoulot:
This PR adds try/except to handle cases were dcm2niix conversion is peculiar, which we have found when converting the whole dataset: a few DWIs are not completely converted, only the imaginary or real part. Since the converter looks for a file with the same name as the name that is supposed to be given as an output by dcm2niix, it does not work because dcm2niix adds a suffix like `_real` or` _imaginary` in these cases.

- [x] add try/except
- [x] add function to delete `_real` and `_imaginary` files


clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,21 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
...
+
+def delete_real_imaginary_files(to: PathLike) -> None:
+    """This function scans the BIDS after it has been written to delete the `_real` and `_imaginary` files.
+
+    Parameters
+    ----------
+    to: PathLike
+        Path to the BIDS
+    """
+
```
NicolasGensollen, line 924: Suggestion: slightly shorter and without os:

``` python
bids_folder = Path(bids_folder)
for file_type in ("real", "imaginary"):
    for f in bids_folder.rglob(f"*_{file_type}.*"):
        f.unlink() 
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,21 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
         raise ValueError(f"{run_num} is outside of the scope.")
     if nb_parts == PhilipsNumberOfParts.OTHER:
         return None
+
+
+def delete_real_imaginary_files(to: PathLike) -> None:
```
NicolasGensollen, line 916:
```suggestion
def delete_real_and_imaginary_files(bids_folder: PathLike) -> None:
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,21 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
...
+    """This function scans the BIDS after it has been written to delete the `_real` and `_imaginary` files.
```
NicolasGensollen, line 917:
```suggestion
    """Delete all files with a `_real` or `_imaginary` suffix from the BIDS folder.
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,21 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
...
+
+    Parameters
+    ----------
+    to: PathLike
+        Path to the BIDS
+    """
+
+    import os
+
```
NicolasGensollen, line 926: Could be nice to add a logging message (INFO ?) for each deleted file (or one for all at once??).
As a user, I would expect the converter to tell me something when deleting files from my system.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -786,6 +786,7 @@ def write_bids(
                 metadata.run_num,
             )
     correct_fieldmaps_name(to)
+    delete_real_imaginary_files(to)
```
NicolasGensollen, line 789: What is the consequence of deleting these files ? Is the output still valid ? BIDS-compliant ?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Some questions on the context, and some small suggestions on the code.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
MatthieuJoulot: This makes the ouput BIDS compliant, as `sub-x_ses-y_dwi_real.nii.gz is not.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -856,14 +857,19 @@ def merge_philips_diffusion(
...
-            data["MultipartID"] = multipart_id
-            json.dump(data, f, indent=4)
+    try:
+        with open(json_file, "r+") as f:
+            if multipart_id is not None:
+                data = json.load(f)
+                data["MultipartID"] = multipart_id
+                json.dump(data, f, indent=4)
+    except:
+        cprint(msg=f"the file {json_file} does not exist.", lvl="warning")
```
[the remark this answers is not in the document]
MatthieuJoulot: This case is, to my knowledge, only triggered by incomplete conversion of dwi from philips which get the `_real` or `_imaginary` extension. so these outputs should be removed afterwards

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,20 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
...
         return None
+
+
+def delete_real_and_imaginary_files(bids_folder: PathLike) -> None:
+    """Delete all files with a `_real` or `_imaginary` suffix from the BIDS folder.
+
+    Parameters
+    ----------
+    to: PathLike
+        Path to the BIDS
```
NicolasGensollen, line 922:
```suggestion
        Path to the BIDS folder.
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -905,3 +911,20 @@ def _get_multipart_id(nb_parts: PhilipsNumberOfParts, run_num: str) -> Optional[
         raise ValueError(f"{run_num} is outside of the scope.")
     if nb_parts == PhilipsNumberOfParts.OTHER:
         return None
+
+
+def delete_real_and_imaginary_files(bids_folder: PathLike) -> None:
+    """Delete all files with a `_real` or `_imaginary` suffix from the BIDS folder.
+
```
NicolasGensollen, line 918: Could you add a little bit more context here ? 
For example : Where do these files come from ? Why are we removing them ? ...
I think you can use some of the sentences you wrote in the PR message.

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
A few small things but LGTM otherwise.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
NicolasGensollen: Thanks. What I'm not sure to understand is: 

- Is there also a `sub-x_ses-y_dwi.nii.gz`  (in addition to its `real` and/or `imaginary` versions) ?
- If not, this means we are removing all dwi images for these cases. Is this the best strategy ?

MatthieuJoulot: in this specific case there is no `sub-x_ses-y_run-w_dwi`. Because, yes, since this is `Philips`, it would only be a run amongst others. That makes the other runs unusable since you need all of them. and it would break BIDS compliance. 
For these reasons, i think it is the best strategy !

NicolasGensollen (approved):

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
NicolasGensollen: Thanks for the explanations @MatthieuJoulot !

PR #926, opened 2023-05-24
[REF, TEST] Add tests and refactor ADNI `get_preferred_visit_name`

NicolasGensollen:
This PR proposes to:

- Make `get_preferred_visit_name` "private" (becomes `_get_preferred_visit_name`) since it is only used in this module.
- Refactor the code to make it easier to understand (the complex imbrication of `if/elif/else` wasn't easy to follow).
- Raise meaningful errors when unexpected values are received.
- Add unit tests.

There are a few weird things that I discovered when writing the unit tests prior to refactoring. I kept them for backward compatibility concerns but it might be worth taking a closer look....

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -1,6 +1,31 @@
+from enum import Enum, auto
 from typing import Optional, Union
 
 import pandas as pd
+
+
+class ADNIStudy(Enum):
+    """Possible versions of ADNI studies."""
+
+    ADNI1 = auto()
+    ADNI2 = auto()
+    ADNI3 = auto()
+    ADNIGO = auto()
+
+    @classmethod
+    def from_string(cls, study_name: str):
+        if study_name == "ADNI1":
+            return cls.ADNI1
+        if study_name == "ADNI2":
+            return cls.ADNI2
+        if study_name == "ADNI3":
+            return cls.ADNI3
+        if study_name == "ADNIGO":
+            return cls.ADNIGO
+        raise ValueError(
+            f"Invalid study name for ADNI: {study_name}. "
+            f"Possible values are {list(ADNIStudy)}"
+        )
 
 
 def visits_to_timepoints(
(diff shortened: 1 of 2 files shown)


test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -27,6 +27,72 @@ def test_bids_id_to_loni(input_value, expected):
...
+
+
+@pytest.mark.parametrize(
+    "study_name,visit_code,expected",
+    [
+        ("ADNI3", "bl", "ADNI Screening"),
+        ("ADNI3", "M000", "ADNI3 Year 0.0 Visit"),
+        ("ADNI3", "M006", "ADNI3 Year 0.5 Visit"),
+        ("ADNI3", "M012", "ADNI3 Year 1.0 Visit"),
+        ("ADNI3", "M013", "ADNI3 Year 1.0833333333333333 Visit"),
```
NicolasGensollen, line 47: For example, this is a weird behavior. I would expect integer values for years...
I suspect the casting to integer wasn't correctly done in the first place (might be a parenthesis order issue??).

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -76,46 +101,81 @@ def visits_to_timepoints(
...
+    if visit_code == "m06":
+        return "ADNI2 Month 6-New Pt"
+    return f"ADNI2 Year {_parse_year_from_visit_code(visit_code)} Visit"
+
+
+def _get_preferred_visit_name_adni1(visit_code: str) -> str:
+    if visit_code == "bl":
+        return "ADNI Screening"
+    if visit_code == "m03":
+        return "ADNIGO Month 3 MRI"  # does not make any sense...
```
NicolasGensollen, line 170: This is also a bit weird. I think that this relies on the fact that 'm03' is never found for ADNI1 ??

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -37,8 +62,8 @@ def visits_to_timepoints(
     # We try to obtain the corresponding image Visit for a given VISCODE
     for adni_row in adnimerge_subj.iterrows():
         visit = adni_row[1]
-
-        preferred_visit_name = get_preferred_visit_name(visit)
+        study = ADNIStudy.from_string(visit.ORIGPROT)
```
MatthieuJoulot, line 65: where does this ORIGPROT come from and what is it ? It doesn't look like it was there before

MatthieuJoulot: nevermind I found it

MatthieuJoulot (approved):
LGTM

clinica/iotools/converters/adni_to_bids/adni_utils.py
NicolasGensollen: I agree with you it's a terrible name but I'm not the one to blame... ;)

NicolasGensollen:

PR #927, opened 2023-05-25
[GENFI] TSV improvements

MatthieuJoulot:
This PR aims at two things:

- [x] Correcting current issues of the `participants.tsv` and `sessions.tsv`, mainly ensuring we don't have useless information.
- [ ] Adding a `scans.tsv`, which comes in handy since it gives details about `modalities` and `runs`, as well as an information that has been required, the `path_to_the_original_dicom`.

This work will need a data CI PR, since it adds new files (`scans.tsv`) to the architecture, and should therefore break current tests.

### DIFF
--- a/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
+++ b/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
@@ -746,6 +746,14 @@ def write_bids(
     to = Path(to)
     fs = LocalFileSystem(auto_mkdir=True)
     # Ensure BIDS hierarchy is written first.
+    print("participants: ", participants)
+    participants = (
+        participants.reset_index()
+        .drop(["session_id", "modality", "run_num", "bids_filename", "source"], axis=1)
+        .drop_duplicates()
+        .set_index("participant_id")
+    )
+    print("participants: ", participants)
     with fs.transaction:
         with fs.open(
             str(to / "dataset_description.json"), "w"
@@ -756,7 +764,7 @@ def write_bids(
 
     for participant_id, data_frame in sessions.groupby(["participant_id"]):
         sessions = data_frame.droplevel(
-            ["participant_id", "modality", "bids_filename"]
+            ["participant_id", "modality", "bids_filename", "run_num"]
         ).drop_duplicates()
 
         sessions_filepath = to / str(participant_id) / f"{participant_id}_sessions.tsv"
@@ -775,6 +783,16 @@ def write_bids(
             metadata["bids_filename"],
             True,
         )
+        if dcm2niix_success:
+            scans_filepath = (
+                to
+                / str(metadata.participant_id)
+                / str(metadata.session_id)
+                / f"{metadata.participant_id}_{metadata.session_id}_scan.tsv"
+            )
+            print(scans_filepath)
+            with fs.open(str(scans_filepath), "w") as scans_file:
+                write_to_tsv(scans, scans_file)
         if (
             "dwi" in metadata["bids_filename"]
             and "Philips" in metadata.manufacturer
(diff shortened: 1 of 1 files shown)


clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -775,6 +783,8 @@ def write_bids(
             metadata["bids_filename"],
             True,
         )
+        dcm2niix_bids_path.append(bids_full_path)
```
NicolasGensollen, line 786: Maybe something along those lines to write rows within the loop only if the `dcm2niix` conversion was successful?

```python
if dcm2niix_success:
    row_to_write = metadata.to_csv(
        index=False, header=False, lineterminator="\t",
    ).rstrip("\t")
    with open(scans_filepath, "a") as scans_file:
         scans_file.write(row_to_write)
    if "dwi" in metadata.bids_filename and "Philips" in metadata.manufacturer:
        merge_philips_diffusion(...)
```

MatthieuJoulot:
The data for this PR is available [here](https://github.com/aramis-lab/clinica_data_ci/pull/43)

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -775,16 +791,40 @@ def write_bids(
...
-                metadata.number_of_parts,
-                metadata.run_num,
+        if dcm2niix_success:
+            scans_filepath = (
+                to
+                / str(metadata.participant_id)
+                / str(metadata.session_id)
+                / f"{metadata.participant_id}_{metadata.session_id}_scan.tsv"
+            )
+            if not Path(scans_filepath).exists():
```
NicolasGensollen, line 801: `scans_filepath` should already be a `Path` object :

```suggestion
            if not scans_filepath.exists():
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -775,16 +791,40 @@ def write_bids(
...
+                index_to_write = scans_columns_names.to_csv(
+                    index=False, header=False, lineterminator="\t"
+                )
```
NicolasGensollen, line 804: I'm not sure to understand why you need to hardcode the column names and do all these steps. You basically want to write the column names (i.e. the header) only once, when creating the file. So something like this should be enough, right ?

```python
if dcm2niix_success:
    scans_filepath = ...
    row_to_write = metadata.drop(["participant_id", "session_id"]).to_csv(
        index=False,
        header=not scans_filepath.exists(),
        lineterminator="\t",
    ).rstrip("\t")
with open(scans_filepath, "a") as scans_file:
    scans_file.write(f"{row_to_write}\n")
```

NicolasGensollen: Hmmm, indeed it's not working as expected. Sorry about that, writing code in chat boxes without testing it is not without danger...

I had a quick look and this is the most readable way to do what you want that I could find (should work this time but to be tested in your settings...):

```python
if dcm2niix_success:
    scans_filepath = ...
    row_to_write = _serialize_row(
        metadata.drop(["participant_id", "session_id"]),
        write_column_names=not scan_filepath.exists(),
    )
with open(scans_filepath, "a") as scans_file:
    scans_file.write(f"{row_to_write}\n")


def _serialize_row(row: pd.Series, write_column_names: bool) -> str:
    row_dict = row.to_dict()
    to_write = [row_dict.keys(), row_dict.values()] if write_column_names else [row_dict.values()]
    return "\n".join([serialize_list(list(_)) for _ in to_write])


def _serialize_list(data: list, sep="\t") -> str:
    return sep.join([str(value) for value in data])
```

MatthieuJoulot: That works and it is definitely a whole lot better than what i've done

NicolasGensollen (approved):
Let's wait for the CI to finish and I'll merge if everything's green.

PR #930, opened 2023-05-31
[GENFI] Handle T1 converted as T2

MatthieuJoulot:
Some T1 are not correctly converted. This is due to the fact that the conversion is based on the series description and that for a large amount of T1, it is wrongly described as a T2 (not the other way around, as far as we know). `dcm2niix` makes the same mistake as well.
Therefore, we if the `SeriesDescription` is T1 or T2, we check the `EchoTime` and `RepetitionTime` that characterise T1 or T2.

This PR will need a change in the data CI because of that.

### DIFF
--- a/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
+++ b/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
@@ -61,13 +61,6 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     ]
 
     df = df.drop_duplicates(subset=["source"])
-    # df = df.assign(
-    #     series_desc=lambda x: x.source_path.apply(
-    #         lambda y: pdcm.dcmread(y).SeriesDescription
-    #     ),
-    #     acq_date=lambda x: x.source_path.apply(lambda y: pdcm.dcmread(y).StudyDate),
-    #     manufacturer=lambda x: x.source_path.apply(lambda y: _handle_manufacturer(y)),
-    # )
     df = df.assign(
         series_desc=lambda x: x.source_path.apply(
             lambda y: _handle_series_description(y)
--- a/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
+++ b/clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
@@ -61,9 +61,16 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     ]
 
     df = df.drop_duplicates(subset=["source"])
+    # df = df.assign(
+    #     series_desc=lambda x: x.source_path.apply(
+    #         lambda y: pdcm.dcmread(y).SeriesDescription
+    #     ),
+    #     acq_date=lambda x: x.source_path.apply(lambda y: pdcm.dcmread(y).StudyDate),
+    #     manufacturer=lambda x: x.source_path.apply(lambda y: _handle_manufacturer(y)),
+    # )
     df = df.assign(
         series_desc=lambda x: x.source_path.apply(
-            lambda y: pdcm.dcmread(y).SeriesDescription
+            lambda y: _handle_series_description(y)
         ),
         acq_date=lambda x: x.source_path.apply(lambda y: pdcm.dcmread(y).StudyDate),
         manufacturer=lambda x: x.source_path.apply(lambda y: _handle_manufacturer(y)),
@@ -73,6 +80,20 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     for file_mod in to_filter:
         df = df[~df["series_desc"].str.contains(file_mod, case=False)]
     return df
+
+
+def _handle_series_description(x: str) -> str:
+    series_description = pdcm.dcmread(x).SeriesDescription
+    if any([modality in series_description for modality in ["t1", "t2", "T1", "T2"]]):
+        echo_time = pdcm.dcmread(x).EchoTime
+        repetition_time = pdcm.dcmread(x).RepetitionTime
+        if (
+            any([modality in series_description for modality in ["t2", "T2"]])
+            and echo_time < 60
+            and repetition_time < 2100
+        ):
+            return "t1"
+    return series_description
 
 
 def _handle_manufacturer(x: str) -> str:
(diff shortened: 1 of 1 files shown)


MatthieuJoulot:
The values 60 and 2100 have been arbitrarily chosen based on the following document and on the values found in the sequences.
Image Pasted at 2023-5-30 15-20

MatthieuJoulot:
The data CI PR is available [here](https://github.com/aramis-lab/clinica_data_ci/pull/45)

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -61,9 +61,16 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     ]
 
     df = df.drop_duplicates(subset=["source"])
+    # df = df.assign(
+    #     series_desc=lambda x: x.source_path.apply(
```
NicolasGensollen, line 65: Should this be removed ?

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -775,21 +802,47 @@ def write_bids(
...
-            and "Philips" in metadata.manufacturer
-            and dcm2niix_success
-        ):
-            merge_philips_diffusion(
-                to / Path(bids_full_path).with_suffix(".json"),
-                metadata.number_of_parts,
-                metadata.run_num,
+        if dcm2niix_success:
+            scans_filepath = (
+                to
```
NicolasGensollen, line 807: These changes should not be there I believe. They belong to a previously merged PR.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -75,6 +82,20 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     return df
 
 
+def _handle_series_description(x: str) -> str:
+    series_description = pdcm.dcmread(x).SeriesDescription
```
NicolasGensollen, line 86: By curiosity, what is the type of `series_description` ?

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -75,6 +82,20 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
...
+    if any([modality in series_description for modality in ["t1", "t2", "T1", "T2"]]):
+        echo_time = pdcm.dcmread(x).EchoTime
+        repetition_time = pdcm.dcmread(x).RepetitionTime
+        if (
+            any([modality in series_description for modality in ["t2", "T2"]])
+            and echo_time < 60
```
NicolasGensollen, line 92: You explained these magic numbers in the PR but not in the code such that they will still seem pretty magical in a few months/years for someone reading this...

I would both use constant with explicit names (`MAX_ECHO_TIME_FOR_A_T1` and `MAX_REPETITION_TIME_FOR_A_T1` maybe???) and put the explanations you wrote in the PR in a docstring of this function.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -75,6 +82,20 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     return df
 
 
+def _handle_series_description(x: str) -> str:
+    series_description = pdcm.dcmread(x).SeriesDescription
+    if any([modality in series_description for modality in ["t1", "t2", "T1", "T2"]]):
```
NicolasGensollen, line 87: Does it happen in practice to have mixtures of upper and lower case ?

This feels a little weird when reading it. I would expect the value(s) in `series_description` to be made lower case and then compared to "t1" and/or "t2". Does this make sense ?

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Looks mostly good besides the problematic rebase.
I just have a few suggestions and questions.

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
MatthieuJoulot: `SeriesDescription` is a string

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -75,6 +75,28 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
...
+    series_description = pdcm.dcmread(x).SeriesDescription
+    if any([modality in series_description.lower() for modality in ["t1", "t2"]]):
+        echo_time = pdcm.dcmread(x).EchoTime
+        repetition_time = pdcm.dcmread(x).RepetitionTime
+        if (
+            "t2" in series_description.lower()
+            and echo_time < MAX_ECHO_TIME_FOR_A_T1
+            and repetition_time < MAX_REPETITION_TIME_FOR_A_T1
+        ):
+            return "t1"
```
NicolasGensollen, line 96: Is it an issue that the function returns only "t1" in this case, but the full series description in other cases ?

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
```diff
@@ -75,6 +75,28 @@ def filter_dicoms(df: DataFrame) -> DataFrame:
     return df
 
 
+def _handle_series_description(x: str) -> str:
+    """This function handles the series description used to identify the modality later on.
+
+    In the  case of T1 and T2, some dicoms are misnamed. Therefore we use the echo time and repetition time
+    which are supposed to be very different for a T1 and T2, to check the series description corresponds to what has been done.
```
NicolasGensollen, line 82:
```suggestion
    are supposed to be very different between a T1 and T2. Therefore, we use this proxy to check that the series description is coherent.
```

clinica/iotools/converters/genfi_to_bids/genfi_to_bids_utils.py
MatthieuJoulot: No it is not. The function `identify_modality` looks for t1 in the series description

NicolasGensollen (approved):
Let's wait for the CI to finish with the updated data and I'll merge if green.

PR #931, opened 2023-06-01
REF: Use PETPVC task from pydra-petpvc

ghisvail:
~~Requires a new release of `pydra-petpvc` exposing the PSF's FWHM in x, y and z directions as separate arguments.~~


NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #937, opened 2023-06-06
[TEST] Add unit tests for groups utils

NicolasGensollen:
Small PR adding unit tests for utility functions in `clinica.utils.groups`.
Also add type hints and Numpydoc compliance.

I noticed that the `extract_group_ids` function simply lists the folder `caps_directory/groups` without any check, and can easily return things that are invalid group IDs (like file names if there are some within `caps_directory/groups`). 
This could be on purpose if the function assumes a validated CAPS folder (but should be clarified in the docstring I think). Otherwise, it could be useful to add a small safety check.

### DIFF
--- a/clinica/utils/group.py
+++ b/clinica/utils/group.py
@@ -2,6 +2,8 @@
 
 See CAPS specifications for details about groups.
 """
+from os import PathLike
+from typing import List
 
 
 def check_group_label(group_label: str) -> None:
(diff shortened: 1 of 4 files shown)


ghisvail (approved):

NicolasGensollen:
Thanks for the review @ghisvail !

PR #945, opened 2023-06-23
[MAINT] Remove `create_phase_in_radsec` and `resample_fmap_to_b0`

NicolasGensollen:
Remove unused functions:

- `resample_fmap_to_b0` defined in two places but never used
- `create_phase_in_radsec` defined in `clinica.utils.fmap` but never used

This leads to the deletion of module `clinica.utils.fmap`.


MatthieuJoulot (approved):

NicolasGensollen:

PR #946, opened 2023-06-23
[MAINT] remove `get_sessions_map_AIBL`

NicolasGensollen:
This PR proposes to remove the function `get_sessions_map_AIBL` which was in the wrong utility module (iotools bids utils instead of AIBL utils) and wasn't using some of its inputs.
Since it basically does a dictionary comprehension without providing much additional value, I think we can remove it completely.


MatthieuJoulot (approved):

NicolasGensollen:

PR #950, opened 2023-07-10
MNT: Bump scikit-image to 0.21 and lxml to 4.9.3

ghisvail:
Both now provide wheels for macOS on Python 3.11.

Adjustments needed to be made on the similarity measure calculation now that [`data_range` is mandatory for ssim](https://github.com/scikit-image/scikit-image/pull/6612).


ghisvail:
Worth noting that `scikit-image` plans to drop support for Python 3.8 from **version 0.22 onwards**.

NicolasGensollen (approved):

PR #951, opened 2023-07-10
WIP: Use pydra-ants for ANTs tasks

ghisvail:


ghisvail:
Released `pydra-ants` version 0.0.5 with generic `Registration` class and `registration_syn_quick` specialization. It should cover all our needs for ANTs registration within Clinica.

PR #964, opened 2023-09-05
[INFRA] Add `MRtrix3` to conda environnement

NicolasGensollen:
This PR proposes to use the Conda package for MRtrix3: https://anaconda.org/MRtrix3/mrtrix3


ghisvail (approved):
Maybe we could think of adding ANTs next?

PR #965, opened 2023-09-05
MNT: Fix lock file checking

ghisvail:
Triggered by this recent warning message:

```shell
$ make install
poetry lock --check is deprecated, use `poetry check --lock` instead.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ help: Makefile
 
 .PHONY: check.lock
 check.lock:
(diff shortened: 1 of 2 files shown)


NicolasGensollen (commented):
LGTM, but we need to upgrade poetry to `1.6` on the CI machines. I'll take care of that.

NicolasGensollen (approved):

PR #988, opened 2023-10-04
CI: Enable cache for Poetry

ghisvail:
Should speed Conda environment setup by re-using the previous Poetry cache if the project's lock file has not been modified.

This PR implements the following:
- Persistence of Poetry's cache using `actions/setup-python`
- Ensure Poetry uses the Conda environment instead of creating a new one
- Pin the version of Python and Poetry to avoid accidental upgrades

This is restricted to Linux *only*, since the macOS runner machines cannot use `actions/setup-python` as they are configured today.

### DIFF
--- a/.github/workflows/test_converters.yml
+++ b/.github/workflows/test_converters.yml
@@ -20,6 +20,10 @@ jobs:
       - macOS
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


NicolasGensollen (approved):
LGTM, thanks !

PR #971, opened 2023-09-11
[INFRA] Add workflow for non regression test for converters

NicolasGensollen:
Test PR for trying to run the non regression tests of converters through GitHub actions with a self-hosted runner.

PR #989, opened 2023-10-05
[INFRA] Run DWI non regression tests with GitHub actions

NicolasGensollen:
Similar to #987 but for DWI, and following initial work done in  #971


NicolasGensollen:
Closing as this was broken in smaller pieces (#1066 #1067 #1071 #1072 and #1074)

PR #1006, opened 2023-10-24
Use pathlib in pipelines

NicolasGensollen:
This PR proposes to use Pathlib Path objects to represent paths in Pipeline classes instead of raw strings.

It does some refactoring of the pipeline engine first and propagate the changes to all Clinica pipelines.

This needs to be tested on the CI to check that pipelines were not broken.


ghisvail:
You can add `"PTH"` to the [list of selectors](https://github.com/aramis-lab/clinica/blob/259e6e20d0523fb786d37a7e4d38505f83016d0d/pyproject.toml#L100) and let Ruff help you with substituting usage of `os.path` throughout the code base.

ghisvail:
It will also force future code to use `pathlib` over `os.path`.

NicolasGensollen:
Great advice ! Thanks @ghisvail ! 👍

PR #1009, opened 2023-10-27
[ENH] Expose `n_procs` option to `Adni2BIDS` `Oasis2BIDS` and `Aibl2BIDS`

### LINKED ISSUE #1007: Enable user control over the number of jobs for converters
I recently received an email from a user encountering the following problem when trying to use the ADNI2BIDS converter on a cluster:

"""
We've been having some problems as it seems like the conversion is trying to use more CPUs on the HPC than we are allocated, and adni-to-bids doesn't seem to have a -np or --n_procs option to specify the number of threads it can start. At the moment, IT won't let us use the tool until we find a way to limit adni-to-bids from trying to use too many cores.
"""

I initially thought that the converters were not exposing the `--n_procs` option because they were totally sequential, but taking a closer look at the ADNI2BIDS code, there is this deeply-burried parallel loop which greedily takes all available CPU minus one:

https://github.com/aramis-lab/clinica/blob/bf8e61d1c97b08293b75026a91aa4d5d9f1d094c/clinica/iotools/converters/adni_to_bids/adni_utils.py#L1259-L1266

I haven't looked at the other converters, but I imagine they have a similar behavior.

That is clearly an issue as demonstrated by the HPC use case above.

As a solution, I think we should, at least, expose the `--n_procs` option to users like we do for other pipelines.

NicolasGensollen:
Closes #1007


ghisvail (approved):
This approach looks good and should work.

Let me propose a different alternative, which might reduce the amount of code duplication:
- Use Clinica's [n_procs](https://github.com/aramis-lab/clinica/blob/bf8e61d1c97b08293b75026a91aa4d5d9f1d094c/clinica/pipelines/cli_param/option.py#L59) Click option wherever needed.
- Put the logic for computing the default value for `n_procs` in the [default](https://click.palletsprojects.com/en/8.1.x/options/#dynamic-defaults-for-prompts) parameter for the option.
- Expose parameterization with `n_procs` to converters with a safe default value of 1, instead of making it optional and duplicating calls to `get_n_procs` to each pool.
- Plugging the `n_procs` option to the parameter should provide the benefits of safe defaults inside Clinica's code, whilst keeping backward compatibility at the CLI layer.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_av45_fbb_pet.py
```diff
@@ -10,6 +10,7 @@ def convert_adni_av45_fbb_pet(
     conversion_dir: PathLike,
     subjects: Optional[List[str]] = None,
     mod_to_update: bool = False,
+    n_procs: Optional[int] = 1,
```
ghisvail, line 13: Optional can be dropped now.

```suggestion
    n_procs: int = 1,
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_av45_fbb_pet.py
```diff
@@ -33,6 +34,11 @@ def convert_adni_av45_fbb_pet(
     mod_to_update : bool
         If True, pre-existing images in the BIDS directory
         will be erased and extracted again.
+
+    n_procs : int, optional
+        The requested number of processes.
+        If specified, it should be between 1 and the number of available CPUs.
+        Default=1.
```
ghisvail, line 41: Same for the docstring

```suggestion
    n_procs : int, default=1
        The requested number of processes.
        If specified, it should be between 1 and the number of available CPUs.
        Default=1.
```

ghisvail (approved):
Exactly what I had in mind :ok_hand: 

I expected you would  reuse the existing `n_procs` in `cli_param` to minimize the diff, but why not. I added a comment regarding the optional vs keyword param with default value.

NicolasGensollen:
Thanks for the review ! 👍 
The failures are unrelated, so I'll merge this one.

PR #1016, opened 2023-11-15
[ADNI] Handle reading new format of clinical csv

### LINKED ISSUE #1002: ADNI converter unable to handle new csv name
The naming convention has changed a little bit when downloading csv data from LONI, and the ADNI converter cannot handle these changes. The old naming convention was `filename.csv` and now it is `filename_somedate.csv`with the date in the format `ddApryyyy`. The month is written using the first three letters of the month the first one being a capital.

The documentation needs to updated as well.

MatthieuJoulot:
Closes #1002 

ADNI has changed the format of the clinical csv, thus breaking the converter. This PR aims at fixing this problem.


clinica/iotools/converters/adni_to_bids/adni_to_bids.py
```diff
@@ -20,10 +20,8 @@ def get_bids_subjs_info(
     )
 
     # Load all participants from ADNIMERGE.
-    adni_merge_path = path.join(clinical_data_dir, "ADNIMERGE.csv")
-    participants = set(
-        read_csv(adni_merge_path, sep=",", usecols=["PTID"], squeeze=True).unique()
-    )
```
MatthieuJoulot, line 26: Not sure about this one

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1582,3 +1580,17 @@ def remove_tmp_dmc_folder(bids_dir, image_id):
     tmp_dcm_folder_path = join(bids_dir, f"tmp_dcm_folder_{str(image_id).strip(' ')}")
     if exists(tmp_dcm_folder_path):
         rmtree(tmp_dcm_folder_path)
+
+
+def load_clinical_csv(clinical_dir: str, filename: str) -> pd.DataFrame:
+    import re
+    from pathlib import Path
+
+    pattern = filename + "(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
```
NicolasGensollen, line 1589: Are we expecting to have exactly one file matching this pattern ?

If yes, I think we should verify this and raise if more than one file was found. Currently the function returns the last one found. Maybe something like this:

```python
files_matching_pattern = [
    f for f in Path(clinical_dir).rglob("*.csv") if re.search(pattern, (z.name))
]
if len(files_matching_pattern) != 1:
    raise IOError(
        f"Expecting to find exactly one file in folder {clinical_dir} "
        f"matching pattern {pattern}. {len(files_matching_pattern)} "
        f"files were found instead : {'\n- '.join(files_matching_pattern)}"
    )
try:
    return ....
```

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1582,3 +1580,17 @@ def remove_tmp_dmc_folder(bids_dir, image_id):
...
+    for z in Path(clinical_dir).rglob("*.csv"):
+        if re.search(pattern, (z.name)):
+            adni_merge_path = z
+    try:
+        return pd.read_csv(adni_merge_path, sep=",", low_memory=False)
+    except:
+        raise ValueError(f"{filename}.csv was not found. Please check your data.")
```
NicolasGensollen, line 1596: I would be a bit more specific here. If you get to this line, the file exists but the reading as a CSV file failed:

```python
raise ValueError(
    f"File {adni_merge_path} was found but could not "
    "be loaded as a DataFrame. Please check your data."
)
```

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -943,41 +943,41 @@ def create_adni_sessions_dict(
     # Iterate over the metadata files
 
     for location in files:
-
         location = location.split("/")[0]
-        if path.exists(path.join(clinical_data_dir, location)):
-
-            file_to_read_path = path.join(clinical_data_dir, location)
-            cprint(f"\tReading clinical data file: {location}")
+        pattern = location.split(".")[0] + "(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
```
NicolasGensollen, line 947: Any reason not to use your `load_clinical_csv()` function here ?

NicolasGensollen (commented):
Thanks for working on this @MatthieuJoulot !
I think this is looking good. I made some suggestions to have clearer error messages.
I think it would be great to add one or two subject/session with the new format to the CI data such that we can test that we are able to handle both formats. WDYT ?

clinica/iotools/converters/adni_to_bids/adni_utils.py
MatthieuJoulot: I must have missed that one

MatthieuJoulot: Ha yes, actually there was. The way this was built before I changed things was different and I did not see how to use `load_clinical_csv`

clinica/iotools/converters/adni_to_bids/adni_utils.py
MatthieuJoulot: That is not actually what happens. The file is actually not found, and it errors because `adni_merge_path` is therefore not defined.

clinica/iotools/converters/adni_to_bids/adni_utils.py
MatthieuJoulot: Unless you download the data twice, yes, we expect only one file.

clinica/iotools/converters/adni_to_bids/adni_utils.py
NicolasGensollen: I meant, if you implement the suggestion above, the file has to exist otherwise an error would have been raised before. So you would get to this line only if the CSV file wasn't formatted as expected. Right ?

MatthieuJoulot: My bad, I have been reading from bottom to top, I'll come back afterwards

clinica/iotools/converters/adni_to_bids/adni_utils.py
MatthieuJoulot: Done !

clinica/iotools/converters/adni_to_bids/adni_utils.py
MatthieuJoulot: Done as well, works better reading from the top

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1582,3 +1580,27 @@ def remove_tmp_dmc_folder(bids_dir, image_id):
...
     if exists(tmp_dcm_folder_path):
         rmtree(tmp_dcm_folder_path)
+
+
+def load_clinical_csv(clinical_dir: str, filename: str) -> pd.DataFrame:
+    import re
+    from pathlib import Path
+
+    pattern = filename + "(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
+    # for z in Path(clinical_dir).rglob("*.csv"):
```
NicolasGensollen, line 1590:
```suggestion
```

clinica/iotools/converters/adni_to_bids/adni_utils.py
NicolasGensollen: I'm probably missing something, but I don't understand why you couldn't do:

```python
for location in files:
    location = location.split("/")[0]
    df_file = load_clinical_csv(clinical_data_dir, location.split(".")[0])
    df_filtered = filter_subj_bids(df_file, location, bids_ids).copy()
    ...
```

MatthieuJoulot: You are right, it looks like this should work

MatthieuJoulot: I nested this part in a try. I'm not fond of it, but it works. WDYT

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -941,18 +941,11 @@ def create_adni_sessions_dict(
...
     for location in files:
-
         location = location.split("/")[0]
-        if path.exists(path.join(clinical_data_dir, location)):
-
-            file_to_read_path = path.join(clinical_data_dir, location)
-            cprint(f"\tReading clinical data file: {location}")
-
-            df_file = pd.read_csv(file_to_read_path, dtype=str)
+        try:
```
NicolasGensollen, line 946: I would put as few lines as possible in the try-catch and catch explicit errors. In this case, we need to catch the IOError which happens when the number of found files isn't exactly one. It could be a good idea to give a warning and continue the loop. WDYT ?

```python
try:
    df_file = load_clinical_csv(clinical_data_dir, location.split(".")[0])
except IOError as e:
    warnings.warn(e)
    continue
 df_filtered = ...
```

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1582,3 +1568,26 @@ def remove_tmp_dmc_folder(bids_dir, image_id):
     tmp_dcm_folder_path = join(bids_dir, f"tmp_dcm_folder_{str(image_id).strip(' ')}")
     if exists(tmp_dcm_folder_path):
         rmtree(tmp_dcm_folder_path)
+
+
+def load_clinical_csv(clinical_dir: str, filename: str) -> pd.DataFrame:
```
NicolasGensollen, line 1573: Maybe add a small docstring explaining why we need this function and what the pattern enables to capture.

Apart from that, I think this PR looks ready

NicolasGensollen (approved):

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -339,3 +339,26 @@ def test_get_closest_visit(closest_visit_timepoints, image_acquisition_date, exp
...
+            "foobar": [4, 5, 6],
+        }
+    )
+
+
+@pytest.mark.parametrize(
+    "csv_name,csv_to_look_for",
+    [("adnimerge.csv", "adnimerge"), ("adnimerge_20Oct2023.csv", "adnimerge")],
+)
+def test_load_clinical_csv(tmp_path, input_df, csv_name, csv_to_look_for):
```
NicolasGensollen, line 360: Thanks for adding a test @MatthieuJoulot ! 👍

Could you also add a test verifying that the `IOError` is raised when there is no file or more than one file found.
It could also be nice to add a third test testing that the `ValueError` is raised when a single file is found but cannot be opened by Pandas (you could use another delimiter or insert a row with a wrong number of field for instance..).
WDYT?

NicolasGensollen (approved):

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -339,3 +339,68 @@ def test_get_closest_visit(closest_visit_timepoints, image_acquisition_date, exp
...
+    from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
+
+    input_df.to_csv(tmp_path / csv_name, index=False)
+    assert_frame_equal(load_clinical_csv(tmp_path, csv_to_look_for), input_df)
+
+
+@pytest.mark.parametrize(
```
NicolasGensollen, line 367: I would remove the parametrization if you have only one value.

remove `csv_to_look_for` from the arguments of the test function and simply use "adnimerge" in the test function's body.

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -339,3 +339,68 @@ def test_get_closest_visit(closest_visit_timepoints, image_acquisition_date, exp
...
+        match=re.escape(
+            f"Expecting to find exactly one file in folder {tmp_path} "
+            f"matching pattern {csv_to_look_for}{pattern}. 0 "
+            f"files were found instead : \n[- ]"
+        ),
+    ):
+        load_clinical_csv(tmp_path, csv_to_look_for)
+
+
+@pytest.mark.parametrize(
```
NicolasGensollen, line 388: Same comment here

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -339,3 +339,68 @@ def test_get_closest_visit(closest_visit_timepoints, image_acquisition_date, exp
...
+)
+def test_load_clinical_csv_value_error(tmp_path, csv_to_look_for):
+    import re
+
+    from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
+
+    with open(tmp_path / "adnimerge.csv", "w") as fp:
+        fp.write("col1,col2,col3\n1,2,3\n1,2,3,4")
+
+    # input_df.to_csv(tmp_path / csv_name, sep="\t", index=False)
```
NicolasGensollen, line 400:
```suggestion
```

NicolasGensollen (commented):
Thanks @MatthieuJoulot !
Just a small suggestion on removing parametrization for tests using a single value.
LGTM otherwise !

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -339,3 +339,58 @@ def test_get_closest_visit(closest_visit_timepoints, image_acquisition_date, exp
...
+    "csv_name,csv_to_look_for",
+    [("adnimerge.csv", "adnimerge"), ("adnimerge_20Oct2023.csv", "adnimerge")],
+)
+def test_load_clinical_csv(tmp_path, input_df, csv_name, csv_to_look_for):
+    from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
+
+    input_df.to_csv(tmp_path / csv_name, index=False)
+    assert_frame_equal(load_clinical_csv(tmp_path, csv_to_look_for), input_df)
+
+def test_load_clinical_csv_error(tmp_path, ):
```
NicolasGensollen, line 366:
```suggestion
def test_load_clinical_csv_error(tmp_path):
```

NicolasGensollen (commented):
Looks like the linter is unhappy... Could you format ?

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
MatthieuJoulot: strange I thought I did. I will again then

PR #1021, opened 2023-11-22
[ADNI] Expose multiband fMRI conversion

### LINKED ISSUE #740: Why are MB fMRI scans ignored (adni-to-bids)?
**Description**
Per the [adni-to-bids documentation](https://aramislab.paris.inria.fr/clinica/docs/public/latest/Converters/ADNI2BIDS/), multiband fMRI images are ignored. While I understand there might be some differences that need to be accounted for in analysis, I'm wondering why they shouldn't be included in BIDS conversion? Based on my experience, ADNI3 contains a not insignificant number of MB scans. In fact, the only reason I discovered MB images weren't being converted was because many of my subjects have _only_ had MB scans even over multiple years of sessions (examples: 007_S_6255, 023_S_6661).

For reference, [this comment](https://neurostars.org/t/fmri-multiband-preprocessing/6598/2) by Chris Rorden is the best description of multiband fMRI that I've seen.

NicolasGensollen:
Closes #740 

Initial draft which enables this through the list of modalities.

I'm not sure this is the best way to implement this because the code assumes a one-to-many mapping between modalities and converters. However, we do not have separate converters for fMRI and fMRIMB, everything is within the same converter.

Another possibility would be to add a new option to the CLI instead of relying on modalities.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_to_bids.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_to_bids.py
@@ -41,9 +41,18 @@ class AdniToBids(Converter):
         """Return a list of modalities supported.
 
         Returns: a list containing the modalities supported by the converter
-        (T1, PET_FDG, PET_AMYLOID, PET_TAU, DWI, FLAIR, fMRI)
+        (T1, PET_FDG, PET_AMYLOID, PET_TAU, DWI, FLAIR, fMRI, fMRIMB)
         """
-        return ["T1", "PET_FDG", "PET_AMYLOID", "PET_TAU", "DWI", "FLAIR", "fMRI"]
+        return [
+            "T1",
+            "PET_FDG",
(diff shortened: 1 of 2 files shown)


clinica/iotools/converters/adni_to_bids/adni_to_bids.py
```diff
@@ -248,10 +255,19 @@ def convert_images(
...
 
         for modality in modalities:
             if modality not in converters:
-                raise Exception(f"{modality} is not a valid input modality")
+                if modality == "fMRIMB":
+                    cprint(
+                        "The converter will also convert fMRI multiband scans.",
+                        lvl="info",
+                    )
+                    continue
```
NicolasGensollen, line 268: Not super happy with this for example...

NicolasGensollen:
Closing in favor of #1041

PR #1024, opened 2023-11-30
CI: Use pre-commit for linting

ghisvail:
Since we have got an existing `pre-commit` configuration, we might as well use it to lint on our CI.

This PR simplifies the lint job to use the official `pre-commit` GHA, which will be responsible to running `black`, `isort`, `codespell` and future other hooks. It also upgrades versions of some hooks and ensures they can be run with any version of Python available on the host.

Compared to our current setup, `pre-commit` ensures the different linting steps are run in isolation (different environment) and that their corresponding dependencies don't pollute each other or the project. It goes with the recent recommendation to avoid mixing true dev dependencies (such as the testing and coverage libraries) with code maintenance tools (such as black, isort et al.) in the same dev environment.

A follow-up to this PR would be to also change the `Makefile` to use a local version of `pre-commit` (easily installable with `pipx`) when running `make lint` or `make format`.


ghisvail:
Looks like our linting job was a bit sloppy until now. The `pre-commit` one found quite a few style mistakes and a few typos.

NicolasGensollen (approved):
LGTM, thanks @ghisvail !

PR #1025, opened 2023-11-30
MAINT: Add hook to ensure consistency of Poetry files

ghisvail:
In case there is an inconsistency, then one would get:

```
poetry-check.............................................................Failed
- hook id: poetry-check
- exit code: 1

Error: poetry.lock is not consistent with pyproject.toml. Run `poetry lock [--no-update]` to fix it.
```


NicolasGensollen (approved):
LGTM, thanks !

PR #1028, opened 2023-12-04
Add flair to processing check

MatthieuJoulot:
A `flair-linear` pipeline has been added sometime ago. But there is no check in `check-missing-processing` for it.


NicolasGensollen (approved):

PR #1030, opened 2023-12-05
Fix bad warning 

camillebrianceau:
There was a problem with the function `_list_subjects_sub_folders` that prints the sentence 'Could not determine if {folder} is a CAPS or BIDS directory. Clinica will assume this is a CAPS directory.' every time it is called.

### DIFF
--- a/clinica/utils/inputs.py
+++ b/clinica/utils/inputs.py
@@ -75,7 +75,7 @@ def _list_subjects_sub_folders(
     subjects_sub_folders = [
         sub
         for sub in folder_content
-        if (sub.name.startswith("sub-") and (root_dir / sub).is_dir())
+        if (sub.name.startswith("sub-") and sub.is_dir())
     ]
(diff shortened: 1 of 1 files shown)


clinica/utils/inputs.py
```diff
@@ -75,7 +75,7 @@ def _list_subjects_sub_folders(
     subjects_sub_folders = [
         sub
         for sub in folder_content
-        if (sub.name.startswith("sub-") and (root_dir / sub).is_dir())
+        if (sub.name.startswith("sub-") and sub.is_dir())
```
NicolasGensollen, line 78: Replace by 

```python
subjects_sub_folders = [
    sub for sub in folder_content if (sub.name.startswith("sub-") and sub.is_dir())
]
```
to make linter happy

NicolasGensollen (commented):
Thanks @camillebrianceau !
Looks like there is a linting issue but LGTM otherwise !

NicolasGensollen (approved):

PR #1057, opened 2024-01-22
[ENH] Allow the clinica file reader to read dwi dti images

### LINKED ISSUE #1056: Allow the Clinica file reader to get DWI DTI output images
This issue is linked to the following ClinicaDL issue: https://github.com/aramis-lab/clinicadl/issues/515

To be able to read DTI images obtained with the `DWIDTI` pipeline of Clinica, ClinicaDL will need to use the Clinica file reader to get these outputs. 

https://github.com/aramis-lab/clinica/blob/33bb51ab7ba9691f487d9cfdedf6559ea82c11cc/clinica/utils/inputs.py#L517-L526

Currently, the file reader doesn't support that since we never need to read the output of the `DWIDTI` pipeline.

Adding this functionality shouldn't incur a huge maintenance cost though as it mostly revolves around having the proper regexp in `clinica.utils.input_files`, for example here:

https://github.com/aramis-lab/clinica/blob/33bb51ab7ba9691f487d9cfdedf6559ea82c11cc/clinica/utils/input_files.py#L533-L540

NicolasGensollen:
Closes #1056

### DIFF
--- a/clinica/pipelines/dwi_dti/dwi_dti_pipeline.py
+++ b/clinica/pipelines/dwi_dti/dwi_dti_pipeline.py
@@ -231,7 +231,7 @@ class DwiDti(cpe.Pipeline):
         # ==============
         get_bids_identifier = npe.Node(
             interface=nutil.Function(
-                input_names=["caps_dwi_filename"],
+                input_names=["dwi_filename"],
                 output_names=["bids_identifier"],
                 function=extract_bids_identifier_from_filename,
             ),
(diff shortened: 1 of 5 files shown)


NicolasGensollen:
@camillebrianceau I've added you to the reviewers as you will probably be the first user of this feature though ClinicaDL.

The DWI DTI pipeline outputs files in a folder: `caps/subjects/sub-01/ses-M000/dwi/dti_based_processing`.

In this folder, it creates three subfolders: `native_space`, `normalized_space`, and `atlas_statistics`.

The idea of this PR is to be able to query the DTI based measures outputted by the DWI DTI pipeline which have the following names: `<source_file>_space-[b0|T1w]_{FA|MD|AD|RD}.nii.gz` and which can either be in the `native_space` or `normalized_space` folder depending on the space entity (not very BIDS-like but that's the way it currently is).

This PR proposes the following API to query those files:

```python
from clinica.utils.input_files import dwi_dti
from clinica.utils.inputs import clinica_file_reader

# Query the fractional anisotropy in T1w space
query = dwi_dti("FA", space="T1w")

# This is the usual call to the file reader
found_files, errors = clinica_file_reader(
    ["sub-01"],
    ["ses-M000"],
    caps_dir,
    query,
    raise_exception=True
)
```

If you need to query multiple files (for example multiple measures), you can use the `clinica_list_of_files_reader` function which takes a list of queries instead of a single query.

Hopefully, this should be pretty straightforward to implement in ClinicaDL but let me know if you can spot any issue.

Issue #820, opened 2022-11-29
Provide better documentation for SPM standalone setup


ghisvail:
### Discussed in https://github.com/aramis-lab/clinica/discussions/819

<div type='discussions-op-text'>

<sup>Originally posted by **AudreyDuran** November 28, 2022</sup>
Hi,

I am trying to run `t1-volume` pipeline with SPM12 standalone and clinica v0.7.2, on a custom dataset.

Here are some environment variables values defined in my `~/.bashrc`:

```
SPM_HOME="/opt/spm12"
SPM_VERSION=12
SPM_REVISION=r7771
MATLAB_VERSION=R2019b
MCR_VERSION=v97
MCR_UPDATE=9
MCR_INHIBIT_CTF_LOCK=1
MATLABCMD="${SPM_HOME}/run_spm12.sh /opt/mcr/${MCR_VERSION} script"
```

When I run the pipeline, I get the following output:
```
2022-11-28 10:17:48,336:INFO:The t1-volume pipeline is divided into 4 parts:
        t1-volume-tissue-segmentation pipeline: Tissue segmentation, bias correction and spatial normalization to MNI space
        t1-volume-create-dartel pipeline: Inter-subject registration with the creation of a new DARTEL template
        t1-volume-dartel2mni pipeline: DARTEL template to MNI
        t1-volume-parcellation pipeline: Atlas statistics
2022-11-28 10:17:55,639:INFO:Part 1/4: Running t1-volume-segmentation pipeline.




2022-11-28 10:25:57,830:INFO:The pipeline will be run on the following 1282 image(s):
2022-11-28 10:25:57,830:INFO:   sub-00098E7F | ses-w96, ses-w0,
2022-11-28 10:25:57,830:INFO:   sub-00331D3D | ses-w0, ses-w96,
2022-11-28 10:25:57,830:INFO:   sub-003D930B | ses-w0, ses-w96,
2022-11-28 10:25:57,830:INFO:   sub-00438273 | ses-w96, ses-w0,
2022-11-28 10:25:57,830:INFO:   sub-006ACF86 | ses-w96, ses-w0,
2022-11-28 10:25:57,830:INFO:   sub-0095A525 | ses-w0, ses-w96,
2022-11-28 10:25:57,830:INFO:   sub-00DBEADC | ses-w0, ses-w96,
2022-11-28 10:25:57,830:INFO:   sub-01BA0A03 | ses-w96, ses-w0,
[...]

Issue #980, opened 2023-09-29
Update documentation for pipeline `StatisticsSurface`


NicolasGensollen:
The current documentation still refers to the previous Matlab implementation and needs to be updated as we changed for a Brainstat implementation a couple releases back.

The third party page also needs to be updated as well (the pipeline table).

PR #1099, opened 2024-03-13
[DOC] Improve third-party software documentation page

NicolasGensollen:
Partly linked to #820, this PR proposes to improve the documentation for the installation of third party softwares.

Instead of having them simply listed in a bullet point list, this PR proposes to have a proper section for each software, with subsections concerning installation steps and configuration steps.

This also changes the explanations about environment variables which were all grouped in a single section.
This could be annoying to users who might only need a subset of them.
With the proposed organization, they only need to define the environment variables described in the configuration section of the software they are interested in.

Finally, this way of organizing things makes it possible to link directly to specific software from pipeline pages instead of linking the third-party page and let the user scroll through it.

In addition, this PR proposes to add documentation about installing and configuring SPM standalone, which was missing from current doc.

### DIFF
--- a/docs/Pipelines/Stats_Surface.md
+++ b/docs/Pipelines/Stats_Surface.md
@@ -19,7 +19,6 @@ You need to process your data with the [`t1-freesurfer` pipeline](../T1_FreeSurf
 Do not hesitate to have a look at the paragraph **[Specifying what surface data to use](#advanced-specifying-what-surface-data-to-use)** if you want to use your own surface feature.
 
 ## Dependencies
-<!--If you installed the docker image of Clinica, nothing is required.-->
 
 If you only installed the core of Clinica, this pipeline needs the installation of **Matlab** and **FreeSurfer 6.0** on your computer.
 You can find how to install these software packages on the [third-party](../../Third-party) page.
--- a/docs/Third-party.md
+++ b/docs/Third-party.md
@@ -3,8 +3,7 @@
 
 ## Converters
 
-Some converters require a recent version of [dcm2niix](https://github.com/rordenlab/dcm2niix) to transform DICOM files
-into NIfTI:
+Some converters require a recent version of [dcm2niix](https://github.com/rordenlab/dcm2niix) to transform DICOM files into NIfTI:
 
 |                   | dcm2niix |
 |:------------------|:--------:|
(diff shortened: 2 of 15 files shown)


NicolasGensollen:
@thibaultdvx @MatthieuJoulot if you have some time to take a look at this, I'd be interested to know if you think this is more clear.

- The current doc page: https://aramislab.paris.inria.fr/clinica/docs/public/dev/Third-party/
- With the current PR: https://aramislab.paris.inria.fr/clinica/docs/public/PR-1099/Third-party/

NicolasGensollen:
> @NicolasGensollen sure! But do you expect us to follow the whole installation procedure to see if it works, or do you only want comments on the documentation?

Just comments on whether this way of organizing things feel more clear and practical. Installing all third party software is a lot of work ! 😅

docs/Pipelines/Stats_Surface.md
```diff
@@ -19,7 +19,6 @@ You need to process your data with the [`t1-freesurfer` pipeline](../T1_FreeSurf
 Do not hesitate to have a look at the paragraph **[Specifying what surface data to use](#advanced-specifying-what-surface-data-to-use)** if you want to use your own surface feature.
 
 ## Dependencies
-<!--If you installed the docker image of Clinica, nothing is required.-->
 
 If you only installed the core of Clinica, this pipeline needs the installation of **Matlab** and **FreeSurfer 6.0** on your computer.
```
thibaultdvx, line 23: You didn't change this dependencies section like the others. Is it normal?

docs/Third-party.md
```diff
@@ -44,92 +43,230 @@ _*You only need to install ITK if you plan to perform partial volume correction
...
+##### On MacOS
+
+- [Use the official conda package](https://www.mrtrix.org/download/macos-anaconda/).
+- [Use the MacOS pre-compiled application package installer](https://www.mrtrix.org/download/macos-application/).
+- [Use the Homebrew formula](https://github.com/MRtrix3/homebrew-mrtrix3) (although large dependencies such as `XCode` and `Qt5` are required).
+
+!!! note
+    As for Linux, note that using the conda package should be easier in most cases.
+
+### Matlab
```
thibaultdvx, line 196: Matlab is the only software that doesn't have an installation section. Is it ok?

thibaultdvx (commented):
It is much much clearer! 

All links and cross-references are working.

I just left two comments.

docs/Pipelines/Stats_Surface.md
NicolasGensollen: Good catch ! 👍 
I didn't change it on purpose because the whole page is totally outdated.
We dropped SurfStat (to use Brainstat instead which is pure Python) more than one year ago but the documentation page was never updated.
There is an issue opened to remember us to do it one day: #980

docs/Third-party.md
NicolasGensollen: Good question.
It was already like this prior to this PR and I'm not sure why this is the case tbh... 🤔 
I think Matlab users will have a lot of ways to install it depending on specificities which are way outside of Clinica's scope (maybe your university or institution provides some licenses and custom ways to install it for example).
I think we can keep it like this.

PR #1133, opened 2024-04-09
[MAINT] Change to regex for subject conversation to RID + tests

### LINKED ISSUE #1130: Use regex in `_convert_subject_to_rid()`
The small function `_convert_subject_to_rid()` is not as robust as we'd hope since it implicitly assumes a string in the format "XXX_S_XXXX" and should return the last 4 digits:

https://github.com/aramis-lab/clinica/blob/4b1c2a4743b4d7f6d4d06804fb5a3c49bc8f5096/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fdg_pet.py#L307-L323

Returning the slice `[-4:]` works as long as the input string is correctly formatted but could return crazy results if badly formatted. Currently, there is no guarantee that this will be the case as the subject input string comes from the `subjects` list, which is simply extracted from a CSV file:

https://github.com/aramis-lab/clinica/blob/4b1c2a4743b4d7f6d4d06804fb5a3c49bc8f5096/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fdg_pet.py#L95-L98

Using a regex like mentioned in the TODO could improve the function robustness a bit.

AliceJoubert:
Closes #1130 issue

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fdg_pet.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fdg_pet.py
@@ -314,7 +314,7 @@ def _convert_subject_to_rid(subject: str) -> int:
 
     from re import search
 
-    match = search(r'\d{3}_S_(\d{4})', subject)
+    match = search(r"\d{3}_S_(\d{4})", subject)
 
     if match:
         return int(match.group(1))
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1135, opened 2024-04-10
[MAINT] Move read_participant_tsv function and make it private

### LINKED ISSUE #1129: Move `read_participant_tsv` function and make it private
The utility function `read_participant_tsv()` is actually a public function from the `clinica.utils.filemanip` module: 

https://github.com/aramis-lab/clinica/blob/4b1c2a4743b4d7f6d4d06804fb5a3c49bc8f5096/clinica/utils/filemanip.py#L431-L437

even though it is only used within the `get_subject_session_list()` function from the `clinica.utils.participants` module.

https://github.com/aramis-lab/clinica/blob/4b1c2a4743b4d7f6d4d06804fb5a3c49bc8f5096/clinica/utils/participant.py#L178-L181

We probably should move it to `clinica.utils.participants` and make it private.

AliceJoubert:
Closes #1129 issue

### DIFF
--- a/clinica/utils/filemanip.py
+++ b/clinica/utils/filemanip.py
@@ -12,7 +12,6 @@ __all__ = [
     "get_parent",
     "get_subject_id",
     "load_volume",
-    "read_participant_tsv",
     "save_participants_sessions",
     "unzip_nii",
     "zip_nii",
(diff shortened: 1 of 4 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1139, opened 2024-04-12
[MAINT] Delete clinica.utils.mri_registration module

### LINKED ISSUE #1137: Delete `clinica.utils.mri_registration` module
## Description

The module `clinica.utils.mri_registration` only contains a single function `convert_flirt_transformation_to_mrtrix_transformation` which is only used in a single place in the code base:

https://github.com/aramis-lab/clinica/blob/7d2819a9db4a7e06d829e606a1714c854ef912fb/clinica/pipelines/dwi/connectome/pipeline.py#L353

## Steps

- [ ] I propose that we delete this useless module and move the function to `clinica.pipelines.dwi.connectome.utils`.
- [ ] Since it is used as a Nipype interface, it would be great to implement the corresponding adapter in `clinica.pipelines.dwi.connectome.tasks`.
- [ ] Finally, it would be awesome to implement some unit tests for it since they are currently missing.

**Hints for the unit testing:**
The function is basically doing a subprocess call, so we could refactor a bit to have a helper function building and returning the command that should be run (that would be easy to test).
Then, by monkey patching `check_software` and `os.system` we could test the behavior of the function without running the underlying MRTrix command.

AliceJoubert:
Closes #1137 issue


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1140, opened 2024-04-12
[REF] Refactor converters

NicolasGensollen:
This PR proposes a refactoring of the Clinica converters, the primary goal being to offer a consistent API across converters.

- Consistent API for all converters (all converters implement a `convert` function with standardized API)
- Remove logic from the CLI such that it becomes a simple client of the Python API
- Clearly define the user boundary and parse user inputs once (for example paths from strings to Path objects)
- Re-organize ADNI-to-BIDS modality converters to follow the same logic (consistent API, factory...)
- Clarify public / private objects

Python API example:

```python
from clinica.iotools.converters import convert

# Use the ADNI-to-BIDS converter
convert("ADNI", "./raw_data", "./bids", "./clinical_data")

# Use the Oasis-to-BIDS converter
convert("OASIS", "./raw_data", "./bids", "./clinical_data")

...
```


NicolasGensollen:
@AliceJoubert I believe this is finally ready to be merged as all tests are passing for converters.
Let me know if you want to take a look before I merge (the diff is huge so no worries if not).

AliceJoubert:
Thanks for the work @NicolasGensollen !! I think I am going to avoid reviewing this one, you can merge

PR #1147, opened 2024-04-23
[MAINT] Move `merge_nifti_images_in_time_dimension_task` in dwi tasks module

NicolasGensollen:
This PR proposes to move the adapter function `merge_nifti_images_in_time_dimension_task` from `clinica.utils.image` to `clinica.pipelines.dwi.preprocessing.t1.tasks` where other tasks for this pipeline are defined.
The function definition is kept in the image module.

### DIFF
--- a/clinica/pipelines/dwi/preprocessing/t1/tasks.py
+++ b/clinica/pipelines/dwi/preprocessing/t1/tasks.py
@@ -99,3 +99,10 @@ def rename_into_caps_task(
         Path(b_vectors_preproc_filename),
         Path(b0_brain_mask_filename),
     )
+
+
+def merge_nifti_images_in_time_dimension_task(image1: str, image2: str) -> str:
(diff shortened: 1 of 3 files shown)


AliceJoubert (approved):

AliceJoubert:
Thanks Nicolas ! I juste have a question ; should those functions be unit tested ? ì don't think I saw any corresponding unit test

NicolasGensollen:
> Thanks Nicolas ! I juste have a question ; should those functions be unit tested ? ì don't think I saw any corresponding unit test

The function `merge_nifti_images_in_time_dimension` is unit tested here:

https://github.com/aramis-lab/clinica/blob/3fdfea6895d0626f9e66b942863aad048c7ab754/test/unittests/utils/test_image.py#L46-L83

The adapter `merge_nifti_images_in_time_dimension_task` isn't tested by the unit test suite since it is a simple wrapper. It is indirectly tested by the non regression tests though.

PR #1156, opened 2024-04-25
[ENH] Move crop_nifti function

### LINKED ISSUE #1148: Remove code duplication for function `crop_nifti`
The `crop_nifti` function is currently defined in two places:

https://github.com/aramis-lab/clinica/blob/81e058632a332c1a2447e8cae6267bc91015d9d7/clinica/pipelines/t1_linear/anat_linear_utils.py#L59-L94

and 

https://github.com/aramis-lab/clinica/blob/81e058632a332c1a2447e8cae6267bc91015d9d7/clinica/pipelines/pet_linear/pet_linear_utils.py#L105-L133

By a quick inspection, there are exactly the same except that the later was modernized a bit and that the first one returns both the cropped image and the reference.

It would be great to have a single function with a better signature:

```python
def crop_nifti(input_image: Path, reference_image: Path) -> Path:
    ...
```

 and an adapter task for Nipype:

```python
def crop_nifti_task(input_image: str, reference_image: str) -> str:
    return str(crop_nifti(Path(input_image), Path(reference_image)))
```

Note: I don't think we need to return the reference as it is one of the inputs. Modifying the connections in the related pipeline should do the trick in my opinion.

I'm not sure where to put these as it will be used by PETLinear and T1Linear. Maybe we could have the function in `clinica.utils.image` and the adapter within a `tasks.py` module under `clinica.pipelines` which defines tasks shared by all pipelines ?

AliceJoubert:
Closes #1148 issue.

### DIFF
--- a/clinica/pipelines/pet_linear/pet_linear_pipeline.py
+++ b/clinica/pipelines/pet_linear/pet_linear_pipeline.py
@@ -298,6 +298,7 @@ class PETLinear(PETPipeline):
         from nipype.interfaces import ants
 
         import clinica.pipelines.pet_linear.pet_linear_utils as utils
+        from clinica.pipelines.tasks import crop_nifti_task
 
         # Utilitary nodes
         init_node = npe.Node(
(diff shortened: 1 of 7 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1162, opened 2024-04-26
[ENH] Unify DatasetType class

### LINKED ISSUE #1151: Unify the two `DatasetType` enumerations
The `DatasetType` enumeration is defined in two places:

https://github.com/aramis-lab/clinica/blob/f4c10f555333255e0f85706927dafe4ceef5ceb8/clinica/iotools/bids_dataset_description.py#L18-L20

and

https://github.com/aramis-lab/clinica/blob/f4c10f555333255e0f85706927dafe4ceef5ceb8/clinica/utils/inputs.py#L12-L16

in a slightly different way which is confusing.

At the moment there aren't used a lot, and in very different areas of the code base, but if we don't fix this, it might become difficult to understand at some point.

Since a BIDS dataset is considered as a RAW BIDS dataset while a CAPS dataset is considered as a DERIVATIVE BIDS dataset, both enumerations are equivalent within Clinica. Since raw and derivatives are the terms used by the specifications, I would be in favor of having a single enumeration:

```python
class DatasetType(str, Enum):
    RAW = "raw"
    DERIVATIVE = "derivative"
```

Moreover, I don't think it should be defined in the iotools module as it is more general than that (it could very well be used by pipelines or other utilities for example).

There might be better options, but for now, having it in `clinica.utils.input` or `clinica.utils.bids` should already be a better option.

AliceJoubert:
Closes #1151 issue

### DIFF
--- a/test/unittests/utils/test_utils_inputs.py
+++ b/test/unittests/utils/test_utils_inputs.py
@@ -274,7 +274,7 @@ def test_validate_folder_existence(folder_type):
         _validate_folder_existence(1, folder_type)  # noqa
 
     with pytest.raises(
-        ClinicaBIDSError if folder_type == DatasetType.BIDS else ClinicaCAPSError,
+        ClinicaBIDSError if folder_type == DatasetType.RAW else ClinicaCAPSError,
         match=f"The {folder_type.value} directory you gave is not a folder.",
     ):
         _validate_folder_existence(Path("fooooo"), folder_type)
--- a/test/unittests/utils/test_utils_inputs.py
+++ b/test/unittests/utils/test_utils_inputs.py
@@ -286,7 +286,7 @@ def test_check_bids_folder_missing_dataset_description_error(tmp_path):
     with pytest.raises(
         ClinicaBIDSError,
         match=re.escape(
-            f"The BIDS directory ({tmp_path}) you provided is missing a dataset_description.json file."
+            f"The raw directory ({tmp_path}) you provided is missing a dataset_description.json file."
         ),
     ):
         check_bids_folder(tmp_path)
--- a/clinica/iotools/bids_dataset_description.py
+++ b/clinica/iotools/bids_dataset_description.py
@@ -6,6 +6,7 @@ from cattr.gen import make_dict_unstructure_fn, override
 from cattr.preconf.json import make_converter
 
 from clinica.utils.bids import BIDS_VERSION
+from clinica.utils.inputs import DatasetType
 
 
 @define
@@ -15,13 +16,9 @@ class BIDSDatasetDescription:
     See https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html
     """
 
-    class DatasetType(str, Enum):
-        raw = "raw"
-        derivative = "derivative"
-
     name: str
     bids_version: str = BIDS_VERSION
-    dataset_type: DatasetType = DatasetType.raw
+    dataset_type: DatasetType = DatasetType.RAW
 
     def write(self, to: IO[str]):
         import json
--- a/clinica/utils/inputs.py
+++ b/clinica/utils/inputs.py
@@ -12,8 +12,8 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 class DatasetType(str, Enum):
     """Defines the possible types of datasets in Clinica."""
 
-    BIDS = "BIDS"
-    CAPS = "CAPS"
+    RAW = "raw"
+    DERIVATIVE = "derivative"
 
 
 RemoteFileStructure = namedtuple("RemoteFileStructure", ["filename", "url", "checksum"])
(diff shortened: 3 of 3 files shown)


test/unittests/utils/test_utils_inputs.py
```diff
@@ -274,7 +274,7 @@ def test_validate_folder_existence(folder_type):
         _validate_folder_existence(1, folder_type)  # noqa
 
     with pytest.raises(
-        ClinicaBIDSError if folder_type == DatasetType.BIDS else ClinicaCAPSError,
+        ClinicaBIDSError if folder_type == DatasetType.RAW else ClinicaCAPSError,
```
NicolasGensollen, line 277: This test is still broken because of the match. You need to update the following line with:

```python
match=f"The {'BIDS' if folder_type == DatasetType.RAW else 'CAPS'} directory you gave is not a folder.",
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
Besides the broken unit test, it looks good.
The CI is a mess so it's very difficult to tell what's broken by this PR. Could you rebase on newest dev to have the latest fixes ?

test/unittests/utils/test_utils_inputs.py
AliceJoubert: I've juste rebased, ran the tests again and I get no issue without modifying that line. Could you try again ? Thanks !

AliceJoubert: I've just rebased, ran the tests again and they pass without changing the match line since it matches the function `_validate_folder_existence` error message : 
```
    if not directory.is_dir():
        raise (
            ClinicaBIDSError if folder_type == DatasetType.RAW else ClinicaCAPSError
        )(
            f"The {folder_type.value} directory you gave is not a folder.\n"
            "Error explanations:\n"
            f"\t- Clinica expected the following path to be a folder: {directory}\n"
            "\t- If you gave relative path, did you run Clinica on the good folder?"
        )
```
However, I get an error with `test_check_bids_folder_mismatch_with_caps_error` test.

AliceJoubert: It was also a mismatch error so I modified that. I noticed there are currently two ways the error messages are written. With raw/derivative :

```
def test_check_bids_folder_missing_dataset_description_error(tmp_path):
    from clinica.utils.inputs import check_bids_folder

    with pytest.raises(
        ClinicaBIDSError,
        match=re.escape(
            f"The raw directory ({tmp_path}) you provided is missing a dataset_description.json file."
        ),
    ):
        check_bids_folder(tmp_path)
```
or with BIDS/CAPS : 
```
def test_check_bids_folder_mismatch_with_caps_error(tmp_path):
    from clinica.utils.inputs import check_bids_folder

    (tmp_path / "dataset_description.json").touch()
    (tmp_path / "subjects").mkdir()

    with pytest.raises(
        ClinicaBIDSError,
        match=re.escape(
            f"The BIDS directory ({tmp_path}) you provided seems to be a CAPS "
            "directory due to the presence of a 'subjects' folder."
        ),
    ):
        check_bids_folder(tmp_path)

```

Would you also suggest unifying this _ie_ changing everything to raw/derivative ?

NicolasGensollen (commented):
Hi @AliceJoubert 
How did you rebase ? It looks like you have a bunch of commits which shouldn't be there.

AliceJoubert:
@NicolasGensollen 

I might have done this the wrong way. I pulled the changes from my upstream remote to my local dev, then rebased the local datasetType branch on dev. Things got complicated after when trying to get that on my origin remote because `push origin datasetType` failed due to the branches diverging. I think I did `pull origin datasetType --rebase` before pushing again and that's probably where it went wrong.

NicolasGensollen:
@AliceJoubert No worries, I fixed it (but I forced pushed to your branch). I can show you tomorrow how to rebase in a clean way.

The PR should have the latest commits of dev, and we should be back with the unit test error that we need to fix:

```
=========================== short test summary info ============================
FAILED test/unittests/utils/test_utils_inputs.py::test_check_bids_folder_missing_dataset_description_error - AssertionError: Regex pattern did not match.
 Regex: 'The\\ BIDS\\ directory\\ \\(/tmp/pytest\\-of\\-runner/pytest\\-0/test_check_bids_folder_missing0\\)\\ you\\ provided\\ is\\ missing\\ a\\ dataset_description\\.json\\ file\\.'
 Input: 'The raw directory (/tmp/pytest-of-runner/pytest-0/test_check_bids_folder_missing0) you provided is missing a dataset_description.json file.'
================= 1 failed, 857 passed, 43 warnings in 11.73s ==================
```

AliceJoubert:
Alright, thanks @NicolasGensollen !!

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

Issue #1174, opened 2024-05-07
Refactor and test `adni_fmap.py`


NicolasGensollen:
Following PR #1119 there is some cleaning and refactoring to be done around `adni_fmap.py`.

More precisely, we need to do:

- [x] Add non regression test (requires adding fmap data to CI dataset and making sure that the results are consistent)
- [x] Add unit tests
- [x] Refactor portions of the code that are difficult to understand

NicolasGensollen:
Closing this as refactoring has been done, non regression tests have been added, and adding unit tests will be pursued as a continuous effort.

PR #1175, opened 2024-05-07
[ENH] Proposition for adni_fmap renaming function

AliceJoubert:
Refactor of adni_fmap.py following issue #1174 .

It mainly tackles the renaming of the files after using dcm2nix to convert the files. I chose the easy way _ie_ relying on what dcm2nix outputs to rename the files but that can be discussed.


clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -244,165 +245,35 @@ def fmap_image(
...
+    from clinica.utils.stream import cprint
+
+    bids_path = Path(bids_path)
+
+    for file_path in bids_path.rglob(pattern=r"*.json"):
+        str_file_path = str(file_path)
+        if "fmap" in file_path.name:
+            with open(file_path, "r") as f:
+                file_json = json.load(f)
+            if "BidsGuess" in file_json:
```
NicolasGensollen, line 263: We can discuss that next week, but I have some concerns with this approach.
I haven't had time to look closely at the PR, but looking at this: https://github.com/rordenlab/dcm2niix/blob/master/BidsGuess/README.md?plain=1, it seems like `BidsGuess` is quite recent and still experimental.

If I understand correctly, this means that users with a dcm2niix older than `v1.0.20230731` (quite recent) will get an error at this point.

Also, I was a bit concerned by sections of the README linked above like "Therefore, it is unable to resolve fieldmap...". Might be outdated, but still a bit concerning...

I'm not saying that we shouldn't use this option, but I think we should have a backup way in case things go wrong or the user's version of dcm2niix is not recent enough.

WDYT ?

AliceJoubert: I only use 'BidsGuess' to determine what type of image I have (phase diff, magnitude, echo), without having to make the checks myself and from what I have read and tested it seems to work well !

However, I agree there should be another option if the field 'BidsGuess' does not appear in the .json files anyway. I will work on this next week, based on what was proposed by @tharpm95 and https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#types-of-fieldmaps. 

This way, both options will be available. What do you think ?

likeajumprope: Hi all, apologies for just jumping in - I have run the dev branch with the new fieldmaps conversion feature over three subjects. We had a short discussion about the "BidsGuess", but  we are planning to use the output of the converter for fmriprep and we would need an "IntendedFor" key.  Would it be possible to add that?

likeajumprope: Second, I had to make some (small) adjustments to get the fieldmap conversion to run (some fieldmap file extensions were not included) - I might open a small PR.

tharpm95: Commenting here to confirm I've also had the tmp_dcm issue. I had the code to remove those automatically which I think we decided to remove during the merge. It could likely be reintegrated into the more robust for loop to remove these files, or added somewhere else in adni_utils.py

NicolasGensollen: Hi all,

The issue with the `tmp_dcm` folder which is not removed is caused by the function `create_file` (in `adni_utils.py`) returning early.

More precisely, the flag `nifti_exists` is very often set to `False` because the converted image does not have the expected name. This causes the function to warn and return early:

https://github.com/aramis-lab/clinica/blob/a4984bcd2c9ff96e3696d43110955f3375a949c4/clinica/iotools/converters/adni_to_bids/adni_utils.py#L1551-L1556

Such that the following line is not executed:

https://github.com/aramis-lab/clinica/blob/a4984bcd2c9ff96e3696d43110955f3375a949c4/clinica/iotools/converters/adni_to_bids/adni_utils.py#L1602-L1603
 
For example, the expected file

"sub-ADNI941S6471/ses-M024/fmap/sub-ADNI941S6471_ses-M024_fmap.nii.gz"

does not exist, but 

"sub-ADNI941S6471/ses-M024/fmap/sub-ADNI941S6471_ses-M024_fmap_e2.nii.gz"

does.

A quick and dirty way around this, would be to modify the `nifti_exists` flag to take fmap specificities into account. Something like this maybe:

```python
if modality == "fmap":
    nifti_exists = (
        nifti_exists
        or file_without_extension.with_name(f"{file_without_extension.name}_e2.nii.gz").is_file()
        or file_without_extension.with_name(f"{file_without_extension.name}_e2_ph.nii.gz").is_file()
    )
```

@likeajumprope the converter explicitly delete images that have "real" and "imaginary" suffixes added by dcm2niix:

https://github.com/aramis-lab/clinica/blob/a4984bcd2c9ff96e3696d43110955f3375a949c4/clinica/iotools/converters/adni_to_bids/adni_utils.py#L1517-L1521

I believe those are not BIDS compliant, and will definitely break any pipeline that you'd run with clinica. Is there a reason you'd need those ? (my knowledge of field maps is quite limited, so I might be missing something here).

PR #1178, opened 2024-05-13
Fix logic in adni_utils by restoring loops in create_file

AliceJoubert:
Follow-up fix for problems mentioned by users while working on PR #1175 . Temporary directories will be properly deleted and files with "imaginary" and "real" in their names handled accordingly (PR #1177 )


clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1489,119 +1491,112 @@ def create_file(
...
-        output_image = file_without_extension.with_suffix(".nii.gz")
 
-        if image.Is_Dicom:
-            success = run_dcm2niix(
-                input_dir=image_path,
-                output_dir=output_path,
-                output_fmt=output_filename,
-                compress=True if zip_image == "y" else False,
-                bids_sidecar=True if generate_json == "y" else False,
+    if image.Is_Dicom:
```
NicolasGensollen, line 1495: If the modality is "fmap", this will also get executed. I don't think anything will happen since `run_dcm2niix` was already run but it might be better to reorganize things a bit. Maybe something along those lines:

```python
if image.Is_Dicom:
    if modality == "fmap":
        success = all([run_dcm2niix(...) for fmap_image_path in Path(image_path).parent.iterdir()])
    else:
        success = run_dcm2niix(...)
    if not success:
        cprint(....)
for trigger_time in output_path.glob(f"{output_filename}_t[0-9]*"):
    ....
```

WDYT ?

AliceJoubert: > If the modality is "fmap", this will also get executed. I don't think anything will happen since `run_dcm2niix` was already run but it might be better to reorganize things a bit. Maybe something along those lines:
> 
> ```python
> if image.Is_Dicom:
>     if modality == "fmap":
>         success = all([run_dcm2niix(...) for fmap_image_path in Path(image_path).parent.iterdir()])
>     else:
>         success = run_dcm2niix(...)
>     if not success:
>         cprint(....)
> for trigger_time in output_path.glob(f"{output_filename}_t[0-9]*"):
>     ....
> ```
> 
> WDYT ?

You're right, I went a little fast on this

AliceJoubert:
@NicolasGensollen I fixed what you mentioned and also added a check to avoid looking at hidden files.

WDYT ?

clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1476,132 +1476,148 @@ def create_file(
...
+            success = all(
+                [
+                    run_dcm2niix(
+                        input_dir=fmap_image_path,
+                        output_dir=output_path,
+                        output_fmt=output_filename,
+                        compress=True if zip_image == "y" else False,
+                        bids_sidecar=True if generate_json == "y" else False,
+                    )
+                    for fmap_image_path in Path(image_path).parent.iterdir()
```
NicolasGensollen, line 1493: I'm going to approve and merge this PR as it fixes some issues with fmap conversion we currently have on `dev`.

However, I really think we need to be sure about the following point.
For some subjects, there are multiple images for the same session and fmap modality.

For example, at this point of the code, the variable `image_path` could be `raw/177_S_6448/Field_Mapping/2020-09-16_10_31_12.0/I1340867` (assuming `raw` is the folder in which ADNI data were downloaded).

But taking a look at the data, we have several image folders:

```bash
$ ll raw/177_S_6448/Field_Mapping/2020-09-16_10_31_12.0
I1340861
I1340867
I1340868
```

I haven't investigated a lot, but there seem to be three options to run the conversion:

- Run `dcm2niix` with the `image_path` and ignore the others image folders (they do not seem to be taken into account later in the code)
- Run `dcm2niix` with the parent folder of the three images
- Loop over the three image folders and run `dcm2niix` on each (what is implemented here)

These three options do not seem to output the same result such that we need to clarify which one is the "right" one. 

@tharpm95 @likeajumprope does this make sense ? Do you have encountered this case in your tests ? Were you able to decide on which option to chose ?

NicolasGensollen (approved):
LGTM ! Thanks @AliceJoubert !

tharpm95:
Glad to see this is getting merged. Just a heads up I will be out of office for the next couple weeks.

clinica/iotools/converters/adni_to_bids/adni_utils.py
NicolasGensollen: Thanks for your reply @tharpm95 !
I believe we will go with option 2 since it looks like dcm2niix is able to make a slightly better inference for the magnitude1 image name and metadata. But indeed, options 2 and 3 are almost equivalent and both capture all the available phase and magnitude files.

PR #1181, opened 2024-05-16
[FIX] Fix logging in adni to bids

NicolasGensollen:
The function `create_file` in the adni-to-bids converter is executed in a multiprocessing context but doesn't handle the logging of messages in the sense that child processes do not configure their own logger.

This has a few consequences like the fact that messages logged from child processes aren't formatted or might even be missing.

This PR proposes a quick fix of re-configuring the clinica logger in child processes.

Ideally, we should have a process-safe logging architecture which would avoid race conditions (especially when logging to files for instance).
A classical solution is to dedicate a process to the logging task while other working processes send their logging messages through a Queue.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -1356,7 +1356,7 @@ def create_file(
     # Note that logging messages could easily be lost (for example when logging
     # to a file from two different processes). A better solution would be to
     # implement a logging process consuming logging messages from a multiprocessing.Queue...
-    setup_clinica_logging("DEBUG")
+    setup_clinica_logging("INFO")
 
     modality_specific = {
         "t1": {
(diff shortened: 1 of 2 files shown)


clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1344,12 +1344,20 @@ def create_file(
...
     from clinica.iotools.utils.data_handling import center_nifti_origin
     from clinica.utils.pet import ReconstructionMethod, Tracer
     from clinica.utils.stream import cprint
 
+    # This function is executed in a multiprocessing context
+    # such that we need to re-configure the clinica logger in the child processes.
+    # Note that logging messages could easily be lost (for example when logging
+    # to a file from two different processes). A better solution would be to
+    # implement a logging process consuming logging messages from a multiprocessing.Queue...
+    setup_clinica_logging("DEBUG")
```
AliceJoubert, line 1359: I would use the info level here : 
```suggestion
    setup_clinica_logging("INFO")
```

AliceJoubert (approved):
Thank you @NicolasGensollen !!
Messages of all levels from these files now appear properly in the terminal. That also means we get even more info messages than before and we might want to get rid of some of them later

PR #1184, opened 2024-05-17
[FIX] Fix handlers addition to the clinica logger

NicolasGensollen:
Fix a bug introduced in #1181 where handlers were added to the logger in an uncontrolled way.
Thanks for catching this @AliceJoubert !


AliceJoubert (approved):
LGTM @NicolasGensollen 

Thanks for the fix !!

PR #1188, opened 2024-05-22
[FIX] Properly deleting real and imaginary files for FMAP in adni-to-bids converter.

AliceJoubert:
The adni-to-bids converter does not remove the "imaginary" and "real" files in case the modality is FMAP because of the way the files are named after dcm2nix conversion. This PR proposes a quick fix and resolves PR #1177 .

For testing, the subject 002_S_1261 from ADNI can be used.


NicolasGensollen:
> This might be revamped later : in the case of the FMAP modality if there is at least one wrong image (containing "eX_imaginary" or "eX_real" or ".eX") all the files of the session can be deleted.
> 
> WDYT @NicolasGensollen ?

I agree, if there is at least one image file (magnitude or phase diff) with a problematic conversion (as the "real" or "imaginary" suffixes would indicate) we need to remove all of them for this specific conversion, and give a warning to the user.

We might want to extract this piece of logic in a separate function and implement some unit tests to make sure file removal is happening as expected. But, let's target this in a different PR.

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1190, opened 2024-05-23
[ENH] Proposition for adni-to-bids converter FMAP renaming

AliceJoubert:
Refactor of adni-to-bids FMAP modality renaming functionality to address part of issue #1174 

It is based on the outputs of dcm2nix (https://github.com/rordenlab/dcm2niix/blob/master/FILENAMING.md) and what is expected in BIDS specifications (https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#types-of-fieldmaps). 

The recognition of cases is based on file extension (e1, e1_ph...) and the number of files present in the directory. The files are renamed only if a BIDS case is recognized. Else, the files are deleted. Case 1 and 2 are handled, not cases 3 and 4.

For BIDS case 3 (GE scanners) :
- not in ADNI https://adni.loni.usc.edu/support/experts-knowledge-base/question/?QID=1027 ie no need to support it
- confusing case https://neurostars.org/t/does-direct-field-mapping-for-field-map-require-magnitude-nii-gz/20299/6

Should be handled in other PR : 
- Bids Guess option : integrating what was suggested in PR #1175 
- After conversion deletion and checks - continuation from PR #1188


clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +252,220 @@ def fmap_image(
...
+    bids_case = {
+        "case1": case1_1phase_2mag,
+        "case2": case2_2phase_2mag,
+        "case3": direct_fieldmap,
+        "unrecognized": unrecognized_fmap_case,
+    }
+
+    lst_subjects = [d for d in os.listdir(bids_path) if "sub-ADNI" in d]
+
+    for subject in lst_subjects:
```
NicolasGensollen, line 433: Since `bids_path` is a `Path` object, you can simplify these lines into:

```python
for subject in (folder for folder in bids_path.iterdir() if folder.name.startswith("sub-ADNI")):
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +252,220 @@ def fmap_image(
...
+        lst_ses = [
+            d
+            for d in os.listdir(bids_path / subject)
+            if "ses" in d and (bids_path / subject / d).is_dir()
+        ]
+        for ses in lst_ses:
```
NicolasGensollen, line 439: Same here:

```python
for session in (folder for folder in subject.iterdir() if folder.name.startswith("ses-") and folder.is_dir()):
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +252,220 @@ def fmap_image(
...
+    cprint(
+        f"The following files were found in {fmap_path} : {to_delete}."
+        f"They will be deleted as they are not usable.",
+        lvl="info",
+    )
+    for file in to_delete:
+        os.remove(fmap_path / file)
+
+
+def check_case_fmap(fmap_path: Path) -> str or None:
```
NicolasGensollen, line 368: This function infers which BIDS case we are dealing with. Depending on this inference, it returns either a string or None. This return value is then used as a key in the `bids_case` dictionary if it is a string in order to get a function to handle the specific case, and nothing happens if it is None.

This totally works but I would vote for a slightly more robust pattern. First, we can model all of these cases as an enumeration rather than using plain strings and Nones, something like that:

```python
class BIDSFMAPCase(int, Enum):  # used int but could be auto as well since we don't really need the value
    EMPTY_FOLDER = 0
    ONE_PHASE_TWO_MAGNITUDES = 1
    TWO_PHASES_TWO_MAGNITUDES = 2
    DIRECT_FIELDMAPS = 3
    NOT_SUPPORTED = 4
```

Then, you could have a factory for your handlers. Its only responsibility is to return the handler corresponding to the case passed as an input:

```python
def fmap_case_handler_factory(case: BIDSFMAPCase) -> Callable:
    if case == BIDSFMAPCase.EMPTY_FOLDER:
        return empty_folder_handler
    if case == BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES:
        return one_phase_two_magnitudes_handler
    ...
```

All handlers must implement the same interface:

```python
def handler_name(fmap_path: Path):
    ...
```

This means that the empty handler would be something like that:

```python
def empty_folder_handler(fmap_path: Path):
    pass  # could also warn the user for example
```

The function `check_case_fmap` (`infer_case_fmap` maybe ?) would then have the following signature:

```python
def infer_case_fmap(fmap_path: Path) -> BIDSFMAPCase:
    ...
```

And return a variant of the `BIDSFMAPCase` enumeration. Finally, your entry point, i.e. the function `reorganize_fmap` would do something like that:

```python
for subject...:
    for session...:
        fmap_case_handler_factory(infer_case_fmap(fmap_path))(fmap_path)
```

WDYT ?

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +252,220 @@ def fmap_image(
...
+    nb_files = len(files)
+
+    extensions = [
+        re.search(r"fmap(.*).json", f).group(1)
+        for f in os.listdir(fmap_path)
+        if re.search(r"fmap(.*).json", f)
+    ]
+
+    if nb_files == 0:
+        cprint(f"Folder for {sub}, {ses} is empty", lvl="warning")
```
NicolasGensollen, line 382: I would rather put these logging calls into the different handlers

NicolasGensollen (commented):
Thanks for taking care of this @AliceJoubert !
I made a quick first pass with a few suggestions along the way.
The main suggestion will probably change a few things so I'll make a second round once implemented if you agree with it.
It simply restructures a bit what you implemented using slightly more robust patterns.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
AliceJoubert: Thanks @NicolasGensollen !! I was not sure how to do it cleanly, it seems much better like this.

AliceJoubert: I updated the PR with your suggestions. The only change I made was using StrEnum with auto to be more explicit in case at some point it becomes interesting to look at the value

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+    )
+    for file in to_delete:
+        os.remove(fmap_path / file)
+
+
+def infer_case_fmap(fmap_path: Path) -> BIDSFMAPCase:
+    files = [f for f in os.listdir(fmap_path) if not f.startswith(".")]
+    nb_files = len(files)
+
+    extensions = [
```
NicolasGensollen, line 408: You only need a set for `extensions`, so you can cast it to a Set here which would avoid doing it multiple times later.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+
+
+def empty_folder_handler(fmap_path: Path):
+    sub = fmap_path.parent.parent.name
+    ses = fmap_path.parent.name
+    cprint(f"Folder for {sub}, {ses} is empty", lvl="warning")
+
+
+def one_phase_two_magnitudes_handler(fmap_path: Path):
+    "Performs the checks and renaming associated to BIDS specifications, case 1 for fieldmaps"
```
NicolasGensollen, line 319:
```suggestion
    """Performs the checks and renaming associated to BIDS specifications, case 1 for fieldmaps."""
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,235 @@ def fmap_image(
...
+        )
+        not_supported_handler(fmap_path)
+        return
+    # Renaming
+    pattern = r"(.*_)fmap(.*?)(\..*)"
+    for previous_filename in files:
+        rgx = re.search(pattern, previous_filename)
+        cut, extension, type = rgx.group(1), rgx.group(2), rgx.group(3)
+        new_name = cut + direct_fieldmaps_renamer(extension) + type
+        os.rename(fmap_path / previous_filename, fmap_path / new_name)
```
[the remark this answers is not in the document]
AliceJoubert: I modified this a bit since I removed the renaming for case 3. We end up with only two cases that are very similar so I made a simpler renaming function handling both cases and did not use any factory.

WDYT ?

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+    sub = fmap_path.parent.parent.name
+    ses = fmap_path.parent.name
+
+    cprint(
+        f"BIDS Case 1 : expecting 1 phasediff and at least 1 magnitude files for {sub}, {ses}.",
+        lvl="info",
+    )
+
+    files = [f for f in os.listdir(fmap_path) if not f.startswith(".")]
+    js = [f for f in files if f.endswith("ph.json")][0]
```
NicolasGensollen, line 329: If there is no JSON file matching this filter, you'd get an `IndexError`, and if there are multiple files for some reason you might have a weird behavior.  I think we should make sure to have a single file matching this query.
You also have this pattern multiple times (in multiple handlers requiring reading from a JSON file), so I'd vote for having a dedicated function.
Also, you can combine these two lines into one:

```python
def get_json_file_matching_pattern(fmap_path: Path, pattern: str) -> Path:
    json_file = [f for f in fmap_path.iter_dir() if not f.name.startswith(".") and f.name.endswith(pattern)]
    if len(json_file) != 1:
        msg = f"Expected only a single JSON file ending in '{pattern}' in {fmap_path}, but got:"
        msg += "\n".join([f.name for f in json_file])
        raise ValueError(msg)
    return json_file[0]
```

In this function, it would be called as:

```python
json_file = get_json_file_matching_pattern(fmap_path, pattern="ph.json")
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+
+    # Checking for Echotime keys in phase json
+    with open(fmap_path / js, "r") as file:
+        json_data = json.load(file)
+    if "EchoTime1" not in json_data or "EchoTime2" not in json_data:
+        cprint(f'Invalid file {js}, missing "EchoTime" key.', lvl="warning")
+        unrecognized_fmap_case(fmap_path)
+        return
+    # Renaming
+    rename_files(fmap_path, files, BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES)
```
NicolasGensollen, line 339: Now that I see this, I think we can simplify this API by avoiding passing `files` as it can be easily computed from `fmap_path` directly in `rename_files`. I think we can afford the performance cost.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+    js = [f for f in files if f.endswith("ph.json")][0]
+
+    # Checking for Echotime keys in phase json
+    with open(fmap_path / js, "r") as file:
+        json_data = json.load(file)
+    if "EchoTime1" not in json_data or "EchoTime2" not in json_data:
+        cprint(f'Invalid file {js}, missing "EchoTime" key.', lvl="warning")
+        unrecognized_fmap_case(fmap_path)
+        return
+    # Renaming
```
NicolasGensollen, line 338: The code is clear enough

```suggestion
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,217 @@ def fmap_image(
...
+    check_json = [f for f in files if "ph.json" in f]
+    # Checking for Echotime keys in phase jsons
+    for js in check_json:
+        with open(fmap_path / js, "r") as file:
+            json_data = json.load(file)
+        if "EchoTime" not in json_data:
+            cprint(f'Invalid file {js}, missing "EchoTime" key.', lvl="warning")
+            unrecognized_fmap_case(fmap_path)
+            return
+    # Renaming
```
NicolasGensollen, line 362:
```suggestion
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
A few other small suggestions to make the code a bit DRYer.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,225 @@ def fmap_image(
...
+    "Performs the checks and renaming associated to BIDS specifications, case 1 for fieldmaps"
+    sub = fmap_path.parent.parent.name
+    ses = fmap_path.parent.name
+
+    cprint(
+        f"BIDS Case 1 : expecting 1 phasediff and at least 1 magnitude files for {sub}, {ses}.",
+        lvl="info",
+    )
+
+    files = [f for f in os.listdir(fmap_path) if not f.startswith(".")]
```
NicolasGensollen, line 340:
```suggestion
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,225 @@ def fmap_image(
...
+    json_file = get_json_file_matching_pattern(fmap_path, pattern="ph.json")
+
+    # Checking for Echotime keys in phase json
+    with open(fmap_path / json_file, "r") as file:
```
NicolasGensollen, line 344:
```suggestion
    with open(json_file, "r") as file:
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,225 @@ def fmap_image(
...
+            unrecognized_fmap_case(fmap_path)
+            return
+    rename_files(fmap_path, BIDSFMAPCase.TWO_PHASES_TWO_MAGNITUDES)
+
+
+def direct_fieldmaps_handler(fmap_path: Path):
+    sub = fmap_path.parent.parent.name
+    ses = fmap_path.parent.name
+
+    files = [f for f in os.listdir(fmap_path) if not f.startswith(".")]
```
NicolasGensollen, line 380:
```suggestion
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,225 @@ def fmap_image(
...
+    # Assuming filename ends with "fmap" for the fieldmap
+    json_file = get_json_file_matching_pattern(fmap_path, pattern="fmap.json")
+
+    # Checking for 'Units' key in fmap json
+    with open(fmap_path / json_file, "r") as file:
```
NicolasGensollen, line 385:
```suggestion
    with open(json_file, "r") as file:
```

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,225 @@ def fmap_image(
...
+    """Performs the checks and renaming for BIDS spec case 2 for fieldmaps"""
+
+    sub = fmap_path.parent.parent.name
+    ses = fmap_path.parent.name
+    cprint(
+        f"BIDS Case 2 : expecting 2 phase and 2 magnitude files for {sub}, {ses}.",
+        lvl="info",
+    )
+
+    files = [f for f in os.listdir(fmap_path) if not f.startswith(".")]
```
NicolasGensollen, line 363: Let's remove this last bit of redundancy:

```python
def check_json_contains_keys(json_file: Path, keys: Iterable[str]) -> bool:
    with open(json_file, "r") as file:
        json_data = json.load(file)
    missing_keys = [key for key in keys if key not in json_data]
    if missing_keys:
        cprint(f'Invalid file {json_file}, missing the following keys: {missing_keys}.', lvl="warning")
        return False
    return True
```

Here, would be used like:

```python
if not all([check_json_contains_keys(f, "EchoTime") for f in check_json]):
    unrecognized_fmap_case(fmap_path)
else:
    rename_files(fmap_path, BIDSFMAPCase.TWO_PHASES_TWO_MAGNITUDES)
```

In `one_phase_two_magnitudes_handler`:

```python
if not check_json_contains_keys(json_file, ("EchoTime1", "EchoTime2")):
    unrecognized...
else:
    ...
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
Another small suggestion to remove additional redundancy.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -245,165 +253,215 @@ def fmap_image(
...
-                os.rename(fmap_filename, phase_json)
-                if os.path.exists(nifti_file):
-                    phase_nifti = os.path.join(
-                        subroot, filename[:-15] + "phase2.nii.gz"
-                    )
-                    os.rename(nifti_file, phase_nifti)
+            return BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES
+    elif nb_files == 6 and extensions == set(["_e1", "_e2", "_e2_ph"]):
+        return BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES
+    elif nb_files == 8 and extensions == set(["_e1", "_e2", "_e1_ph", "_e2_ph"]):
```
NicolasGensollen, line 423:
```suggestion
    elif nb_files == 8 and extensions == {"_e1", "_e2", "_e1_ph", "_e2_ph"}:
```

NicolasGensollen (commented):
Thanks @AliceJoubert for taking my comments into account !
I believe the code looks good. I just made very minor suggestions to ease readability.
With a few unit tests this should be good to go ! 🚀

AliceJoubert:
Thanks @NicolasGensollen for the comments ! I have added some unit tests, do you mind looking at them ?

test/unittests/iotools/converters/adni_to_bids/adni_modalities/test_adni_fmap.py
```diff
@@ -0,0 +1,306 @@
...
+    ):
+        phase_magnitude_renamer(input_value, case)
+
+
+@pytest.mark.parametrize(
+    "case, input_value",
+    [
+        (BIDSFMAPCase.DIRECT_FIELDMAPS, "e2"),
+        (BIDSFMAPCase.NOT_SUPPORTED, "e1"),
+        (BIDSFMAPCase.EMPTY_FOLDER, "toto"),
```
NicolasGensollen, line 55: `toto` is very French. ;) 
Usually the convention is to name these "foo", "bar", and "baz".

```suggestion
        (BIDSFMAPCase.EMPTY_FOLDER, "foo"),
```

test/unittests/iotools/converters/adni_to_bids/adni_modalities/test_adni_fmap.py
```diff
@@ -0,0 +1,306 @@
...
+    fmap = tmp_path / "fmap"
+    fmap.mkdir()
+    filenames = [
+        "sub01-sesM0_fmap_e1.nii.gz",
+        "sub01-sesM0_fmap_e1.json",
+        "sub01-sesM0_fmap_e2.nii.gz",
+        "sub01-sesM0_fmap_e2.json",
+        "sub01-sesM0_fmap_e2_ph.nii.gz",
+        "sub01-sesM0_fmap_e2_ph.json",
+        ".toto.txt",
```
NicolasGensollen, line 84:
```suggestion
        ".foo.txt",
```

test/unittests/iotools/converters/adni_to_bids/adni_modalities/test_adni_fmap.py
```diff
@@ -0,0 +1,306 @@
...
+        (fmap / f).touch()
+
+    expected_result = {
+        fmap / "sub01-sesM0_magnitude1.nii.gz",
+        fmap / "sub01-sesM0_magnitude1.json",
+        fmap / "sub01-sesM0_magnitude2.nii.gz",
+        fmap / "sub01-sesM0_magnitude2.json",
+        fmap / "sub01-sesM0_phasediff.nii.gz",
+        fmap / "sub01-sesM0_phasediff.json",
+        fmap / ".toto.txt",
```
NicolasGensollen, line 96:
```suggestion
        fmap / ".foo.txt",
```

test/unittests/iotools/converters/adni_to_bids/adni_modalities/test_adni_fmap.py
```diff
@@ -0,0 +1,306 @@
...
+
+def test_rename_files_success_case1(tmp_path):
+    from clinica.iotools.converters.adni_to_bids.adni_modalities.adni_fmap import (
+        rename_files,
+    )
+
+    fmap = tmp_path / "fmap"
+    fmap.mkdir()
+    filenames = [
+        "sub01-sesM0_fmap_e1.nii.gz",
```
NicolasGensollen, line 78: Is this the file format that you get ? I would expect "sub-01_ses-M000_fmap_e1.nii.gz"

test/unittests/iotools/converters/adni_to_bids/adni_modalities/test_adni_fmap.py
```diff
@@ -0,0 +1,306 @@
...
+
+    rename_files(fmap, BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES)
+
+    assert set(fmap.iterdir()) == expected_result
+
+
+def test_rename_files_success_case2(tmp_path):
+    from clinica.iotools.converters.adni_to_bids.adni_modalities.adni_fmap import (
+        rename_files,
+    )
```
NicolasGensollen, line 107: I believe you can simplify a bit these two tests since they share a lot of logic. I think something like the following would work (but I haven't tested it so you might need to tweak it a bit...):

```python
@pytest.fixture
def fmap_case_builder(tmp_path: Path, case: BIDSFMAPCase):
    fmap = tmp_path / "fmap"
    fmap.mkdir()
    filenames = [
        "sub01_sesM0_fmap_e1.nii.gz",
        "sub01_sesM0_fmap_e1.json",
        "sub01_sesM0_fmap_e2.nii.gz",
        "sub01_sesM0_fmap_e2.json",
        "sub01_sesM0_fmap_e2_ph.nii.gz",
        "sub01_sesM0_fmap_e2_ph.json",
        ".toto.txt",
    ]
    if case == BIDSFMAPCase.TWO_PHASES_TWO_MAGNITUDES:
        filenames += [
            "sub01_sesM0_fmap_e1_ph.nii.gz",
            "sub01_sesM0_fmap_e1_ph.json",
        ]
    for f in filenames:
        (fmap / f).touch()

@pytest.fixture
def expected(tmp_path: Path, case: BIDSFMAPCase) -> Set[Path]:
    expected_result = {
        fmap / "sub01-sesM0_magnitude1.nii.gz",
        fmap / "sub01-sesM0_magnitude1.json",
        fmap / "sub01-sesM0_magnitude2.nii.gz",
        fmap / "sub01-sesM0_magnitude2.json",
        fmap / ".toto.txt",
    }
    if case == BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES:
        expected = expected.union(
            {fmap / "sub01-sesM0_phasediff.nii.gz", fmap / "sub01-sesM0_phasediff.json"}
        )
    if case == BIDSFMAPCase.TWO_PHASES_TWO_MAGNITUDES:
        expected = expected.union(
            {
                fmap / "sub01_sesM0_phase2.nii.gz",
                fmap / "sub01_sesM0_phase2.json",
                fmap / "sub01_sesM0_phase1.nii.gz",
                fmap / "sub01_sesM0_phase1.json",
            }
        )
    return expected

@pytest.mark.parametrize(
    "case", [BIDSFMAPCase.ONE_PHASE_TWO_MAGNITUDES, BIDSFMAPCase.TWO_PHASES_TWO_MAGNITUDES]
)
def test_rename_files_success(tmp_path, case, fmap_case_builder, expected):
    from clinica.iotools.converters.adni_to_bids.adni_modalities.adni_fmap import rename_file

    rename_fmap(tmp_path / "fmap", case)

    assert set((tmp_path / "fmap").iterdir()) == expected
```

NicolasGensollen (commented):
Thanks @AliceJoubert !

 I made one or two suggestions but I think the tests look great ! 👍
Could you also rebase on latest dev ? The CI seems not happy for some reason.

AliceJoubert:
Thanks for the review @NicolasGensollen ! I modified the tests according to your suggestions. Do you see anything else ?

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_fmap.py
```diff
@@ -1,11 +1,17 @@
 """Module for converting field maps of ADNI."""
+import json
+import os
+import re
+import shutil
+from enum import StrEnum, auto
```
NicolasGensollen, line 6: I didn't realize it when reviewing, but `StrEnum` is only available in Python 3.11 onward (see https://docs.python.org/3/library/enum.html#enum.StrEnum).
Which is explaining all these unit test failures.

Since we support Python 3.9 onward, I would vote for changing the implementation:

```python
from enum import Enum

class BIDSFMAPCase(str, Enum):
    CASE1 = "case1"
    CASE2 = "case2"
    ...
```

AliceJoubert: Ok, no problem, I will change that !

The PR should not be merged yet though, I noticed the conversion fails when trying to convert all modalities (but succeeds with FMAP only)

Issue #1163, opened 2024-04-26
Ignore hidden CSV files in `load_clinical_csv`


NicolasGensollen:
The function `load_clinical_csv` from the ADNI-to-BIDS converter is using regex to find and open a clinical CSV file, no matter the date that ADNI appends at the end of the file name.

https://github.com/aramis-lab/clinica/blob/31543c22118914e56165c28fd525c4fe5ded9987/clinica/iotools/converters/adni_to_bids/adni_utils.py#L1624-L1661

When there are multiple such files, clinica raises an error which is desired since we shouldn't make a choice for the user.

However, it was reported recently by @HuguesRoy that under MacOS, it can happen that a hidden file is created when the CSV file is opened in a GUI application like Excel. In this situation, the `load_clinical_csv` function crashes by finding two files, the correct one and the temporary hidden file.

Even though this should not happen a lot in practice, making this function a bit more robust by ignoring hidden files should be pretty straightforward.

PR #1195, opened 2024-05-29
ADNI_to_BIDS utils : `load_clinical_csv`, hidden csv file matching, issue #1163

HuguesRoy:
Fix issue #1163

Change `re.search ` by `re.match`. 

It will avoid hidden file matching by constraining to an exact match with the pattern 

if we have two files: 
- correct file: `ADNIMERGE_25Apr2024.csv`
- hidden file: `.ADNIMERGE_25Apr2024.csv`

the pattern is `"ADNIMERGE"+ r"(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"`, only the correct file will be returned.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -1644,7 +1644,7 @@ def load_clinical_csv(clinical_dir: str, filename: str) -> pd.DataFrame:
 
     pattern = filename + r"(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
     files_matching_pattern = [
-        f for f in Path(clinical_dir).rglob("*.csv") if re.search(pattern, (f.name))
+        f for f in Path(clinical_dir).rglob("*.csv") if re.match(pattern, (f.name))
     ]
     if len(files_matching_pattern) != 1:
         raise IOError(


clinica/iotools/converters/adni_to_bids/adni_utils.py
```diff
@@ -1644,7 +1644,7 @@ def load_clinical_csv(clinical_dir: str, filename: str) -> pd.DataFrame:
 
     pattern = filename + r"(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
     files_matching_pattern = [
-        f for f in Path(clinical_dir).rglob("*.csv") if re.search(pattern, (f.name))
+        f for f in Path(clinical_dir).rglob("*.csv") if re.match(pattern, (f.name))
```
NicolasGensollen, line 1647: I agree with @AliceJoubert. Not listing hidden files and folders will be more explicit than changing search for match.
I believe changing the rglob pattern for `.rglob("[!.]*.csv")` should do the trick.
Would be nice to add some unit tests to make sure this is working as expected, especially if there are csv files in subfolders where both the subfolders and the files could be hidden.

NicolasGensollen (commented):
Thanks for the PR @HuguesRoy !

NicolasGensollen:
Closed by #1343

Issue #398, opened 2021-08-30
adni-to-bids converter wastes time on QC for participants not in collection


ghisvail:
## Context

This issue happens when converting collections with a reduced set of participants, for instance when doing triage on a bug for a specific participant.

## Description

The converter will analyze QC irrespective of the participants present in the input collection, which wastes a significant amount of time. Granted, users can restrict the set of participants being processed with a list via the `--subjects_list` option. However, our converters could be a bit more clever and only process the participants present in the input collection without further input.

In practice, this means indexing the input collection and filtering the list of participants appropriately prior to the conversion process. This is what's being done in the revamped NIFD converter for instance.

NicolasGensollen:
@AliceJoubert I think this could be an interesting issue (a bit more involved than the previous ones) which is somehow related to QC.
Of course, there is no rush. Just tagging you to not forget about it.

AliceJoubert:
Resolved with PR #1196 !

PR #1196, opened 2024-05-29
[ENH] ADNI-to-BIDS converts only subjects in the provided ADNI directory if no subjects list is given

AliceJoubert:
This PR adresses the behavior described in issue #398. 

- New functions were added in adni_utils.py to define the subjects list to convert, either from the given raw ADNI dataset or a subject list. Handles cases where there is no corresponding clinical data. 
- Unit tests were also added.

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
@@ -50,9 +50,6 @@ def convert_adni_tau_pet(
     )
     from clinica.utils.stream import cprint
 
-    if not subjects:
-        cprint(f"Processing an empty list of subjects.", lvl="warning")
-
     cprint(
         f"Calculating paths of TAU PET images. Output will be stored in {conversion_dir}."
     )
--- a/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
@@ -50,6 +50,9 @@ def convert_adni_tau_pet(
     )
     from clinica.utils.stream import cprint
 
+    if not subjects:
+        cprint(f"Processing an empty list of subjects.", lvl="warning")
+
     cprint(
         f"Calculating paths of TAU PET images. Output will be stored in {conversion_dir}."
     )
--- a/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
@@ -8,7 +8,7 @@ def convert_adni_tau_pet(
     csv_dir: PathLike,
     destination_dir: PathLike,
     conversion_dir: PathLike,
-    subjects: Optional[List[str]] = None,
+    subjects: List[str],
     mod_to_update: bool = False,
     n_procs: Optional[int] = 1,
 ):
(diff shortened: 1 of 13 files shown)


test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -5,6 +5,55 @@
...
 
 
+@pytest.mark.parametrize(
+    "input, expected",
+    [
+        (
+            {"001_S_0001", "001_S_0002", "001_S_0003"},
+            {"001_S_0001", "001_S_0002", "001_S_0003"},
+        ),
+        ({"001_S_0001", "001_S_00014", ".001_S_0001", "001S0001"}, {"001_S_0001"}),
```
NicolasGensollen, line 15: I might be missing something but I don't understand why you need to parametrize.
Moreover, looking at this, it feels like the same input produces different expectations (which is not what you're testing, but still a bit misleading...).

Maybe split into two test functions without any parametrization. One will test with a provided subject list, and the other one without.
WDYT ?

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -5,6 +5,55 @@
...
+    ],
+)
+def test_check_subjects_list(tmp_path, write_all, input, expected):
+    from clinica.iotools.converters.adni_to_bids.adni_utils import check_subjects_list
+
+    clinical_dir = tmp_path / "clinical_dir"
+    clinical_dir.mkdir()
+
+    if not write_all:
+        input.remove(input[-1])
```
NicolasGensollen, line 49: Bit simpler:

```suggestion
        input.pop()
```

NicolasGensollen (commented):
Thanks @AliceJoubert !

I made a first pass and I think this looks good.
I made some small suggestions to simplify things but nothing major.

clinica/iotools/converters/adni_to_bids/adni_modalities/adni_tau_pet.py
```diff
@@ -51,8 +51,7 @@ def convert_adni_tau_pet(
     from clinica.utils.stream import cprint
 
     if not subjects:
-        adni_merge = load_clinical_csv(csv_dir, "ADNIMERGE")
-        subjects = list(adni_merge.PTID.unique())
+        cprint(f"Processing an empty list of subjects.", lvl="warning")
```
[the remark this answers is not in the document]
AliceJoubert: You're right ! I will do this

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
```diff
@@ -5,6 +5,55 @@
...
+
+    source_dir = tmp_path / "source_dir"
+    source_dir.mkdir()
+
+    for subject in input:
+        (source_dir / subject).touch()
+
+    subjs_list_path = tmp_path / "subjects_list.txt"
+    with open(subjs_list_path, "w") as f:
+        f.write("".join([subj + "\n" for subj in input]))
```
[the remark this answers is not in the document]
AliceJoubert: It does, thanks !

test/unittests/iotools/converters/adni_to_bids/test_adni_utils.py
AliceJoubert: Thanks for the review ! You're right that I should have split in two test functions (list or without), so I changed that. However I still wanted to check that the regex I use to get the subjects from the directory when there is no list is working well, which is why I put a parametrization. Now that the functions are split it might be a bit clearer. 

WDYT ?

NicolasGensollen: Hi @AliceJoubert !

Re-reading this with a rested mind, I think this is fine like this. Sorry for the noise.

AliceJoubert: Alright, no worries :-) ! For clarity I'm thinking of leaving the tests as is (split test). Do you see anything else ?

NicolasGensollen: Sounds good ! Nope, I think this is ready to merge ! 👍

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1199, opened 2024-06-03
[DOC] ADNI-to-BIDS : add Field Mapping (FMAP) specificities

AliceJoubert:
Complete ADNI-to-BIDS documentation :
- Add description for FMAP process steps
- Add known problematic subjects/sessions

### DIFF
--- a/docs/Converters/ADNI2BIDS.md
+++ b/docs/Converters/ADNI2BIDS.md
@@ -69,12 +69,13 @@ Currently, the modalities supported by our converter are:
 | Florbetapir (AV45) PET (amyloid)          | -      | ✓         | ✓      |
 | Florbetaben (FBB) PET (amyloid)           | -      | -         | ✓      |
 | Flortaucipir (AV1451) PET (tau)           | -      | -         | ✓      |
+| Field mapping (FMAP)                      | ✓      | ✓         | ✓      |
 | Clinical data                             | ✓      | ✓         | ✓      |
 
 To convert the imaging data to BIDS, a list of subjects with their sessions is first obtained from the `ADNIMERGE` spreadsheet.
 This list is compared for each modality of interest to the list of scans available, as provided by modality-specific csv files (e.g. `MPRAGEMETA.csv`).
 If the modality was acquired for a specific pair of subject-session, and
-several scans and/or preprocessed images are available, only one is converted:
+several scans and/or preprocessed images are available, only one is converted except for FMAP:
 
 - **T1-weighted MRI** When several scans are available for a single session, the preferred scan (as identified in `MAYOADIRL_MRI_IMAGEQC_12_08_15.csv`) is chosen.
 If a preferred scan is not specified then the higher quality scan (as defined in `MRIQUALITY.csv`) is selected.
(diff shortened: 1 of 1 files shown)


docs/Converters/ADNI2BIDS.md
```diff
@@ -69,12 +69,13 @@ Currently, the modalities supported by our converter are:
...
 | Florbetaben (FBB) PET (amyloid)           | -      | -         | ✓      |
 | Flortaucipir (AV1451) PET (tau)           | -      | -         | ✓      |
+| Field mapping (FMAP)                      | ✓      | ✓         | ✓      |
 | Clinical data                             | ✓      | ✓         | ✓      |
 
 To convert the imaging data to BIDS, a list of subjects with their sessions is first obtained from the `ADNIMERGE` spreadsheet.
 This list is compared for each modality of interest to the list of scans available, as provided by modality-specific csv files (e.g. `MPRAGEMETA.csv`).
 If the modality was acquired for a specific pair of subject-session, and
-several scans and/or preprocessed images are available, only one is converted:
+several scans and/or preprocessed images are available, only one is converted except for FMAP:
```
NicolasGensollen, line 78: I'm not sure about this. Actually, I don't know what would happen if multiple images (meaning multiple full images with magnitude and phases) are available for FMAP since we are passing the parent folder to dcm2niix.
Although we never saw it in our testing datasets, I suppose this could happen if multiple runs were acquired. It would help a lot if we could find a subject in ADNI having multiple FMAP images for one session.

I think we should have a "warning" box saying that FMAP conversion was added in Clinica `0.9.0` and is still experimental. And maybe add a sentence explaining that bugs could occur especially if multiple FMAP images are available for one session.

WDYT?

NicolasGensollen (commented):
Thanks @AliceJoubert !

I think this looks good but I would be a bit clearer on the fact that FMAP conversion is new and experimental (see comment).

AliceJoubert:
Thanks for the review @NicolasGensollen ! I modified to add your suggestion. Do you see anything else ?

docs/Converters/ADNI2BIDS.md
```diff
@@ -100,6 +102,11 @@ The user can easily modify this file if they want to convert additional clinical
 For further details regarding clinica data, we recommend to look at the [ADNI Data Package
 ](https://adni.bitbucket.io/index.html) developed by the Alzheimer's Disease Neuroimaging Initiative.
 
+??? failure "Warning : Field Mapping modality"
```
NicolasGensollen, line 105: I don't see this block in the rendered documentation: https://aramislab.paris.inria.fr/clinica/docs/public/PR-1199/Converters/ADNI2BIDS/

AliceJoubert: I don't see it either, as well as all the changes I made in the last commit (ex : removing the "except for FMAP" part as suggested). Did I miss something I should have done for it to appear ?

NicolasGensollen: Indeed, there is a bug in the automatic documentation deployment pipeline... 😱 I'll open a PR to fix it.
I've updated the docs for this PR manually and the changes you made seem to be there.
Could you just double check that everything looks good in the rendered page ?

AliceJoubert: I am not sure if it is still relevant now that you merged the PR but no I can't see the new changes in the rendered page. I will rebase a bit later :-)

NicolasGensollen: That's weird, I can see them.

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !
I'm gonna merge since I can see the changes in the rendered doc and they look good to me.

PR #1200, opened 2024-06-06
[REF] BIDS : refactor function `create_participants_df` from `bids_utils` 

AliceJoubert:
Small refactor of `create_participants_df` of `bids_utils` : 

Previously : 
- Compared BIDS to clinical data
- Adapted to OASIS and ADNI only
- Using dataframes but not the corresponding existing methods

Planned :
- If possible adapt it to all datasets, at least check on dataset if should not be used for it ; ex AIBL has redundant function
- Simplify code loops
- Tests if needed

Maybe in another function : 
- Compare BIDS result to initial raw dataset, available for all types of datasets (same as cleaning sessions/subjects)

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -75,6 +75,10 @@ def create_participants_df(
 
     from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
     from clinica.utils.stream import cprint
+    # todo : need both clinical_data_dir and clinical_data_spec ?
+    # todo : diff between what is in ADNI raw and BIDS can do get subjects (? not same naming type)
+    # todo : maybe simplify loops
+    # todo : used in all converters ? (only OASIS and ADNI, what happens if used elsewhere?)
 
     fields_bids = ["participant_id"]
     prev_location = ""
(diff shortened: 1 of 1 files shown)


clinica/iotools/bids_utils.py
```diff
@@ -47,147 +47,152 @@ class StudyName(str, Enum):
...
     """
-    import os
     from os import path
 
     import numpy as np
     import pandas as pd
 
     from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
     from clinica.utils.stream import cprint
+    # todo : diff between what is in ADNI raw and BIDS can do get subjects (? not same naming type)
```
NicolasGensollen, line 76: I don't understand what this means

clinica/iotools/bids_utils.py
```diff
@@ -47,147 +47,152 @@ class StudyName(str, Enum):
...
+    # todo : used in all converters ? (only OASIS and ADNI, what happens if used elsewhere?)
+    # todo : typing, test
```
NicolasGensollen, line 78: Actually, I don't believe this function is tested at all.
Even if it is a little bit of work, I think it would be better to start by adding some unit tests to ensure we are not breaking things while refactoring (or at least knowing when it happens). This could be in another PR if you prefer.

clinica/iotools/bids_utils.py
```diff
@@ -47,147 +47,152 @@ class StudyName(str, Enum):
...
 
-    fields_bids = ["participant_id"]
-    prev_location = ""
-    index_to_drop = []
-    subjects_to_drop = []
     study_name = StudyName(study_name)
     location_name = f"{study_name.value} location"
 
-    # Load the data from the clincal specification file
+    # Load the data from the clinical specification file
```
NicolasGensollen, line 83: I don't know if this changes something or not, but I made some changes related to those files in #1140 
Basically, I moved these files to the converter they apply to and used consistent names instead of encoding the converter within the name.

NicolasGensollen (commented):
Thanks for starting working on this @AliceJoubert !

I think it might be better to pause this PR a bit while:

- PR #1140 is merged (I'm confident this can happen soon 🤞 )
- unit tests for this function are added

I know it's hard to start refactoring a piece of code by writing unit tests for the code to be changed, but it is a much safer approach than refactoring first and writing tests afterwards.
Let me know what you think !

PR #1206, opened 2024-06-07
[MAINT] Missing unit tests for bids_utils

AliceJoubert:
Create a unit test for the function `create_participants_df` from `bids_utils` to test before refactoring it. Parametrized only with datasets ADNI, OASIS1 and OASIS3 since the function is only used for ADNI and OASIS1.


AliceJoubert:

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1207, opened 2024-06-13
[ENH] Use enumeration for SUVR reference regions

NicolasGensollen:
The SUVR reference region is modeled as a plain string in Clinica with a set of authorized values defined in a list:

https://github.com/aramis-lab/clinica/blob/42dcbd5fb1641011d2881eb9ce8d790b5a86c730/clinica/utils/pet.py#L138

This PR proposes to use a proper enumeration instead.

### DIFF
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -629,7 +629,7 @@ def bids_pet_nii(
 def pet_volume_normalized_suvr_pet(
     acq_label: Union[str, Tracer],
     group_label: str,
-    region: Union[str, SUVRReferenceRegion],
+    suvr_reference_region: Union[str, SUVRReferenceRegion],
     use_brainmasked_image: bool,
     use_pvc_data: bool,
     fwhm: int = 0,
@@ -637,7 +637,7 @@ def pet_volume_normalized_suvr_pet(
     from pathlib import Path
 
     acq_label = Tracer(acq_label)
-    region = SUVRReferenceRegion(region)
+    region = SUVRReferenceRegion(suvr_reference_region)
 
     if use_brainmasked_image:
         mask_key_value = "_mask-brain"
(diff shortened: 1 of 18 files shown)


clinica/utils/input_files.py
```diff
@@ -603,8 +603,14 @@ def bids_pet_nii(
...
-    rec = "" if reconstruction is None else f"_rec-{reconstruction.value}"
+    trc = ""
+    if tracer is not None:
+        tracer = Tracer(tracer)
+        trc = f"_trc-{tracer.value}"
+    rec = ""
+    if reconstruction is not None:
+        reconstruction = ReconstructionMethod(reconstruction)
+        rec = f"_rec-{reconstruction.value}"
     description = f"PET data"
```
AliceJoubert, line 614: Maybe the variable `description` can be defined earlier as it is now so you can get rid of the redundancy and it would look like this : 

```
description = f"PET data"
trc = ""
if tracer is not None:
        tracer = Tracer(tracer)
        trc = f"_trc-{tracer.value}"
        description += f" with {tracer.value} tracer"
rec = ""
if reconstruction is not None:
        reconstruction = ReconstructionMethod(reconstruction)
        rec = f"_rec-{reconstruction.value}"
        description += f" and reconstruction method {reconstruction.value}"
return {...}
```

AliceJoubert (approved):
LGTM @NicolasGensollen !

clinica/utils/input_files.py
NicolasGensollen: Indeed ! Thanks for the suggestion @AliceJoubert ! 👍 
Implemented in [91adb0a](https://github.com/aramis-lab/clinica/pull/1207/commits/91adb0ada44ac7441bf220159b59285a66d2f9f4).

PR #1208, opened 2024-06-13
[REF] BIDS : proposition for function create_participants_df from bids_utils

AliceJoubert:
Small refactor of create_participants_df of bids_utils to simplify the code. It uses dataframe methods now. The corresponding unit test had to be adapted a little. The function is currently working for OASIS and ADNI datasets but maybe the code could work with AIBL since there is a very similar function in the code specific to the AIBL dataset.

Maybe in another function :
Compare BIDS result to initial raw dataset, available for all types of datasets (same as cleaning sessions/subjects)

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -159,7 +159,6 @@ def create_participants_df(
             extracted_df, how="outer", on="alternative_id_1"
         )
 
-    participants_df.reset_index(inplace=True, drop=True)
     # Adding BIDS id based on study specific id
     participants_df["participant_id"] = participants_df["alternative_id_1"].apply(
         lambda x: rename_study_to_bids_id(study_name, x)
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -49,6 +49,17 @@ BIDS_VALIDATOR_CONFIG = {
 }
 
 
+def rename_study_to_bids_id(study_name: StudyName, study_id: str) -> str:
+    # todo : probably useful somewhere else
+    # todo : ok with "sub-truc" devant ?
+    if study_name == StudyName.OASIS:
+        return "sub-OASIS1" + study_id.split("_")[1]
+    elif study_name == StudyName.OASIS3:
+        return "sub-OASIS3" + study_id.replace("OAS3", "")
+    elif study_name == StudyName.ADNI:
+        return "sub-ADNI" + remove_space_and_symbols(study_id)
+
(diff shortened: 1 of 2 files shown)


clinica/iotools/bids_utils.py
```diff
@@ -85,117 +96,86 @@ def create_participants_df(
     import numpy as np
 
     from clinica.iotools.converters.adni_to_bids.adni_utils import load_clinical_csv
-    from clinica.utils.stream import cprint
 
-    fields_bids = ["participant_id"]
-    prev_location = ""
-    index_to_drop = []
-    subjects_to_drop = []
```
NicolasGensollen, line 93: I would also make sure `study_name` is a supported value here. Something like:

```python
if study_name not in (StudyName.ADNI, StudyName.OASIS, StudyName.OASIS3):
    raise ValueError(f"Study {study_name.value} is not supported by 'create_participants_df'.")
```

And it'd be great to add a unit test covering this error.
If we adapt this function for other studies later, we'll update these.
WDYT ?

clinica/iotools/bids_utils.py
```diff
@@ -49,6 +49,17 @@ class StudyName(str, Enum):
 }
 
 
+def rename_study_to_bids_id(study_name: StudyName, study_id: str) -> str:
```
NicolasGensollen, line 52:
```suggestion
def _rename_study_to_bids_id(study_name: StudyName, study_id: str) -> str:
```

clinica/iotools/bids_utils.py
```diff
@@ -85,117 +96,86 @@ def create_participants_df(
...
+                ~extracted_df["alternative_id_1"].str.endswith("_MR2")
+            ]
+
+        participants_df = participants_df.merge(
+            extracted_df, how="outer", on="alternative_id_1"
         )
+
+    # Adding BIDS id based on study specific id
+    participants_df["participant_id"] = participants_df["alternative_id_1"].apply(
+        lambda x: rename_study_to_bids_id(study_name, x)
```
NicolasGensollen, line 164:
```suggestion
        lambda x: _rename_study_to_bids_id(study_name, x)
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
I made a quick pass with some small suggestions. I will try to take a closer look later today.
In the mean time, it looks like this PR broke the ADNI and OASIS non regression tests: https://github.com/aramis-lab/clinica/actions/runs/9513036831/job/26222285415?pr=1208

PR #1217, opened 2024-06-27
[FIX] Fix issue with enumeration as default value in CLI

NicolasGensollen:
Should fix the error `TypeError: EnumType.call() missing 1 required positional argument: 'value'` when using the ADNI-to-BIDS converter without any modality.
Recently introduced in #1140


AliceJoubert (approved):
All good, thanks @NicolasGensollen !!

Issue #1172, opened 2024-05-06
[Converter] ADNI-to-BIDS does not take into account history of conversion when ran successively to the same folder


AliceJoubert:
While testing for PR #1119 , I found a small bug with the ADNI to BIDS converter. Trying to run successively different modalities and save them in the same folder raises an error when creating the 'scans.tsv' file.

**To Reproduce**
Run successively the ADNI-2-BIDS converter on the same dataset for two different modalities and direct it to the same output folder. For example, running :

`clinica convert adni-to-bids dir_raw_data dir_clinical_data dir_output_bids -m T1`
`clinica convert adni-to-bids dir_raw_data dir_clinical_data dir_output_bids -m FLAIR`

Gives an error : 
 ```
...
 File "/Users/alice.joubert/clinica/clinica/iotools/converters/adni_to_bids/adni_to_bids_cli.py", line 79, in cli
    adni_to_bids.convert_clinical_data(
  File "/Users/alice.joubert/clinica/clinica/iotools/converters/adni_to_bids/adni_to_bids.py", line 155, in convert_clinical_data
    adni_utils.create_adni_scans_files(conversion_path, bids_subjs_paths)
  File "/Users/alice.joubert/clinica/clinica/iotools/converters/adni_to_bids/adni_utils.py", line 1124, in create_adni_scans_files
    conversion_df = converted_dict[converted_mod]
                    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 't1'

```

And the 'scans.tsv' file is saved with only FLAIR - related lines in it.

**Expected behavior**
While it may not be the primary way the adni-to-bids converter is used, it would be great to take into account what was already done with the dataset.

NicolasGensollen:
Indeed. Thanks for opening this @AliceJoubert ! 👍

PR #1218, opened 2024-06-28
[FIX] ADNI-to-BIDS now takes into account existing scans.tsv

AliceJoubert:
Resolves bug described in #1172 .


NicolasGensollen (commented):
Thanks for working on this @AliceJoubert !
The PR LGTM, but it'd be awesome to add some unit tests to verify the behavior.

PR #1220, opened 2024-06-28
[ENH] Proposition of a function to translate study ids to bids ids

AliceJoubert:
Add a function translating source/study ids to bids ids.

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -50,8 +50,10 @@ BIDS_VALIDATOR_CONFIG = {
 
 
 def _rename_study_to_bids_id(study: StudyName, study_id: str) -> str:
-    if study == StudyName.ADNI or study == StudyName.NIFD:  # X_S_X -> sub-ADNIXSX
-        return "sub-" + StudyName(study) + remove_space_and_symbols(study_id)
+    if study == StudyName.ADNI:  # X_S_X -> sub-ADNIXSX
+        return "sub-ADNI" + remove_space_and_symbols(study_id)
+    if study == StudyName.NIFD:  # X_S_X -> sub-NIFDXSX
+        return "sub-NIFD" + study_id.replace("_", "")
     if study == StudyName.AIBL or study == StudyName.UKB:  # X -> sub-AIBLX
         return "sub-" + StudyName(study) + study_id
     if study == StudyName.GENFI or study == StudyName.OASIS3:  # Y -> sub-Y
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -51,22 +51,16 @@ BIDS_VALIDATOR_CONFIG = {
 
 # todo : test it + use it
 def study_to_bids_id(study: StudyName, study_id: str) -> str:
-    if study == StudyName.ADNI:  # X_S_X -> sub-ADNIXSX
-        return "sub-ADNI" + remove_space_and_symbols(study_id)
-    if study == StudyName.AIBL:  # X -> sub-AIBLX
-        return "sub-AIBL" + study_id
-    if study == StudyName.GENFI:  # Y -> sub-Y
+    if study == StudyName.ADNI or study == StudyName.NIFD:  # X_S_X -> sub-ADNIXSX
+        return "sub-" + StudyName(study) + remove_space_and_symbols(study_id)
+    if study == StudyName.AIBL or study == StudyName.UKB:  # X -> sub-AIBLX
+        return "sub-" + StudyName(study) + study_id
+    if study == StudyName.GENFI or study == StudyName.OASIS3:  # Y -> sub-Y
         return "sub-" + study_id
     if study == StudyName.HABS:  # P_X -> sub-HABS_X
         return study_id.replace("P_", "sub-HABS")
-    if study == StudyName.NIFD:  # X_S_X -> sub-NIFDXSX
-        return "sub-NIFD" + remove_space_and_symbols(study_id)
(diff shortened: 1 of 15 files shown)


clinica/iotools/bids_utils.py
```diff
@@ -49,6 +49,26 @@ class StudyName(str, Enum):
 }
 
 
+def _rename_study_to_bids_id(study: StudyName, study_id: str) -> str:
```
NicolasGensollen, line 52: I think this could be the opportunity to properly model these IDs. The issue with this function is that inputs and outputs are still plain strings. Somewhere in the code base, when finding something like:

```python
def my_func_adni(subject_id: str):
    a, b =subject_id.split("_S_") # only works if ADNI raw ID, fails otherwise
```

There is no way to know whether `subject_id` is in BIDS format or raw format (or some crazy value resulting from an error somewhere else). And so, it is difficult to know, without browsing the code base or the doc if it exists, whether the call to `.split` is valid or not for example.

These IDs have very standardized structures that we do not use to our advantage when modeling them as plain strings. I think they should have their own types such that we know the associated value has been validated already.

There are multiple ways to do this, but maybe something like this would make sense:

```python
import re
from abc import ABC, abstractmethod
from collections import UserString

class BIDSSubjectID(ABC, UserString):
    """This is the interface that BIDS subject IDs have to implement."""
    def __init__(self, value: str):
        instance = super().__init__(self.validate(value))
        return instance

    @abstractmethod
    def validate(self, value: str) -> str:
        raise NotImplementedError

    @classmethod
    @abstractmethod
    def from_original_study_id(cls, study_id: str):
        raise NotImplementedError

    @abstractmethod
    def to_original_study_id(self) -> str:
        """Not sure we need that..."""
        raise NotImplementedError

class ADNIBIDSSubjectID(BIDSSubjectID):
    """Here is the implementation for ADNI."""
    def validate(self, value: str) -> str:
        if re.fullmatch(r"sub-ADNI\d{3}S\d{4}", value):
            return value
        raise ValueError(
            f"ADNI subject ID {value} is not properly formatted. "
            "Expecting a 'sub-ADNIXXXSXXX' format."
        )
        
    @classmethod
    def from_original_study_id(cls, study_id: str):
        if re.fullmatch(r"\d{3}_S_\d{4}", study_id):
            return cls("sub-" + "ADNI" + remove_space_and_symbols(study_id))
        raise ValueError(
            f"Raw ADNI subject ID {study_id} is not properly formatted. "
            "Expecting a 'XXX_S_XXX' format."
        )

    def to_original_study_id(self) -> str:
        return "_S_".join(self.split("ADNI")[1].split("S"))
```

This would result in:

```python
>>> s = ADNIBIDSSubjectID.from_original_study_id("001_S_1234")
>>> s
'sub-ADNI001S1234'
>>> ADNIBIDSSubjectID.from_original_study_id("foo")
ValueError: Raw ADNI subject ID foo is not properly formatted. Expecting a 'XXX_S_XXX' format.
>>> s = ADNIBIDSSubjectID("sub-ADNI123S4567")
>>> s
'sub-ADNI123S4567'
>>> ADNIBIDSSubjectID("foo")
ValueError: ADNI subject ID foo is not properly formatted. Expecting a 'sub-ADNIXXXSXXX' format.
>>> s.to_original_study_id()
'123_S_4567'
```

WDYT ?

NicolasGensollen (commented):
Thanks @AliceJoubert !

I made a suggestion which would require more work but that could help us making the subject ID related logic more robust in the converters.
Let me know what you think.

clinica/iotools/bids_utils.py
AliceJoubert: Thanks @NicolasGensollen for the suggestion ! I will apply it.

clinica/iotools/bids_utils.py
```diff
@@ -49,6 +52,240 @@ class StudyName(str, Enum):
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def to_original_study_id(self) -> str:
+        """Not sure we need that..."""
+        # todo : verify needed
+        raise NotImplementedError
+
+
+def _id_factory(study: StudyName) -> Type[BIDSSubjectID]:
```
NicolasGensollen, line 78: Looks like you are using this function outside the `bids_utils` module, so probably means it should be public. And maybe `bids_id_factory` ?

```suggestion
def bids_id_factory(study: StudyName) -> Type[BIDSSubjectID]:
```

clinica/iotools/bids_utils.py
```diff
@@ -49,6 +52,240 @@ class StudyName(str, Enum):
...
+    if study == StudyName.OASIS:
+        return OASISBIDSSubjectID
+    if study == StudyName.OASIS3:
+        return OASIS3BIDSSubjectID
+    if study == StudyName.HABS:
+        return HABSBIDSSubjectID
+
+
+class ADNIBIDSSubjectID(BIDSSubjectID):
+    """Here is the implementation for ADNI."""
```
NicolasGensollen, line 98: Maybe rephrase this a bit.

NicolasGensollen (commented):
Thanks @AliceJoubert !
Not sure whether this was ready for review or not, but took a quick look anyway.
I just made a super small suggestion, but LGTM otherwise so far.

AliceJoubert:
> Thanks @AliceJoubert ! Not sure whether this was ready for review or not, but took a quick look anyway. I just made a super small suggestion, but LGTM otherwise so far.

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1223, opened 2024-07-08
[REF] Move class `PETPipeline` to new `clinica.pipelines.pet` module

NicolasGensollen:
PR #1122 is applying the same kind of refactoring to PET pipelines as the one we did a few months ago to DWI pipelines for example. The basic idea is to regroup similar pipelines and related code in the same sub-module.

Since #1122 is enormous and hasn't been updated in a while, I believe it would be easier to split it in smaller PRs.

This is the first of the series and it does the following:

- create new `pet` submodule under `clinica.pipelines`
- move the `PETPipeline` class definition to this new module
- updates the imports in the existing pipelines


AliceJoubert (approved):
LGTM ! @NicolasGensollen

PR #1224, opened 2024-07-08
[REF] Move `PETLinear` pipeline to new `clinica.pipelines.pet` module

NicolasGensollen:
Following #1223, this PR is a portion of #1122 focusing on the `PETLinear` pipeline.

More precisely, it does the following:

- Create module `clinica.pipelines.pet.linear`
- Move previous `clinica.pipelines.pet_linear` module to this new module
- Use the classical structure in the new module:
    - `cli.py`: Contains the CLI
    - `pipeline.py`: Contains the pipeline definition
    - `utils.py`: Contains the utility functions used by the `PETLinear` pipeline **only**
    - `tasks.py`: Contains the adapters for Nipype function interfaces built around utility functions
- Use better variable names
- Use type hints everywhere with better types (`Path` vs `str` mainly)
- Update tests and some imports in the Pydra module

Note: Needs to be tested against regression test suite to make sure I didn't break anything.


AliceJoubert (approved):
Looks good, @NicolasGensollen  !

PR #1226, opened 2024-07-10
[REF] Move `PETVolume` pipeline to new `clinica.pipelines.pet` module 

NicolasGensollen:
Same as #1224 but for `PETVolume`.

Also add `clinica.pipelines.pet.utils` for PET-specific utilities which are used by multiple PET pipelines (might not be true at the moment since `PETSurface` still needs to be moved as well in a follow-up PR).


clinica/pipelines/pet/volume/utils.py
```diff
@@ -187,91 +183,79 @@ def pet_pvc_name(pet_image: str, pvc_method: str) -> str:
...
-        out_atlas_statistics = Path.cwd() / f"{base}_space-{atlas}_statistics.tsv"
-        statistics_on_atlas(in_image, atlas, out_atlas_statistics)
+    for atlas in atlas_names:
+        out_atlas_statistics = (
+            Path.cwd() / f"{get_filename_no_ext(image)}_space-{atlas}_statistics.tsv"
+        )
+        statistics_on_atlas(image, atlas, out_atlas_statistics)
         atlas_statistics_list.append(out_atlas_statistics)
-        break
+        break  # Why is there a break here ????
```
AliceJoubert, line 256: Feels like it would happen all the time ; is it not possible just to remove it ?

NicolasGensollen: It will indeed happen all the time and results in this function returning a list with a single value.
It doesn't make any sense and contradicts the documentation of the function, but I haven't checked whether calling code is relying on this single value list or not.
I suggest we investigate later (i.e. not in this PR) since the aim is only to refactor things.

AliceJoubert: Alright then, the PR LGTM !

NicolasGensollen: Created #1228 to remind us about that.

PR #1227, opened 2024-07-10
[FIX] Change check on type in converter NIFD 

AliceJoubert:
This PR modifies a check on the type of a data frame to adapt it to data changes.


NicolasGensollen (commented):
Any chance we could add a non regression test (failing with the previous code and passing with new) ?

Since this function is currently untested, if you think this is too much work, I'm OK to merge as is.

AliceJoubert:
Since I wanted to do that modification for a user I would be happy to merge this quickly actually ! But I can keep in mind that there is a test to do on this one for later.

NicolasGensollen (approved):
Sounds good, let's merge then !

PR #1235, opened 2024-07-15
[FIX] Fix cross sectional to longitudinal conversion logic

### LINKED ISSUE #1234: [BUG] Type issue when trying to make a cross-sectional BIDS dataset longitudinal
While trying to use T1-linear with the latest dev version on an old BIDS dataset, I found the following : 

- that dataset is not BIDS compliant because it is not organized per session (there is no ses-M000 folder)
- Clinica tries to reorganize it to make it BIDS compliant (meaning adding that ses-M000 folder) after asking the user
- the code breaks when applying the function ```_copy2_add_session_label``` from ```pipelines.engine.py``` because one of the arguments is actually a string when it is expected to be a path :

```
  File "/Users/alice.joubert/clinica/clinica/pipelines/engine.py", line 248, in _copy2_add_session_label
    return copy2(src, dst / _add_session_label(src.name))
                                               ^^^^^^^^
AttributeError: 'str' object has no attribute 'name'
```

The function is called here, where objects are defined as Paths :
```
def _copy_cross_sectional(file_to_copy: Path, output_folder: Path) -> None:
    from shutil import copy2, copytree

    if not (output_folder / file_to_copy.name).exists():
        if file_to_copy.is_dir():
            copytree(
                file_to_copy,
                output_folder / file_to_copy.name,
                copy_function=_copy2_add_session_label,
            )
```

and actual paths are given to that function :

```
> .../clinica/clinica/pipelines/engine.py(305)_copy_cross_sectional()
-> copytree(
(Pdb) file_to_copy
PosixPath('.../BIDS/sub-IXI640/anat')
(Pdb) output_folder / file_to_copy.name
PosixPath('.../BIDS/BIDS_clinica_compliant/sub-IXI640/ses-M000/anat')
```

However what ends up being called in the actual ```_copy2_add_session_label``` function defined as : 
```
def _copy2_add_session_label(src: Path, dst: Path) -> Path:
    """Calls copy2 function from shutil, but modifies the filename of the
    copied files if they match the regex template described in the
    _add_session_label() function
    ...
    """
    from shutil import copy2

    return copy2(src, dst / _add_session_label(src.name))
```

is actually a str : 
```
> .../clinica/clinica/pipelines/engine.py(250)_copy2_add_session_label()
-> return copy2(src, dst / _add_session_label(src.name))
(Pdb) src
'.../BIDS/sub-IXI640/anat/sub-IXI640_T1w.nii.gz'
```

NicolasGensollen:
Closes #1234

### DIFF
--- a/clinica/pipelines/engine.py
+++ b/clinica/pipelines/engine.py
@@ -225,7 +225,7 @@ def _add_session_label(filename: str) -> str:
         return filename
 
 
-def _copy2_add_session_label(src: Path, dst: Path) -> Path:
+def _copy2_add_session_label(src: str, dst: str) -> str:
     """Calls copy2 function from shutil, but modifies the filename of the
     copied files if they match the regex template described in the
     _add_session_label() function
@@ -245,7 +245,9 @@ def _copy2_add_session_label(src: Path, dst: Path) -> Path:
     """
     from shutil import copy2
 
-    return copy2(src, dst / _add_session_label(src.name))
+    filename = Path(src).name
+
+    return copy2(src, str(Path(dst).parent / _add_session_label(filename)))
 
 
 def _convert_cross_sectional(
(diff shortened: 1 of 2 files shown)


AliceJoubert (approved):
LGTM thanks @NicolasGensollen ! Just an additional test I made : in case you give a participants.tsv the whole cross-sectional dataset will be converted to longitudinal ; however the pipeline itself does run properly only on the selected participants.

PR #1236, opened 2024-07-16
[ENH] Add enumeration for hemispheres

NicolasGensollen:
Small PR proposing to add a proper enumeration for hemispheres instead of hardcoding "lh" and "rh" across the code base.

I haven't used it in `PETSurface` related code since I will directly use it in a future refactoring PR similar to the ones I did for `PETLinear` and `PETVolume`.


AliceJoubert (approved):
LGTM thanks @NicolasGensollen !!

PR #1238, opened 2024-07-17
[ENH] Add function `log_and_raise`

NicolasGensollen:
Very small PR which proposes to add a `log_and_raise` function to `clinica.utils.stream` to automatically log (using `cprint`) and raise a given error message. This will be useful for example in #1237 

Also modifies `adni_json.py` where a less generic function was used to log and raise a specific error type.


AliceJoubert (approved):

PR #1239, opened 2024-07-17
[CONV] Implementing new converter for IXI dataset

AliceJoubert:
This PR adds a new converter for the IXI dataset. It covers the implementation and the unit testing only, since non regression tests and documentation will be added in other PRs.

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -25,6 +25,7 @@ class StudyName(str, Enum):
     OASIS = "OASIS"
     OASIS3 = "OASIS3"
     UKB = "UKB"
+    IXI = "IXI"
 
 
 BIDS_VALIDATOR_CONFIG = {
@@ -90,6 +91,8 @@ def bids_id_factory(study: StudyName) -> Type[BIDSSubjectID]:
         return OASIS3BIDSSubjectID
     if study == StudyName.HABS:
         return HABSBIDSSubjectID
+    if study == StudyName.IXI:
+        return IXIBIDSSubjectID
 
 
 class ADNIBIDSSubjectID(BIDSSubjectID):
@@ -290,6 +293,34 @@ class HABSBIDSSubjectID(BIDSSubjectID):
 
     def to_original_study_id(self) -> str:
         return str(self.replace("sub-HABS", "P_"))
+
+
+class IXIBIDSSubjectID(BIDSSubjectID):
+    """Implementation for IXI of the BIDSSubjectIDClass, allowing to go from the source id ?
+    to a bids id sub-? and reciprocally."""
+
+    # todo : add test part
+
+    def validate(self, value: str) -> str:
+        if re.fullmatch(r"sub-IXI\d{3}", value):
+            return value
+        raise ValueError(
+            f"BIDS IXI subject ID {value} is not properly formatted. "
+            "Expecting a 'sub-IXIXXX' format."
+        )
+
+    @classmethod
+    def from_original_study_id(cls, study_id: str) -> str:
+        if re.fullmatch(r"IXI\d{3}", study_id):
+            return "sub-" + study_id
+            pass
+        raise ValueError(
+            f"Raw IXI subject ID {study_id} is not properly formatted. "
+            "Expecting a 'Y' format."
+        )
+
+    def to_original_study_id(self) -> str:
+        return str(self.replace("sub-", ""))
 
 
 # -- Methods for the clinical data --
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -299,8 +299,6 @@ class IXIBIDSSubjectID(BIDSSubjectID):
     """Implementation for IXI of the BIDSSubjectIDClass, allowing to go from the source id IXI###
     to a bids id sub-IXI### and reciprocally."""
 
-    # todo : add test part
-
     def validate(self, value: str) -> str:
         if re.fullmatch(r"sub-IXI\d{3}", value):
             return value
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -296,8 +296,8 @@ class HABSBIDSSubjectID(BIDSSubjectID):
 
 
 class IXIBIDSSubjectID(BIDSSubjectID):
-    """Implementation for IXI of the BIDSSubjectIDClass, allowing to go from the source id ?
-    to a bids id sub-? and reciprocally."""
+    """Implementation for IXI of the BIDSSubjectIDClass, allowing to go from the source id IXI###
+    to a bids id sub-IXI### and reciprocally."""
 
     # todo : add test part
 
@@ -312,8 +312,7 @@ class IXIBIDSSubjectID(BIDSSubjectID):
     @classmethod
     def from_original_study_id(cls, study_id: str) -> str:
         if re.fullmatch(r"IXI\d{3}", study_id):
-            return "sub-" + study_id
-            pass
+            return f"sub-{study_id}"
         raise ValueError(
             f"Raw IXI subject ID {study_id} is not properly formatted. "
             "Expecting a 'Y' format."
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -8,6 +8,7 @@ import numpy as np
 
 from clinica.iotools.bids_utils import write_modality_agnostic_files
 from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
+    check_modalities,
     define_participants,
     read_clinical_data,
     write_ixi_subject_data,
@@ -36,6 +37,7 @@ def convert(
 
     path_to_dataset = validate_input_path(path_to_dataset)
     bids_dir = validate_input_path(bids_dir, check_exist=False)
+    # todo : see if bids does not exist already at this place
     path_to_clinical = validate_input_path(path_to_clinical)
     if subjects:
         subjects = validate_input_path(subjects)
@@ -48,12 +50,14 @@ def convert(
 
     clinical_data = read_clinical_data(path_to_clinical)
     participants = define_participants(path_to_dataset, clinical_data, subjects)
+    check_modalities(data_directory=path_to_dataset, participants=participants)
+
     write_participants(
         bids_dir=bids_dir, clinical_data=clinical_data, participants=participants
     )
 
     for participant in participants:
-        cprint(f"Converting IXI subject {participant} to BIDS")
+        cprint(f"Converting IXI subject {participant} to BIDS", lvl="debug")
         write_ixi_subject_data(
             bids_dir=bids_dir, participant=participant, path_to_dataset=path_to_dataset
         )
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -48,7 +48,7 @@ def convert(
         )
 
     clinical_data = read_clinical_data(path_to_clinical)
-    participants = define_participants(path_to_dataset, clinical_data, subjects)
+    participants = define_participants(path_to_dataset, subjects)
     check_modalities(data_directory=path_to_dataset, participants=participants)
 
     write_participants(
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -37,7 +37,6 @@ def convert(
 
     path_to_dataset = validate_input_path(path_to_dataset)
     bids_dir = validate_input_path(bids_dir, check_exist=False)
-    # todo : see if bids does not exist already at this place
     path_to_clinical = validate_input_path(path_to_clinical)
     if subjects:
         subjects = validate_input_path(subjects)
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -1,4 +1,4 @@
-"""Convert OASIS dataset (https://sites.wustl.edu/oasisbrains/) to BIDS."""
+"""Convert IXI dataset (https://brain-development.org/ixi-dataset/) to BIDS."""
 
 from pathlib import Path
 from typing import Optional
@@ -9,10 +9,10 @@ import numpy as np
 from clinica.iotools.bids_utils import write_modality_agnostic_files
 from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
     define_participants,
-    read_ixi_clinical_data,
-    write_ixi_scans,
-    write_ixi_sessions,
+    read_clinical_data,
     write_ixi_subject_data,
+    write_scans,
+    write_sessions,
 )
 from clinica.utils.filemanip import UserProvidedPath
 
@@ -45,7 +45,7 @@ def convert(
             lvl="warning",
         )
 
-    clinical_data = read_ixi_clinical_data(path_to_clinical)
+    clinical_data = read_clinical_data(path_to_clinical)
     participants = define_participants(path_to_dataset, clinical_data, subjects)
 
     for participant in participants:
@@ -53,10 +53,10 @@ def convert(
         write_ixi_subject_data(
             bids_dir=bids_dir, participant=participant, path_to_dataset=path_to_dataset
         )
-        write_ixi_sessions(
+        write_sessions(
             bids_dir=bids_dir, participant=participant, clinical_data=clinical_data
         )
-        write_ixi_scans(bids_dir=bids_dir, participant=participant)
+        write_scans(bids_dir=bids_dir, participant=participant)
 
     readme_data = {
         "link": "https://brain-development.org/ixi-dataset/",
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -5,7 +5,8 @@ from typing import Optional
 
 import nibabel as nb
 import numpy as np
-from ixi_to_bids_utils import read_ixi_clinical_data
+from iotools.bids_utils import write_modality_agnostic_files
+from ixi_to_bids_utils import *
 
 from clinica.utils.filemanip import UserProvidedPath
 
@@ -36,8 +37,30 @@ def convert(
             lvl="warning",
         )
 
-    path_to_clinical = "/Users/alice.joubert/Downloads"  # todo : testing only here
     clinical_data = read_ixi_clinical_data(path_to_clinical)
 
-    # todo : convert here
+    subjects = subjects if subjects else get_subjects_list_from_data(path_to_dataset)
+    subjects = filter_subjects_list(subjects, clinical_data)
+
+    # todo : treat DTI different (load and merge)
+
+    image_data = get_img_data_df(path_to_dataset)
+
+    for subject in subjects:
+        write_subject(select_subject_data(image_data, subject), bids_dir)
+
+    readme_data = {
+        "link": "https://brain-development.org/ixi-dataset/",
+        "desc": (
+            "IXI is the nickname for the Information eXtraction from Images project, "
+            "which issued a dataset of nearly 600 images from healthy subjects. The MR"
+            "acquisition protocol includes T1,T2, PD weighted, MRA and diffusion-weighted"
+            "images. Three hospitals in London were involved in data collection."
+        ),
+    }
+
+    write_modality_agnostic_files(
+        study_name=StudyName.IXI, readme_data=readme_data, bids_dir=bids_dir
+    )
+
     cprint("Conversion to BIDS succeeded.", lvl="info")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -5,9 +5,9 @@ from typing import Optional
 
 import nibabel as nb
 import numpy as np
-from iotools.bids_utils import write_modality_agnostic_files
-from ixi_to_bids_utils import *
 
+from clinica.iotools.bids_utils import write_modality_agnostic_files
+from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import *
 from clinica.utils.filemanip import UserProvidedPath
 
 __all__ = ["convert"]
@@ -30,6 +30,8 @@ def convert(
     path_to_dataset = validate_input_path(path_to_dataset)
     bids_dir = validate_input_path(bids_dir, check_exist=False)
     path_to_clinical = validate_input_path(path_to_clinical)
+    if subjects:
+        subjects = validate_input_path(subjects)
 
     if n_procs != 1:
         cprint(
@@ -39,10 +41,7 @@ def convert(
 
     clinical_data = read_ixi_clinical_data(path_to_clinical)
 
-    subjects_to_filter = (
-        subjects if subjects else get_subjects_list_from_data(path_to_dataset)
-    )
-    participants = filter_subjects_list(subjects_to_filter, clinical_data)
+    participants = define_participants(path_to_dataset, clinical_data, subjects)
 
     for participant in participants:
         write_subject_no_dti(
@@ -60,8 +59,8 @@ def convert(
         ),
     }
 
-    write_modality_agnostic_files(
-        study_name=StudyName.IXI, readme_data=readme_data, bids_dir=bids_dir
-    )
+    # write_modality_agnostic_files(
+    #     study_name=StudyName.IXI, readme_data=readme_data, bids_dir=bids_dir
+    # )
 
-    cprint("Conversion to BIDS succeeded.", lvl="info")
+    cprint("Conversion to BIDS finished.", lvl="info")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -11,10 +11,10 @@ from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
     check_modalities,
     define_participants,
     read_clinical_data,
-    write_ixi_subject_data,
     write_participants,
     write_scans,
     write_sessions,
+    write_subject_data,
 )
 from clinica.utils.filemanip import UserProvidedPath
 
@@ -57,7 +57,7 @@ def convert(
 
     for participant in participants:
         cprint(f"Converting IXI subject {participant} to BIDS", lvl="debug")
-        write_ixi_subject_data(
+        write_subject_data(
             bids_dir=bids_dir, participant=participant, path_to_dataset=path_to_dataset
         )
         write_sessions(
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -36,6 +36,8 @@ def convert(
             lvl="warning",
         )
 
+    path_to_clinical = "/Users/alice.joubert/Downloads"  # todo : testing only here
     clinical_data = read_ixi_clinical_data(path_to_clinical)
+
     # todo : convert here
     cprint("Conversion to BIDS succeeded.", lvl="info")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -7,7 +7,13 @@ import nibabel as nb
 import numpy as np
 
 from clinica.iotools.bids_utils import write_modality_agnostic_files
-from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import *
+from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
+    define_participants,
+    read_ixi_clinical_data,
+    write_ixi_scans,
+    write_ixi_sessions,
+    write_ixi_subject_data,
+)
 from clinica.utils.filemanip import UserProvidedPath
 
 __all__ = ["convert"]
@@ -44,10 +50,9 @@ def convert(
 
     for participant in participants:
         cprint(f"Converting IXI subject {participant} to BIDS")
-        write_subject_no_dti(
-            select_subject_data(get_img_data_df(path_to_dataset), participant), bids_dir
+        write_ixi_subject_data(
+            bids_dir=bids_dir, participant=participant, path_to_dataset=path_to_dataset
         )
-        write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
         write_ixi_sessions(
             bids_dir=bids_dir, participant=participant, clinical_data=clinical_data
         )
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -11,6 +11,7 @@ from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
     define_participants,
     read_clinical_data,
     write_ixi_subject_data,
+    write_participants,
     write_scans,
     write_sessions,
 )
@@ -47,6 +48,9 @@ def convert(
 
     clinical_data = read_clinical_data(path_to_clinical)
     participants = define_participants(path_to_dataset, clinical_data, subjects)
+    write_participants(
+        bids_dir=bids_dir, clinical_data=clinical_data, participants=participants
+    )
 
     for participant in participants:
         cprint(f"Converting IXI subject {participant} to BIDS")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -40,14 +40,18 @@ def convert(
         )
 
     clinical_data = read_ixi_clinical_data(path_to_clinical)
-
     participants = define_participants(path_to_dataset, clinical_data, subjects)
 
     for participant in participants:
+        cprint(f"Converting IXI subject {participant} to BIDS")
         write_subject_no_dti(
             select_subject_data(get_img_data_df(path_to_dataset), participant), bids_dir
         )
         write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
+        write_ixi_sessions(
+            bids_dir=bids_dir, participant=participant, clinical_data=clinical_data
+        )
+        write_ixi_scans(bids_dir=bids_dir, participant=participant)
 
     readme_data = {
         "link": "https://brain-development.org/ixi-dataset/",
@@ -59,8 +63,8 @@ def convert(
         ),
     }
 
-    # write_modality_agnostic_files(
-    #     study_name=StudyName.IXI, readme_data=readme_data, bids_dir=bids_dir
-    # )
+    write_modality_agnostic_files(
+        study_name=StudyName.IXI, readme_data=readme_data, bids_dir=bids_dir
+    )
 
     cprint("Conversion to BIDS finished.", lvl="info")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -39,15 +39,16 @@ def convert(
 
     clinical_data = read_ixi_clinical_data(path_to_clinical)
 
-    subjects = subjects if subjects else get_subjects_list_from_data(path_to_dataset)
-    subjects = filter_subjects_list(subjects, clinical_data)
+    subjects_to_filter = (
+        subjects if subjects else get_subjects_list_from_data(path_to_dataset)
+    )
+    participants = filter_subjects_list(subjects_to_filter, clinical_data)
 
-    # todo : treat DTI different (load and merge)
-
-    image_data = get_img_data_df(path_to_dataset)
-
-    for subject in subjects:
-        write_subject(select_subject_data(image_data, subject), bids_dir)
+    for participant in participants:
+        write_subject_no_dti(
+            select_subject_data(get_img_data_df(path_to_dataset), participant), bids_dir
+        )
+        write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
 
     readme_data = {
         "link": "https://brain-development.org/ixi-dataset/",
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
@@ -1,0 +1,41 @@
+"""Convert OASIS dataset (https://sites.wustl.edu/oasisbrains/) to BIDS."""
+
+from pathlib import Path
+from typing import Optional
+
+import nibabel as nb
+import numpy as np
+from ixi_to_bids_utils import read_ixi_clinical_data
+
+from clinica.utils.filemanip import UserProvidedPath
+
+__all__ = ["convert"]
+
+
+def convert(
+    path_to_dataset: UserProvidedPath,
+    bids_dir: UserProvidedPath,
+    path_to_clinical: UserProvidedPath,
+    subjects: Optional[UserProvidedPath] = None,
+    n_procs: Optional[int] = 1,
+    **kwargs,
+):
+    from clinica.iotools.bids_utils import StudyName
+    from clinica.iotools.converters.factory import get_converter_name
+    from clinica.utils.stream import cprint
+
+    from ..utils import validate_input_path
+
+    path_to_dataset = validate_input_path(path_to_dataset)
+    bids_dir = validate_input_path(bids_dir, check_exist=False)
+    path_to_clinical = validate_input_path(path_to_clinical)
+
+    if n_procs != 1:
+        cprint(
+            f"{get_converter_name(StudyName.IXI)} converter does not support multiprocessing yet. n_procs set to 1.",
+            lvl="warning",
+        )
+
+    clinical_data = read_ixi_clinical_data(path_to_clinical)
+    # todo : convert here
+    cprint("Conversion to BIDS succeeded.", lvl="info")
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
@@ -17,6 +17,7 @@ def get_subjects_list_from_data(data_directory: Path) -> List[str]:
         dict.fromkeys(
             re.match(r"IXI\d{3}", path.name).group(0)
             for path in data_directory.rglob(pattern="IXI*.nii.gz")
+            if re.match(r"IXI\d{3}", path.name)
         )
     )
 
--- a/clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
+++ b/clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
@@ -7,9 +7,11 @@ from typing import List
 
 import nibabel as nib
 import pandas as pd
+from nilearn import plotting
 from nilearn.image import concat_imgs
 
 from clinica.iotools.bids_utils import StudyName, bids_id_factory
+from clinica.utils.stream import cprint
 
 
 def get_subjects_list_from_data(data_directory: Path) -> List[str]:
@@ -52,8 +54,9 @@ def rename_ixi_modalities(input_mod: str) -> str:
         return "angio"
     elif input_mod == "PD":
         return "PDw"
-    elif input_mod == "DWI":
-        return "dwi"
+    elif input_mod == "DTI":
+        # todo: doubt on this one
+        return "dti"
     else:
         raise ValueError(
             f"The modality {input_mod} is not recognized in the IXI dataset."
@@ -68,11 +71,16 @@ def define_magnetic_field(hospital: str) -> str:
 
 
 def get_img_data_df(data_directory: Path) -> pd.DataFrame:
-    # todo : works for all except DTI !!! (I mean we should drop it after this)
+    # todo : get all modalities other than DTI
     df = pd.DataFrame(
-        {"img_path": [path for path in data_directory.rglob(pattern="IXI*.nii.gz")]}
+        {
+            "img_path": [
+                path
+                for path in data_directory.rglob(pattern="IXI*.nii.gz")
+                if re.search(r"IXI\d{3}(-\w*){3}.nii.gz$", str(path))
+            ]
+        }
     )
-    # todo : regex pattern for 3 groups IXI\d{3}(-\w*){3}.nii.gz vs 4 groups IXI\d{3}(-\w*){4}.nii.gz
     df = (
         df.assign(img_name=lambda df: df.img_path.apply(lambda x: x.name))
         .assign(img_name_no_ext=lambda df: df.img_name.apply(lambda x: x.split(".")[0]))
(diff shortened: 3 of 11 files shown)


clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    bids_path: Path, subject: str, data_directory: Path
+) -> None:
+    """Processes DTI data if found for a subject"""
+    if dti_paths := find_subject_dti_data(data_directory, subject):
+        cprint(f"Converting modality DTI for subject {subject}.", lvl="debug")
+        dti_to_save = merge_dti(dti_paths)
+        bids_id = bids_id_factory(StudyName.IXI).from_original_study_id(subject)
+        data_path = bids_path / bids_id / "ses-M000" / "dwi"
+        data_path.mkdir(parents=True, exist_ok=True)
+        filename = f"{bids_id}_ses-M000_dwi"
```
NicolasGensollen, line 168: A bit more readable and robust (avoid the hardcoded '/'):

```python
filename = Path(f"{bids_id}_ses-M000_dwi")
dti_to_save.to_filename(data_path / filename.with_suffix(".nii.gz"))
...
write_ixi_json_subject(data_path / filename.with_suffix(".json"), ...)
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    return df
+
+
+def select_subject_data(data_df: pd.DataFrame, subject: str) -> pd.DataFrame:
+    """Select subject images to copy based on IXI id"""
+    return data_df[data_df["subject"] == subject]
+
+
+def bids_filename_from_ixi(img: pd.Series) -> str:
```
NicolasGensollen, line 125: Maybe something like `get_bids_filename_from_ixi` to start with an action verb.

Also, what is 'ixi' ? 

I would vote for something more explicit like `get_bids_filename_from_image_data` or `get_bids_filename_from_subject_row`... WDYT ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
```diff
@@ -0,0 +1,70 @@
+"""Convert OASIS dataset (https://sites.wustl.edu/oasisbrains/) to BIDS."""
+
+from pathlib import Path
+from typing import Optional
+
+import nibabel as nb
+import numpy as np
+
+from clinica.iotools.bids_utils import write_modality_agnostic_files
+from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import *
```
NicolasGensollen, line 10: Usually it is recommended to avoid this kind of * import.
Prefer either importing the module (and prefixing all objects with the module name):

```python
from clinica.iotools.converters.ixi_to_bids import ixi_to_bids_utils
....
ixi_to_bids_utils.define_participants(...)
```

, or importing explicitly what you need:

```python
from .ixi_to_bids_utils import (
    read_ixi_clinical_data,
    define_participants,
   ...
)
```

For simple cases where there is no name conflicts, I tend to use the second option.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+from typing import List, Optional
+
+import nibabel as nib
+import pandas as pd
+from nilearn.image import concat_imgs
+
+from clinica.iotools.bids_utils import StudyName, bids_id_factory
+from clinica.utils.stream import cprint
+
+
```
NicolasGensollen, line 14: Here it would be great to know what are the functions that are exposed to the outside and what are the internal private functions. You can prefix all private functions with "_", and list all exported objects in `__all__`.

Moreover, for exported functions, it would be nice to have proper docstrings with the description of parameters, and return values.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+        cprint("Loading a subjects list provided by the user...")
+        subjects_to_filter = subjs_list_path.read_text().splitlines()
+    else:
+        subjects_to_filter = get_subjects_list_from_data(data_directory)
+    return filter_subjects_list(subjects_to_filter, clinical_data)
+
+
+def read_ixi_clinical_data(clinical_data_path: Path) -> pd.DataFrame:
+    """Reads and formats IXI clinical data"""
+    # todo : what keys ?
```
NicolasGensollen, line 52: still relevant ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    using clinical data"""
+    if subjs_list_path:
+        cprint("Loading a subjects list provided by the user...")
+        subjects_to_filter = subjs_list_path.read_text().splitlines()
+    else:
+        subjects_to_filter = get_subjects_list_from_data(data_directory)
+    return filter_subjects_list(subjects_to_filter, clinical_data)
+
+
+def read_ixi_clinical_data(clinical_data_path: Path) -> pd.DataFrame:
```
NicolasGensollen, line 50: We are within the `ixi_to_bids` module, I would vote for `read_clinical_data`.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    clinical_data.drop("DATE_AVAILABLE", axis=1, inplace=True)
+    clinical_data.rename(lambda x: x.lower(), axis=1, inplace=True)
+    clinical_data["ixi_id"] = clinical_data.ixi_id.apply(
+        lambda x: "IXI" + "0" * (3 - len(str(x))) + str(x)
+    )
+    clinical_data["session_id"] = "ses-M000"
+    return clinical_data
+
+
+def rename_ixi_modalities(input_mod: str) -> str:
```
NicolasGensollen, line 64: same here, `rename_modalities` ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+def find_subject_dti_data(data_directory: Path, subject: str) -> List[Path]:
+    pattern = subject + r"(-\w*){4}.nii.gz$"
+    return [
+        path
+        for path in data_directory.rglob(pattern="IXI*.nii.gz")
+        if re.search(pattern, str(path))
+    ]
+
+
+def merge_dti(dti_images: List[Path]):
```
NicolasGensollen, line 187: The return type is missing here.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    return concat_imgs([nib.load(img) for img in dti_images])
+
+
+def write_ixi_scans(bids_dir: Path, participant: str):
```
NicolasGensollen, line 191: `write_scans` ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+                for path in (bids_dir / bids_id).rglob("*.nii.gz")
+            ]
+        }
+    )
+    to_write.to_csv(
+        f"{bids_dir}/{bids_id}/ses-M000/{bids_id}_ses-M000_scans.tsv", sep="\t"
+    )
+
+
+def write_ixi_sessions(bids_dir: Path, clinical_data: pd.DataFrame, participant: str):
```
NicolasGensollen, line 206: `write_sessions` ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    line = clinical_data[clinical_data["ixi_id"] == participant]
+    bids_id = bids_id_factory(StudyName.IXI).from_original_study_id(participant)
+    line.to_csv(f"{bids_dir}/{bids_id}/{bids_id}_sessions.tsv", sep="\t")
```
NicolasGensollen, line 209: More robust (avoids hardcoded '/'):

```suggestion
    line.to_csv(bids_dir / bids_id / f"{bids_id}_sessions.tsv", sep="\t")
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def define_magnetic_field(hospital: str) -> str:
+    if hospital == "Guys" or hospital == "IOP":
+        return "1.5"
+    if hospital == "HH":
+        return "3"
+
+
+def get_img_data_df(data_directory: Path) -> pd.DataFrame:
```
NicolasGensollen, line 88: You probably tired of me debating names 😅, but I think that we can afford the two missing letters to have a name easier to read : `get_image_data_df`. It doesn't sound like much but it contributes in reducing the cognitive load on the reader.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
```diff
@@ -0,0 +1,70 @@
...
+
+    clinical_data = read_ixi_clinical_data(path_to_clinical)
+    participants = define_participants(path_to_dataset, clinical_data, subjects)
+
+    for participant in participants:
+        cprint(f"Converting IXI subject {participant} to BIDS")
+        write_subject_no_dti(
+            select_subject_data(get_img_data_df(path_to_dataset), participant), bids_dir
+        )
+        write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
```
NicolasGensollen, line 50: These two functions are doing something very similar and take more or less the same parameters but in different orders.
I think this would be a bit simpler to have a single call to a function, for example:

```python
write_subject_data(bids_dir, participant, path_to_dataset)
```

And within the utils module:

```python
def write_subject_data(bids_dir: Path, participant, path_to_dataset: Path):
    write_subject_no_dti(
            select_subject_data(get_img_data_df(path_to_dataset), participant), bids_dir
        )
    write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
```

It also has the advantage of reducing the number of exported functions since functions like `select_subject_data` will probably become private with such a move.
WDYT ?

clinica/iotools/bids_utils.py
```diff
@@ -292,6 +295,34 @@ def to_original_study_id(self) -> str:
...
+        raise ValueError(
+            f"BIDS IXI subject ID {value} is not properly formatted. "
+            "Expecting a 'sub-IXIXXX' format."
+        )
+
+    @classmethod
+    def from_original_study_id(cls, study_id: str) -> str:
+        if re.fullmatch(r"IXI\d{3}", study_id):
+            return "sub-" + study_id
+            pass
```
NicolasGensollen, line 316:
```suggestion
```

clinica/iotools/bids_utils.py
```diff
@@ -292,6 +295,34 @@ def to_original_study_id(self) -> str:
...
+            return value
+        raise ValueError(
+            f"BIDS IXI subject ID {value} is not properly formatted. "
+            "Expecting a 'sub-IXIXXX' format."
+        )
+
+    @classmethod
+    def from_original_study_id(cls, study_id: str) -> str:
+        if re.fullmatch(r"IXI\d{3}", study_id):
+            return "sub-" + study_id
```
NicolasGensollen, line 315:
```suggestion
            return f"sub-{study_id}"
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
The code looks good and is easy to read which is great ! 👍 
I just made a quick first pass without testing anything. I made some comments and suggestions along the way. Nothing major, mostly naming and import related.
I don't believe we will do many changes to what you did, so you can probably go ahead and start adding unit tests for the functions you implemented.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids.py
AliceJoubert: You're right ! Thanks

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
AliceJoubert: It is but the question is badly worded : 

I am currently using the clinical data as is to put inside the sessions.tsv. However the keys of this dataframe are probably not BIDS 'compliant'. I am unsure of :
- what keys should actually go in the sessions.tsv
- under what name 
- if I should retrieve the info that is masked by identifiers (ex : male/female rather than 1/2 for sex)

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
AliceJoubert: I was not sure what to put here since it is a nifti image. What would you use ?

AliceJoubert:
Thanks for the review @NicolasGensollen ! I made the modifications you suggested. I have some questions regarding what I wrote :
- the question about sessions.tsv above 
- I am currently writing one json per image but they seem to contain the same info for one subject and it becomes very redundant when there are several modalities per folder (ex 'anat'). Should I keep as is or reduce to only one .json per type of modality ?
- I am not fully satisfied with the way I am treating the DTI modality compared to the others. I took the simplest way from my pov to handle all the other modalities but DTI but it creates inconsistencies. For instance, I warn the user if there is no DTI data but not if any other modality is missing for a subject. WDYT ?

AliceJoubert:
I also have interrogations regarding the way the subjects list is defined : 
- Currently it is possible to provide the converter a list of subjects with a .txt just like in ADNI
- I realized it is almost not tested (in ADNI as well): the only thing that is verified is the existence of the file provided, but it could be in another format than .txt and that could cause problems
- When reading the deployed documentation the option (-sl) actually does not appear at all even for ADNI unless I missed it ; it is only written in the help for 'clinica convert adni-to-bids' for ex. If I did not know better I could think I could provide a -tsv file as described in https://aramislab.paris.inria.fr/clinica/docs/public/latest/InteractingWithClinica/#common-options since it is not specified if it is for pipelines or converters. To sum up : the doc is unclear and I do not know if people even use the -sl option for ADNI (only converter where it is deployed).

WDYT ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
NicolasGensollen: Looking at the [specifications](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#sessions-file), I'm not sure we need a `sessions.tsv` file since there is only one session per subject, right ?
If I understand correctly, it still possible to have it. The required columns and their names are indicated in the specs, and it is possible to add other columns if needed. 
However, metadata like sex or age should go in the [participants.tsv](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#participants-file).

What columns do you have in the input dataframe ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
NicolasGensollen: You can do:

```python
import nibabel as nib

def func(...) -> nib.Nifti1Image:
   pass
```

NicolasGensollen:
> * I am currently writing one json per image but they seem to contain the same info for one subject and it becomes very redundant when there are several modalities per folder (ex 'anat'). Should I keep as is or reduce to only one .json per type of modality ?

The specifications say that we should use the [inheritance principle](https://bids-specification.readthedocs.io/en/stable/common-principles.html#the-inheritance-principle) to avoid the proliferation of file copies. I doubt that other converters are using this principle at the moment, but I believe we should rely on it more in the future.
Seems to me that here is the perfect use case since all json files are the same, right ?

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
AliceJoubert: Right, there is only one session per subject. That reminds me I forgot to write the participants.tsv file... 

I have sex, height, weight, ethnicity, marital situation, occupation, qualification, date of birth, age, study date

AliceJoubert:
> > * I am currently writing one json per image but they seem to contain the same info for one subject and it becomes very redundant when there are several modalities per folder (ex 'anat'). Should I keep as is or reduce to only one .json per type of modality ?
> 
> The specifications say that we should use the [inheritance principle](https://bids-specification.readthedocs.io/en/stable/common-principles.html#the-inheritance-principle) to avoid the proliferation of file copies. I doubt that other converters are using this principle at the moment, but I believe we should rely on it more in the future. Seems to me that here is the perfect use case since all json files are the same, right ?


Normally, all the json files are the same. I did not check for every subject but that would be easy to do. If I understand correctly, we could go from this (current version) :

```
├── README
├── dataset_description.json
└──  sub-IXI002
    ├── ses-M000
    │   ├── anat
    │   │   ├── sub-IXI002_ses-M000_PDw.json
    │   │   ├── sub-IXI002_ses-M000_PDw.nii.gz
    │   │   ├── sub-IXI002_ses-M000_T1w.json
    │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
    │   │   ├── sub-IXI002_ses-M000_T2w.json
    │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
    │   │   ├── sub-IXI002_ses-M000_angio.json
    │   │   └── sub-IXI002_ses-M000_angio.nii.gz
    │   ├── dwi
    │   │   ├── sub-IXI002_ses-M000_dwi.json
    │   │   └── sub-IXI002_ses-M000_dwi.nii.gz
    │   └── sub-IXI002_ses-M000_scans.tsv
    └── sub-IXI002_sessions.tsv
```

to this :

```
├── README
├── dataset_description.json
└──  sub-IXI002
    ├── ses-M000
    │   ├── anat
    │   │   ├── sub-IXI002_ses-M000_PDw.nii.gz
    │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
    │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
    │   │   └── sub-IXI002_ses-M000_angio.nii.gz
    │   ├── dwi
    │   │   └── sub-IXI002_ses-M000_dwi.nii.gz
    │   └── sub-IXI002_ses-M000_scans.tsv
    └── sub-IXI002_?.json
```

where the file `sub-IXI002_?.json` replaces the previous .json files. Reading the documentation, I am a bit unsure with what I could replace the `?` with : would `data_acquisition` be allowed ?

NicolasGensollen:
The way I understand the specifications is that, if all JSON files are the same for a given modality (let's say T1w for example), you can have a single `T1w.json` file at the root that should apply to alll T1w scans in the dataset (all subjects, all sessions, all other possible entities). It would look like this:

```
├── README
├── dataset_description.json
├── T1w.json
├── T2w.json
└──  sub-IXI001
    ├── ses-M000
    │   ├── anat
    │   │   ├── sub-IXI001_ses-M000_T1w.nii.gz
    │   │   ├── sub-IXI001_ses-M000_T2w.nii.gz
└──  sub-IXI002
    ├── ses-M000
    │   ├── anat
    │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
    │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
```

Does this make sense ? (I could be wrong, it's just the way I understand the specs...)

AliceJoubert:
> The way I understand the specifications is that, if all JSON files are the same for a given modality (let's say T1w for example), you can have a single `T1w.json` file at the root that should apply to alll T1w scans in the dataset (all subjects, all sessions, all other possible entities). It would look like this:
> 
> ```
> ├── README
> ├── dataset_description.json
> ├── T1w.json
> ├── T2w.json
> └──  sub-IXI001
>     ├── ses-M000
>     │   ├── anat
>     │   │   ├── sub-IXI001_ses-M000_T1w.nii.gz
>     │   │   ├── sub-IXI001_ses-M000_T2w.nii.gz
> └──  sub-IXI002
>     ├── ses-M000
>     │   ├── anat
>     │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
>     │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
> ```
> 
> Does this make sense ? (I could be wrong, it's just the way I understand the specs...)

It does make sense ! Though to clarify : the .jsons are the same for a subject, not between modalities (they indicate the hospital for ex). Hence the proposition I made, I can't make a json per modality

NicolasGensollen:
> > The way I understand the specifications is that, if all JSON files are the same for a given modality (let's say T1w for example), you can have a single `T1w.json` file at the root that should apply to alll T1w scans in the dataset (all subjects, all sessions, all other possible entities). It would look like this:
> > ```
> > ├── README
> > ├── dataset_description.json
> > ├── T1w.json
> > ├── T2w.json
> > └──  sub-IXI001
> >     ├── ses-M000
> >     │   ├── anat
> >     │   │   ├── sub-IXI001_ses-M000_T1w.nii.gz
> >     │   │   ├── sub-IXI001_ses-M000_T2w.nii.gz
> > └──  sub-IXI002
> >     ├── ses-M000
> >     │   ├── anat
> >     │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
> >     │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does this make sense ? (I could be wrong, it's just the way I understand the specs...)
> 
> It does make sense ! Though to clarify : the .jsons are the same for a subject, not between modalities (they indicate the hospital for ex). Hence the proposition I made, I can't make a json per modality

If some key-value pairs are the same for all subjects, you can have a first `T1w.json` at the root with these. Then you'd have another `T1w.json` within each subject folders defining the subject-specific key-value pairs only.
A BIDS reader would read these files top to bottom, updating the dictionaries as it gets closer to the scan.
Given that you'd have to write one file per subject anyway (and that you have one session per subject), it might not be worth it to apply the inheritance principle and leave your initial files... WDYT ?

AliceJoubert:
> > > The way I understand the specifications is that, if all JSON files are the same for a given modality (let's say T1w for example), you can have a single `T1w.json` file at the root that should apply to alll T1w scans in the dataset (all subjects, all sessions, all other possible entities). It would look like this:
> > > ```
> > > ├── README
> > > ├── dataset_description.json
> > > ├── T1w.json
> > > ├── T2w.json
> > > └──  sub-IXI001
> > >     ├── ses-M000
> > >     │   ├── anat
> > >     │   │   ├── sub-IXI001_ses-M000_T1w.nii.gz
> > >     │   │   ├── sub-IXI001_ses-M000_T2w.nii.gz
> > > └──  sub-IXI002
> > >     ├── ses-M000
> > >     │   ├── anat
> > >     │   │   ├── sub-IXI002_ses-M000_T1w.nii.gz
> > >     │   │   ├── sub-IXI002_ses-M000_T2w.nii.gz
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Does this make sense ? (I could be wrong, it's just the way I understand the specs...)
> > 
> > 
> > It does make sense ! Though to clarify : the .jsons are the same for a subject, not between modalities (they indicate the hospital for ex). Hence the proposition I made, I can't make a json per modality
> 
> If some key-value pairs are the same for all subjects, you can have a first `T1w.json` at the root with these. Then you'd have another `T1w.json` within each subject folders defining the subject-specific key-value pairs only. A BIDS reader would read these files top to bottom, updating the dictionaries as it gets closer to the scan. Given that you'd have to write one file per subject anyway (and that you have one session per subject), it might not be worth it to apply the inheritance principle and leave your initial files... WDYT ?

I would say the inheritance principle is not the most adapted here, but I still have some duplicate files that may not be necessary so I am not sure what to do with them. We can see all of that when we both come back !

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+        _get_marital_mapping(clinical_data_path)
+    )
+    clinical_data["OCCUPATION_ID"] = clinical_data["OCCUPATION_ID"].map(
+        _get_occupation_mapping(clinical_data_path)
+    )
+    clinical_data["QUALIFICATION_ID"] = clinical_data["QUALIFICATION_ID"].map(
+        _get_qualification_mapping(clinical_data_path)
+    )
+    clinical_data["IXI_ID"] = clinical_data.IXI_ID.apply(
+        lambda x: "IXI" + "0" * (3 - len(str(x))) + str(x)
```
NicolasGensollen, line 161: I would actually make this a real small helper function with a couple examples in the docstring and/or a few unit tests to help the reader understand what it is expecting and doing.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    clinical_data.rename(lambda x: _rename_clinical_data(x), axis=1, inplace=True)
+    clinical_data.fillna("n/a", inplace=True)
+    clinical_data["session_id"] = "ses-M000"
+    return clinical_data
+
+
+def _rename_modalities(input_mod: str) -> str:
+    if input_mod == "T1":
+        return "T1w"
+    elif input_mod == "T2":
```
NicolasGensollen, line 172: Minor suggestion: you can use "if" instead of "elif" and you don't need the last "else".

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    elif input_mod == "DTI":
+        return "dti"
+    else:
+        raise ValueError(
+            f"The modality {input_mod} is not recognized in the IXI dataset."
+        )
+
+
+def _define_magnetic_field(hospital: str) -> str:
+    if hospital == "Guys" or hospital == "IOP":
```
NicolasGensollen, line 187:
```suggestion
    if hospital in ("Guys", "IOP"):
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+        return "1.5"
+    elif hospital == "HH":
```
NicolasGensollen, line 189: same suggestion for elif and else

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+        .assign(session="ses-M000")
+    )
+    return df
+
+
+def _get_bids_filename_from_image_data(img: pd.Series) -> str:
+    return f"{img['participant_id']}_{img['session']}_{img['modality']}"
+
+
+def write_ixi_subject_data(
```
NicolasGensollen, line 231:
```suggestion
def write_subject_data(
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    bids_dir : Path to the output BIDS directory.
+    participant : Current converted subject study id (str).
+    path_to_dataset : Path to the raw dataset directory.
+    """
+    data_df = _get_img_data(path_to_dataset)
+    _write_subject_no_dti(data_df[data_df["subject"] == participant], bids_dir)
+    _write_subject_dti_if_exists(bids_dir, participant, path_to_dataset)
+
+
+def _write_ixi_json_image(writing_path: Path, hospital: str, field: str) -> None:
```
NicolasGensollen, line 248:
```suggestion
def _write_json_image(writing_path: Path, hospital: str, field: str) -> None:
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    for _, row in subject_df.iterrows():
+        cprint(
+            f"Converting modality {row['modality']} for subject {row['subject']}.",
+            lvl="debug",
+        )
+        filename = _get_bids_filename_from_image_data(row)
+        data_path = bids_path / row["participant_id"] / row["session"] / "anat"
+        data_path.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(row["img_path"], f"{data_path}/{filename}.nii.gz")
+        _write_ixi_json_image(
```
NicolasGensollen, line 280:
```suggestion
        _write_json_image(
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    if dti_paths := _find_subject_dti_data(data_directory, subject):
+        cprint(f"Converting modality DTI for subject {subject}.", lvl="debug")
+        dti_to_save = _merge_dti(dti_paths)
+        bids_id = bids_id_factory(StudyName.IXI).from_original_study_id(subject)
+        data_path = bids_path / bids_id / "ses-M000" / "dwi"
+        data_path.mkdir(parents=True, exist_ok=True)
+        filename = Path(f"{bids_id}_ses-M000_dwi")
+        dti_to_save.to_filename(data_path / filename.with_suffix(".nii.gz"))
+        hospital = dti_paths[0].name.split("-")[1]
+        _write_ixi_json_image(
```
NicolasGensollen, line 298:
```suggestion
        _write_json_image(
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    to_write = pd.DataFrame(
+        {
+            "filename": [
+                f"{path.parent.name}/{path.name}"
+                for path in (bids_dir / bids_id).rglob("*.nii.gz")
+            ]
+        }
+    )
+    to_write.to_csv(
+        bids_dir / bids_id / f"ses-M000/{bids_id}_ses-M000_scans.tsv", sep="\t"
```
NicolasGensollen, line 337:
```suggestion
        bids_dir / bids_id / "ses-M000" / f"{bids_id}_ses-M000_scans.tsv", sep="\t"
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    _get_subjects_list_from_data,
+    _get_subjects_list_from_file,
+    _rename_clinical_data,
+    _rename_modalities,
+    read_clinical_data,
+)
+
+
+def test_get_subjects_list_from_data(tmp_path):
+    for filename in ["IXI1", "IXI123", "IXIaaa", "foo"]:
```
NicolasGensollen, line 19:
```suggestion
    for filename in ("IXI1", "IXI123", "IXIaaa", "foo"):
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        ValueError,
+        match=f"The modality {input_str} is not recognized in the IXI dataset.",
+    ):
+        _rename_modalities(input_str)
+
+
+def test_get_subjects_list_from_file(tmp_path):
+    with open(tmp_path / "subjects.txt", "w") as f:
+        f.write("IXI123\nIXI001")
+    assert _get_subjects_list_from_file(Path(tmp_path / "subjects.txt")) == [
```
NicolasGensollen, line 50:
```suggestion
    assert _get_subjects_list_from_file(tmp_path / "subjects.txt") == [
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    ],
+)
+def test_define_magnetic_field_success(input_str, expected):
+    assert _define_magnetic_field(input_str) == expected
+
+
+def test_define_magnetic_field_success():
+    with pytest.raises(
+        ValueError,
+        match=f"The hospital {'foo'} was not recognized.",
```
NicolasGensollen, line 94:
```suggestion
        match="The hospital 'foo' was not recognized.",
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+        return "acq_time"
+    else:
+        return column.lower()
+
+
+def _get_sex_mapping() -> pd.DataFrame:
+    return pd.DataFrame({"SEX": ["male", "female"]}, index=[1, 2])["SEX"]
+
+
+class Clinical_Data_Mapping(str, Enum):
```
NicolasGensollen, line 101: Since this is a class:

```suggestion
class ClinicalDataMapping(str, Enum):
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+    ETHNIC = "Ethnicity"
+    MARITAL = "Marital Status"
+    OCCUPATION = "Occupation"
+    QUALIFICATION = "Qualification"
+
+
+def _get_mapping(clinical_data_path: Path, map: Clinical_Data_Mapping) -> pd.DataFrame:
```
NicolasGensollen, line 108:
```suggestion
def _get_mapping(clinical_data_path: Path, map: ClinicalDataMapping) -> pd.DataFrame:
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+    try:
+        df = pd.read_excel(
```
NicolasGensollen, line 110:
```suggestion
        return pd.read_excel(
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+            FileNotFoundError,
+        )
+    except (ValueError, KeyError):
+        log_and_raise(
+            f"{map.value} mapping is expected to be contained in a sheet called {map.value} coming from the clinical data excel. "
+            f"Possibilities are supposed to be described in a {map.name} column associated to keys from the 'ID' column. "
+            f"In case the file downloaded from the IXI website changed format, please do not hesitate to report to us !",
+            ValueError,
+        )
+    else:
```
NicolasGensollen, line 126:
```suggestion
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+        return df
```
NicolasGensollen, line 127:
```suggestion
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,449 @@
...
+    else:
+        clinical_data.drop("DATE_AVAILABLE", axis=1, inplace=True)
+        clinical_data["SEX_ID (1=m, 2=f)"] = clinical_data["SEX_ID (1=m, 2=f)"].map(
+            _get_sex_mapping()
+        )
+
+        for column in ("ETHNIC_ID", "MARITAL_ID", "OCCUPATION_ID", "QUALIFICATION_ID"):
+            clinical_data[column] = clinical_data[column].map(
+                _get_mapping(
+                    clinical_data_path, Clinical_Data_Mapping[column.replace("_ID", "")]
```
NicolasGensollen, line 170: You can use your enumeration here:

```python
for mapping in ClinicalDataMapping:
    clinical_data[f"{mapping.name}_ID"] = clinical_data[f"{mapping.name}_ID"].map(
        _get_mapping(clinical_data_path, mapping)
    )
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
+import json
+from asyncore import write
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+from clinica.iotools.converters.ixi_to_bids.ixi_to_bids_utils import (
+    Clinical_Data_Mapping,
```
NicolasGensollen, line 9:
```suggestion
    ClinicalDataMapping,
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+    read_clinical_data,
+    write_participants,
+    write_scans,
+    write_sessions,
+)
+
+
+def test_get_subjects_list_from_data(tmp_path):
+    for filename in ("IXI1", "IXI123", "IXIaaa", "foo"):
+        Path(f"{tmp_path}/{filename}_T1w.nii.gz").touch()
```
NicolasGensollen, line 34:
```suggestion
        (tmp_path / f"{filename}_T1w.nii.gz").touch()
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+        f.write("IXI123\nIXI001")
+    assert _get_subjects_list_from_file(tmp_path / "subjects.txt") == [
+        "IXI123",
+        "IXI001",
+    ]
+
+
+def test_define_participants_filter(tmp_path):
+    for filename in ("IXI001", "IXI002", "IXI003", "IXI004"):
+        Path(f"{tmp_path}/{filename}_T1w.nii.gz").touch()
```
NicolasGensollen, line 49:
```suggestion
        (tmp_path / f"{filename}_T1w.nii.gz").touch()
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+    with open(tmp_path / "subjects.txt", "w") as f:
+        f.write("IXI001\nIXI006")
+    assert define_participants(
+        data_directory=tmp_path, subjs_list_path=tmp_path / "subjects.txt"
+    ) == ["IXI001"]
+
+
+def test_define_participants_optional(tmp_path):
+    for filename in ("IXI001", "IXI002"):
+        Path(f"{tmp_path}/{filename}_T1w.nii.gz").touch()
```
NicolasGensollen, line 59:
```suggestion
        (tmp_path / f"{filename}_T1w.nii.gz").touch()
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+                "Single",
+                "Married",
+                "Divorced or Separated",
+                "Cohabiting",
+                "Widowed",
+            ],
+        }
+    )
+    marital.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Marital Status")
+    assert _get_mapping(tmp_path, Clinical_Data_Mapping.MARITAL).equals(
```
NicolasGensollen, line 208:
```suggestion
    assert _get_mapping(tmp_path, ClinicalDataMapping.MARITAL).equals(
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+                "White",
+                "Black or BlackBritish",
+                "Asian or Asian British",
+                "Chinese",
+                "Other",
+            ],
+        }
+    )
+    ethnic.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Ethnicity")
+    assert _get_mapping(tmp_path, Clinical_Data_Mapping.ETHNIC).equals(
```
NicolasGensollen, line 227:
```suggestion
    assert _get_mapping(tmp_path, ClinicalDataMapping.ETHNIC).equals(
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+                "Full-time housework",
+                "Retired",
+                "Unemployed",
+                "Work for pay at home",
+                "Other",
+            ],
+        }
+    )
+    occup.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Occupation")
+    assert _get_mapping(tmp_path, Clinical_Data_Mapping.OCCUPATION).equals(
```
NicolasGensollen, line 249:
```suggestion
    assert _get_mapping(tmp_path, ClinicalDataMapping.OCCUPATION).equals(
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+                "No qualifications",
+                "O - levels, GCSEs, or CSEs",
+                "A - levels",
+                "Further education e.g.City & Guilds / NVQs",
+                "University or Polytechnic degree",
+            ],
+        }
+    )
+    qualif.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Qualification")
+    assert _get_mapping(tmp_path, Clinical_Data_Mapping.QUALIFICATION).equals(
```
NicolasGensollen, line 268:
```suggestion
    assert _get_mapping(tmp_path, ClinicalDataMapping.QUALIFICATION).equals(
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+        }
+    )
+    qualif.to_excel(excel_writer=tmp_path / "IXI2.xls", sheet_name="Qualification")
+
+    with pytest.raises(
+        FileNotFoundError,
+        match=f"Clinical data stored in the folder {tmp_path} is expected to be an excel file named 'IXI.xls'. "
+        f"In case the file downloaded from the IXI website changed format, please do not hesitate to report to us !",
+    ):
+        _get_mapping(tmp_path, Clinical_Data_Mapping.QUALIFICATION)
```
NicolasGensollen, line 293:
```suggestion
        _get_mapping(tmp_path, ClinicalDataMapping.QUALIFICATION)
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+    )
+    qualif.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Qualification_error")
+
+    with pytest.raises(
+        ValueError,
+        match=f"Qualification mapping is expected to be contained in a sheet called Qualification coming from the clinical data excel. "
+        f"Possibilities are supposed to be described in a QUALIFICATION column associated to keys from the 'ID' column. "
+        f"In case the file downloaded from the IXI website changed format, please do not hesitate to report to us !",
+    ):
+        _get_mapping(tmp_path, Clinical_Data_Mapping.QUALIFICATION)
```
NicolasGensollen, line 317:
```suggestion
        _get_mapping(tmp_path, ClinicalDataMapping.QUALIFICATION)
```

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,454 @@
...
+    )
+    qualif.to_excel(excel_writer=tmp_path / "IXI.xls", sheet_name="Qualification")
+
+    with pytest.raises(
+        ValueError,
+        match=f"Qualification mapping is expected to be contained in a sheet called Qualification coming from the clinical data excel. "
+        f"Possibilities are supposed to be described in a QUALIFICATION column associated to keys from the 'ID' column. "
+        f"In case the file downloaded from the IXI website changed format, please do not hesitate to report to us !",
+    ):
+        _get_mapping(tmp_path, Clinical_Data_Mapping.QUALIFICATION)
```
NicolasGensollen, line 341:
```suggestion
        _get_mapping(tmp_path, ClinicalDataMapping.QUALIFICATION)
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
I made some very small suggestions but I think the PR is looking good and is almost ready. 

Since the PR is getting large, I think we could merge when unit tests are finished and take care of the remaining things in subsequent PRs:
- Adding non regression tests
- Adding documentation

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
```diff
@@ -0,0 +1,614 @@
+import json
+from asyncore import write
```
NicolasGensollen, line 2: This is breaking the unit test suite for Python 3.12 and I don't believe you added this intentionally. Looks more like a little joke from PyCharm... ;)

```suggestion
```

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
NicolasGensollen: I think this is still relevant

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
NicolasGensollen: same for these small Path usage suggestions

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
NicolasGensollen: These small stuffs are also still relevant

test/unittests/iotools/converters/ixi_to_bids/test_ixi_to_bids_utils.py
AliceJoubert: Yup I confirm !

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
```diff
@@ -0,0 +1,436 @@
...
+
+    Parameters
+    ----------
+    data_directory : Path to raw dataset.
+    participants : List of the subject ids of all participants
+
+    """
+    expected_modalities = _identify_expected_modalities(data_directory)
+    message = (
+        f"Modalities : {' , '.join(_rename_modalities(mod) for mod in expected_modalities)}"
```
NicolasGensollen, line 420: The unit tests are failing because this message is not deterministic: it depends on the order of `expected_modalities`, which in turn depends on the folder iteration logic in `_identify_expected_modalities`. Depending on the platform, this can vary.
A simple and quick solution is to sort the `expected_modalities` list before displaying the message.

clinica/iotools/converters/ixi_to_bids/ixi_to_bids_utils.py
NicolasGensollen: I think this is the last remaining point. It's not a major one though. Let me know what you want to do about it.

AliceJoubert: I just discovered the button in git to show unresolved conversations >< will apply this right away !

NicolasGensollen (approved):
LGTM, thanks a lot for this great addition @AliceJoubert  ! 🎉

PR #1244, opened 2024-07-26
[ENH] Add the possibility to use `ANTsPy` instead of `ANTs` for `T1Linear` and `FlairLinear`

NicolasGensollen:
This PR proposes to add support for [ANTsPy](https://antspyx.readthedocs.io/en/latest/index.html) in the `T1Linear` and `FlairLinear` pipelines. These two pipelines currently only depend on Clinica's Python dependencies as well as the third party `ANTs` software which can be problematic to install for some users.

If this PR is merged, it would mean that users without an installation of ANTs on their machine would be able to run these two pipelines if they have installed the Python core of Clinica.

The price to pay on our side is the addition of the [ANTsPy](https://antspyx.readthedocs.io/en/latest/index.html)  Python dependency to our list of dependencies.
I think this is something we can live with, and we could very well have it as an optional dependency such that users only interested on other functionalities would not need to install it.

I tried to propose an implementation that should give results close to the original way of running the pipeline, but **I think we should gather user inputs on whether the pipeline produce sensible results**.

For these reasons, I added a `--use-antspy` option to the CLI such that it is very easy to try out and the original way of running the pipeline is still the default.
Note that a warning is given to the users to say that the functionality is still experimental:

Capture d’écran 2024-07-26 à 12 18 42

The pipeline produces the expected files with correct names:

Capture d’écran 2024-07-26 à 12 24 47

This PR also updates the documentation of these two pipelines to expose this new functionality to the users (again, with a warning saying that it is experimental...):

Capture d’écran 2024-07-26 à 12 47 53 (2)

Finally, the PR adds some unit tests on the added functionalities.


AliceJoubert (approved):
LGTM ! I guess we can merge and see what users say overtime.

NicolasGensollen:
I agree, everything is marked as "experimental", so let's merge and see what users think.

PR #1250, opened 2024-08-05
[MAINT] Remove upper bound constraint on `cattrs` version

### LINKED ISSUE #1249: Incorrect assumption in `cattrs` version constraint
## The problem

We currently constrain `cattrs` to be `1.x.y`:

https://github.com/aramis-lab/clinica/blob/c8c64e02412fb5af13205f49d4281bc591f8f8cd/pyproject.toml#L57

Which is a fine strategy as it usually prevents us from breaking changes in brand new major releases.
The problem comes from the fact that `cattrs` moved from semantic versioning to a specific form of calendar versioning (YY-MINOR-MICRO) in the spring of 2022 (which is similar to the versioning of pip for example). Here is the history of their releases: https://pypi.org/project/cattrs/#history
This means that the version numbers of `cattrs` now look like this: `23.2.3`.

Our current constraint will therefore limit us with `cattrs` `1.10.0` for ever, which is the last version released using semantic versioning.

## Proposed solution 

I believe we should simply remove the upper bound constraint on `cattrs`.

NicolasGensollen:
Fixes #1249


AliceJoubert (approved):
LGTM @NicolasGensollen ! I kinda fail to see exactly what it could break in the future if a new release from cattrs modified greatly its functioning, do you know ? At least it did not break anything I'm currently using

NicolasGensollen:
> LGTM @NicolasGensollen ! I kinda fail to see exactly what it could break in the future if a new release from cattrs modified greatly its functioning, do you know ? At least it did not break anything I'm currently using

It is used in a few places to easily convert between structured data (classes with rich types) and unstructured data (basically dictionaries that can be serialized to JSON). This is mostly used for writing `dataset_description.json` files.
I noticed the version problem described in #1249 while working on #1158 as I had to deal with `cattrs` converters directly. It's very unlikely that I'd have spotted it otherwise.
I don't think we take a huge risk in not having an upper bound constraint on the version number as this tool should be pretty stable.
If they break their API, we should spot it through our tests anyway.

AliceJoubert:
Ok ! Thanks for the explanation

PR #1251, opened 2024-08-06
[ENH] Add function `log_and_warn`

NicolasGensollen:
Similar to #1238 but for warnings.


AliceJoubert (approved):
LGTM ! Thanks Nicolas

PR #1258, opened 2024-08-20
[ENH] Replace spm index tissue map mapping with enum

NicolasGensollen:
Small PR to improve the way SPM index tissues are represented and the way they are mapped to spm indices.
Instead of having a mutable mapping between integers and strings, this PR proposes to have an enumeration for the SPM tissues and a simple function doing the mapping from integer indices to the right variant. This function raises a proper `ValueError` if the index doesn't correspond to a tissue.
The PR also adds some basic unit tests.

### DIFF
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -341,45 +341,48 @@ def aggregator(func):
 
 
 @aggregator
-def t1_volume_native_tpm(tissue_number):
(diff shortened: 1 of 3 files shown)


AliceJoubert (approved):
LGTM, Thanks @NicolasGensollen ! I am taking notes for my own code ahah

PR #1288, opened 2024-09-09
[DOC] Refactor Third-Party documentation page

AliceJoubert:


docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
-    Starting from Clinica `v0.3.7`, the [**CAT12**](http://dbm.neuro.uni-jena.de/cat/) toolbox is no longer needed for the `t1-volume` and `pet-volume` pipelines.
-    For previous versions of Clinica, you will need to download the latest version of the toolbox [here](http://dbm.neuro.uni-jena.de/cat/index.html#DOWNLOAD) and follow the instructions to ensure that your `cat12` folder is located in your `spm/toolbox` folder.
+!!! danger "Provided command lines should be adapted"
+    **Please** do not copy/paste the provided commands without adapting them to your system and without testing them.
+    Most paths provided here require to be adapted to your system and depend on how you installed the software.
+    
+    A good approach is to verify that the different paths you want to assign to a variable exist first : 
 
-_*You only need to install ITK if you plan to perform partial volume correction using PETPVC._
+    :material-arrow-right: If they do, then try running the `export` and `source` **(1.)** commands in your terminal. Verify that the software or the pipeline you want runs as expected.
```
NicolasGensollen, line 26: The `:material-arrow-right:` and `:material-arrow-right:` do not seem to render on the served html version.

docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
+    
+    :material-arrow-right: If this works, then consider modifying your shell configuration file **(2.)** to have these variables automatically defined on every session.
```
NicolasGensollen, line 28: I'm not sure to understand the `**(1.)**` and `**(2.)** `. Why are they in the middle of the sentence ?

docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
 
-Depending on the architecture and OS of your system, setup of third party libraries can change.
-Please refer to each tool’s website for installation instructions:
 
-### Environment variables setup
+## Converters
```
NicolasGensollen, line 31: Now, that I see that, I think we should have both a `Converters` section, and within this section, a `dcm2niix` section. This would match the structure of the pipelines section and will allow us to link to dcm2niix rather than "converters":

 ```markdown
## Converters

bla bla

### dcm2niix

bla bla
```

WDYT ?

File: docs/Third-party.md [hunk 4]
```diff
@@ -1,281 +1,268 @@
...
 
-When installing some of the third party software, environment variables might be needed in order to configure some installations.
+Some converters require **dcm2niix** to transform DICOM files into NIfTI :
 
-If you simply run the `export` and `source` commands in your terminal, these environment variables will be defined only for the duration of your session, and will be lost if you re-launch your terminal.
+<div class="grid">
+  <a href="Converters/ADNI2BIDS.md" class="card">adni-to-bids</a>
```
NicolasGensollen, line 36: These links aren't working. I think you need to provide the path to the html page, not the markdown file.

docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
 
-#### Configuration
+<div markdown="1" class="third_party_table">
 
-We strongly recommend installing `ANTs >= 2.5.0` from which **no environment variable are needed**.
+|                      | ANTs | Convert3D | FreeSurfer | FSL | ITK | Matlab | MRtrix3 | PETPVC | SPM |
+|:--------------------:|:----:|:---------:|:----------:|:---:|:---:|:------:|:-------:|:------:|:---:|
+|    Anat > Volume     |      |           |            |     |     |   ✓    |         |        |  ✓  |
+|  Anat > FreeSurfer   |      |           |     ✓      |     |     |        |         |        |     |
+| DWI > Preprocessing  |  ✓   |     ✓     |            |  ✓  |     |        |    ✓    |        |     |
```
NicolasGensollen, line 64: I think we are missing some pipelines here... For example, I don't see t1-linear / flair-linear

docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
+two options to install it :
+
+- [Use pre-built binaries](http://www.itksnap.org/pmwiki/pmwiki.php?n=Downloads.C3D).
+- [Use the official conda package](https://anaconda.org/conda-forge/convert3d).
 
-#### Installation
+___
 
-##### On Linux
+### FreeSurfer
```
NicolasGensollen, line 110: I realize that the softwares are listed in a totally random order. 
I suggest we use alphabetical order. WDYT ?

docs/Third-party.md
```diff
@@ -1,281 +1,268 @@
...
-    Please do not copy/paste the provided commands without adapting them to your system and without testing them.
-    Most paths provided here require to be adapted to your system, and also depends on how you installed the software.
-    A good approach is to verify that the different paths you want to assign to a variable exist first.
-    If they do, then try running the `export` and `source` commands in your terminal, and verify that the software run as expected.
-    You can also verify that the pipeline you want to use is also running as expected.
-    If this works, then consider modifying your shell configuration file to have these variables automatically defined on every session.
 
 
-### ANTs
+## Pipeline-specific interfaces
```
NicolasGensollen, line 49: Maybe we can simplify this to "Pipelines", which would make sense since we have a "Converters" section juste above. WDYT ?

NicolasGensollen (commented):
Thanks @AliceJoubert !
I think this is looking really good. 👍 
Definitely a nice improvement to the current page. 🚀

docs/Third-party.md
AliceJoubert: Thanks for spotting it !

docs/Third-party.md
AliceJoubert: I wanted to refer to the two steps described above but that's obviously not clear. WDYT now ?

docs/Third-party.md
AliceJoubert: You're right. I am now using https://aramislab.paris.inria.fr/clinica/docs/public/dev/ (+ page) as it is described in the mkdocs.yml as the site_url, WDYT ?

NicolasGensollen: Hmmm, this is the URL of the development version of the documentation. This will cause some issues because old versions of the doc will, at some point in the future, point to dev versions for example. Also, hardcoding the URL in the code isn't great. If you need to change it, you have to change it all over the code base.
I think you should use a relative path to the html file, not the URL of a deployed version of the doc. Does this make sense ?

docs/Third-party.md
NicolasGensollen: OK, I see. I would be explicit to avoid confusion. Maybe something like `see point **(1.)** above` ?

docs/Third-party.md
NicolasGensollen: Yes, they are... I got tricked by the diff 😅 
Sorry for the noise !

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert  !

PR #1297, opened 2024-09-16
[DOC] Adding Support page

AliceJoubert:
- Support.md as snippet
- Contributing.md included in docs


README.md
```diff
@@ -163,7 +163,7 @@ with Alzheimer's disease and healthy controls from the ADNI database:
 ## Contributing
 
 We encourage you to contribute to Clinica!
-Please check out the [Contributing to Clinica guide](CONTRIBUTING.md) for
+Please check out the [Contributing to Clinica guide](docs/Software/CONTRIBUTING.md) for
```
NicolasGensollen, line 166: I know this can feel a bit strange, but most Python softwares (for example scikit-learn, PyTorch, ...) have the contribution guidelines in a `CONTRIBUTING.md` file at the root of the project rather than somewhere in the docs.
I don't know if it is the initial reason, but it makes it easier to contribute to projects when you browse them on GitHub/GitLab because the file is easily spotted.
Anyway, I think we should stick with this convention.

docs/Converters/ADNI2BIDS.md
```diff
@@ -16,6 +16,9 @@
     | Duration / Start date   | 5 years / October 2004 | 2 years / September 2009 | 5 years / September 2011 | 5 years / September 2016 |
     | Cohort                  | 200 elderly controls + 400 MCI + 200 AD | Existing ADNI-1 + 200 early MCI | Existing ADNI-1 and ADNI-GO + 150 elderly controls + 100 early MCI + 150 late MCI + 150 AD | Existing ADNI-1, ADNI-GO, ADNI-2 + 133 elderly controls + 151 MCI + 87 AD |
 
+!!! warning "Supported ADNI versions"
+    Clinica is currently able to convert ADNI 1, 2 and 3. It is not ready for ADNI 4.
```
NicolasGensollen, line 20: What about ADNI GO ?

docs/Software/CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-# Contributing to this repo
+# Contributing to Clinica repository
```
NicolasGensollen, line 1:
```suggestion
# Contributing to Clinica
```

mkdocs.yml
```diff
@@ -83,6 +84,7 @@ nav:
     - Installation: Software/Installation.md
     - Third-party software: Software/Third-party.md
     - Interacting with Clinica: Software/InteractingWithClinica.md
+    - Contributing to the Github repository : Software/CONTRIBUTING.md
```
NicolasGensollen, line 87: It totally makes sense to have this section here.
The usual trick to also satisfy what I wrote in my first comment is to link or include the content of the `CONTRIBUTING.md` file at the root. Let me know if this isn't clear.

NicolasGensollen, line 87:
```suggestion
    - Contributing to Clinica : Software/CONTRIBUTING.md
```

docs/Software/InteractingWithClinica.md
```diff
@@ -133,9 +133,3 @@ You might also use `CNvsAD`, for instance, as group ID for a statistical group c
 --8<-- "snippets/converters_options.md"
 
 --8<-- "snippets/known_issues.md:matlab"
-
-
-## Support
```
NicolasGensollen, line 138: Shouldn't we link the corresponding snippet or am I missing something ?

NicolasGensollen (commented):
Thanks for pursuing this very useful refactoring @AliceJoubert !

README.md
AliceJoubert: Okay !

docs/Software/InteractingWithClinica.md
AliceJoubert: I added it to the list of automatically added snippets in the mkdocs.yml just like abbreviations.md so you get it everywhere :-) ! It is a bit overkill but you cannot miss it that way.

mkdocs.yml
AliceJoubert: I am not sure this is clear but I am testing a way to keep it where it should to follow the convention and add the content as a snippet there !

AliceJoubert: I did not know of this solution, thanks :) !!

AliceJoubert: From what I have tested : 
- Mkdocs does not like symlinks outside its base repository (docs) : it's the current commit which can't be built apparently
- I tried to use snippets but since we set the base_path at docs too we cannot use it either

I could use a hard link like https://github.com/aramis-lab/clinica/blob/dev/README.md but you might not like this since it would be hardcoded with the dev version. WDYT ? @NicolasGensollen

docs/Software/contributing.md
```diff
@@ -0,0 +1 @@
+../CONTRIBUTING.md
```
NicolasGensollen, line 1: I think this symlink should be ../../CONTRIBUTING.md, right ?

mkdocs.yml
```diff
@@ -78,11 +80,12 @@ not_in_nav: |
 
 # Page tree
 nav:
-  - Home: index.md
+  - Home: Home/index.md
```
NicolasGensollen, line 83: I don't understand why you made this change

docs/Software/InteractingWithClinica.md
NicolasGensollen: OK, sounds good ! 👍

docs/Software/contributing.md
AliceJoubert: You're right

mkdocs.yml
AliceJoubert: In my last PR I introduced snippets (parts of code to use in different places) for the list of pipelines and converters in Clinica. This way, it only needs to be updated in one place and can be re-used as needed anywhere in the documentation. The thing is, it contains relative links to the different pages (ex : [ADNI-to-BIDS] ( ../Converters/ADNI2BIDS.md ) ) so if I want to use the same snippet I need to make sure the different files where it is used is placed at the same relative "level" compared to the links it points to.

These snippets with links were used in the .md files from the docs/Software repository and the docs/index.md. The easiest way I found to keep these (because it seems easy to maintain) is to place the index.md in another folder.

docs/Home/index.md
```diff
@@ -3,17 +3,17 @@
...
 Clinica can be installed on **MacOS** and **Linux** (CentOS or Debian/Ubuntu) machines,
 and possibly on Windows computers with a Linux Virtual Machine.
 We assume that users installing and using Clinica are comfortable using the command line.
 
-- [Installation](Software/Installation.md)
-- [Third-party software](Software/Third-party.md)  
-- [Interacting with Clinica](Software/InteractingWithClinica.md)
+- [Installation](../Software/Installation.md)
+- [Third-party software](../Software/Third-party.md)  
+- [Interacting with Clinica](../Software/InteractingWithClinica.md)
```
NicolasGensollen, line 16: Is it on purpose that you didn't add a bullet point for "contributing to Clinica" ?

mkdocs.yml
NicolasGensollen: Alright ! Thanks for the explanations 👍

mkdocs.yml
```diff
@@ -124,15 +127,20 @@ nav:
     - OASIS-3 to BIDS: Converters/OASIS3TOBIDS.md
     - UK Biobank to BIDS: Converters/UKBtoBIDS.md
   - I/O tools:
+    - create-subjects-visits: IO#create-subjects-visits-generate-the-list-all-subjects-and-visits-of-a-given-dataset
+    - check-missing-modalities: IO#check-missing-modalities-check-missing-modalities-for-each-subject
+    - check-missing-processing: IO#check-missing-processing-check-missing-processing-in-a-caps-directory
```
NicolasGensollen, line 132: I think you need to remove some of these as there are duplicates atm

NicolasGensollen (commented):
Thanks @AliceJoubert !
I've tested the rendered version a bit and it seems to work fine.
I just noticed a couple small bugs (see comments).
Also, it looks like the link on the home page "Data handling tools for BIDS and CAPS compliant datasets" is broken atm.

docs/Home/index.md
AliceJoubert: No, I forgot !

AliceJoubert:
Thanks for the review @NicolasGensollen ! Normally I fixed everything. Do you see anything else ?

NicolasGensollen:
> Thanks for the review @NicolasGensollen ! Normally I fixed everything. Do you see anything else ?

Great ! Thanks @AliceJoubert !
The link under "clinica at conferences" seems to be broken as well

NicolasGensollen:
And the one above also ("t1-freesurfer....")

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1301, opened 2024-09-26
[CI] Use rsync instead of scp for deploying doc

NicolasGensollen:


AliceJoubert (approved):
LGTM !! Thanks @NicolasGensollen

PR #1311, opened 2024-10-02
[REF] Refactor and clean `utils.input_files` module

### LINKED ISSUE #1308: Improvements to module `input_files`
The module `clinica.utils.input_files` defines a long list of dictionaries which are used as query patterns in the `clinica_file_reader` functions.

First of all, there are a lot of these objects which are not used in the code base. A few examples:

```
T2W_LINEAR
FLAIR_T2W_LINEAR
T1W_EXTENSIVE
FLAIR_T2W_LINEAR_CROPPED
```

We clearly should get rid of them.

All these patterns could also be better represented with dataclasses since they have a defined structure. Something like that would already enrich the code:

```python
@dataclass
class Pattern:
    pattern: str
    description: str
    needed_pipeline: str
```

Also, a lot of these patterns could be factorized in functions taking some parameters (the hemisphere, the name of the atlas, the pet tracer...) instead of being copy-pasted.

For example this 

https://github.com/aramis-lab/clinica/blob/954d4a950aa341acf54059c1e93b27c9de18fbc2/clinica/utils/input_files.py#L51-L61

could easily be factorized in:

```python
def get_t1_freesurfer_longitudinal_white_matter_surface_pattern(
    hemisphere: Union[str, HemiSphere],
) -> Pattern:
    hemisphere = HemiSphere(hemisphere)
    return Pattern(
        f"t1/long-*/freesurfer_longitudinal/sub-*_ses-*.long.sub-*_*/surf/{hemisphere.value}.white",
        (
            f"{'right' if hemisphere == HemiSphere.RIGHT else 'left'} white matter/gray matter border "
            f"surface ({hemisphere.value}.white) generated with t1-freesurfer-longitudinal."
        ),
        "t1-freesurfer and t1-freesurfer longitudinal"
    )
```

NicolasGensollen:
Closes #1308

### DIFF
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -303,7 +303,7 @@ def get_dwi_preprocessed_brainmask(*args, **kwargs) -> QueryPattern:
 def get_dwi_fmap_phasediff_file(filetype: Union[str, DWIFileType]) -> QueryPattern:
     filetype = DWIFileType(filetype)
     return QueryPattern(
-        f"fmap/sub-*_ses-*_phasediff.{filetype.value}",
+        f"fmap/sub-*_ses-*_phasediff.{filetype.value}*",
         f"phasediff {filetype.value} file",
         "",
     )
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -5,197 +5,335 @@ These dictionaries describe files to grab.
 
 import functools
 from collections.abc import Iterable
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
 from typing import Optional, Union
 
 from clinica.utils.dwi import DTIBasedMeasure
+from clinica.utils.image import HemiSphere
 from clinica.utils.pet import ReconstructionMethod, SUVRReferenceRegion, Tracer
 
-# BIDS
+__all__ = [
+    "Query",
+    "QueryName",
+    "query_factory",
+]
 
-T1W_NII = {"pattern": "sub-*_ses-*_t1w.nii*", "description": "T1w MRI"}
-Flair_T2W_NII = {"pattern": "sub-*_ses-*_flair.nii*", "description": "FLAIR T2w MRI"}
 
-# T1-FreeSurfer
+@dataclass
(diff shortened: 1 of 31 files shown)


NicolasGensollen:
@AliceJoubert I think this is ready for review.

To be honest, I'm not fully convinced by the factory pattern I ended up using. I feel that it is a bit too complex compared to directly importing the pattern getter functions.

One of the main issue I see with my proposal is that when you do

```python
pattern = query_pattern_factory(QueryPatternName.T1W_LINEAR)(...)
```

The IDE is not able to tell you the expected arguments of the pattern getter that the factory is returning. So you basically have to find it and check its signature to know about it, which isn't great...

Also, considering our hackathon project, this code will be moved out of Clinica soonish and is likely to be refactored further. Having a dedicated library is one more reason to have the pattern getters as public functions of the API, since users might want to check their doc page.

Let me know what you think.

clinica/utils/input_files.py
```diff
@@ -5,197 +5,445 @@
...
+        (
+            f"{'left' if hemisphere == HemiSphere.LEFT else 'right'} hemisphere surface-based "
+            f"{parcellation.value} parcellation (label/{_get_annot_file_name(hemisphere, parcellation)}) "
+            "generated with t1-freesurfer-longitudinal."
+        ),
+        "t1-freesurfer and t1-freesurfer-longitudinal",
+    )
 
 
 def aggregator(func):
```
AliceJoubert, line 449: The description for this function needs to be changed a bit since `t1_volume_native_tpm` was removed. I cannot ping you at the right lines with the diff but all the examples here use `t1_volume_native_tpm`.

AliceJoubert (commented):
Thanks for the work @NicolasGensollen ! I agree with you, it's important to not be obligated to search in the code for the input types.

Maybe with the hackathon we will decide on a structure that will be easier to factorize and use in a factory instead of importing all the functions. Anyways, it will be redefined at that moment so it is fine if this PR is not optimal.

I would vote for importing the functions right away and not use the factory :pray:

PR #1313, opened 2024-10-07
[TEST][REF] Move functions specific to OASIS1 converter in a dedicated utils file and add unit tests

AliceJoubert:
- Move to module specific to OASIS1
- Remove parts specific to OASIS3 (never used)
- Add tests.


test/unittests/iotools/converters/oasis_to_bids/test_oasis_to_bids_utils.py
```diff
@@ -0,0 +1,172 @@
...
+        tmp_path / "clinical",
+        tmp_path / "BIDS",
+        tmp_path / "spec",
+        ["sub-OASIS10001", "sub-OASIS10002"],
+    )
+    write_sessions_tsv(tmp_path / "BIDS", sessions)
+    sessions_files = list((tmp_path / "BIDS").rglob("*.tsv"))
+    assert len(sessions_files) == 2
+    for file in sessions_files:
+        assert not assert_frame_equal(
```
NicolasGensollen, line 164: This is weird. What are you trying to assert here ?
I would expect either `assert <boolean condition>` or `assert_frame_equal(df1, df2)`.

test/unittests/iotools/converters/oasis_to_bids/test_oasis_to_bids_utils.py
```diff
@@ -0,0 +1,172 @@
...
+        }
+    )
+    spec.to_csv(tmp_path / "spec" / "sessions.tsv", index=False, sep="\t")
+
+
+@pytest.fixture
+def build_bids_dir(tmp_path: Path) -> None:
+    (tmp_path / "BIDS" / "sub-OASIS10001" / "ses-M000").mkdir(parents=True)
+    (tmp_path / "BIDS" / "sub-OASIS10001" / "ses-M006").mkdir(parents=True)
+    (tmp_path / "BIDS" / "sub-OASIS10002" / "ses-M000").mkdir(parents=True)
```
NicolasGensollen, line 77: Your code works but is a bit surprising in the way you use fixtures.
Nothing wrong, but I would expect something more or less like this:

```python
@pytest.fixture
def bids_dir(tmp_path: Path) -> Path:
    bids_dir = tmp_path / "BIDS"
    _build_bids_dir(bids_dir)
    return bids_dir

def _build_bids_dir(bids_dir: Path) -> None:
    (bids_dir / "sub...").mkdir(....)
    ...

def test_create_bla_bla_bla(tmp_path, bids_dir: Path, ...):
    result = create_sessions_dict(clinical_data_dir, bids_dir, ...)
    assert result == ...
```

Does this make sense ? WDYT ?

test/unittests/iotools/converters/oasis_to_bids/test_oasis_to_bids_utils.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@pytest.fixture
+def get_expected_dict() -> dict:
```
NicolasGensollen, line 81: Again, this works fine, but I would simplify by either having a function or a fixture.

That is, either:

```python
def get_expected_dict() -> dict:
    return {...}

def test_bla_bla(tmp_path, ...):
    ...
   assert result == get_expected_dict()
```

Or:

```python
@pytest.fixture
def expected() -> dict:
    return {...}

def test_bla_bla(tmp_path, expected: dict, ...):
    ...
    assert result == expected
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
The code looks good. I made a couple suggestions on the tests in order to improve readability and use fixtures in a more "traditional" way.

test/unittests/iotools/converters/oasis_to_bids/test_oasis_to_bids_utils.py
NicolasGensollen: Maybe something like

```python
assert not df1.equals(df2)
```

AliceJoubert:
Thanks for the review @NicolasGensollen , do you see anything else ?

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1334, opened 2024-10-14
[MAINT] Add python version classifiers to `pyproject.toml`

NicolasGensollen:


NicolasGensollen:
You're right @AliceJoubert ! The python classifiers are automatically added by Poetry (as visible on Pypi). I totally forgot about that when looking at the pyproject.toml file yesterday...
Sorry for the noise.

PR #1340, opened 2024-10-17
[FIX] Fix logging in OASIS and AIBL converters

AliceJoubert:
Same fix as #1181


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1342, opened 2024-10-21
[FIX] ADNI-to-BIDS : incorporate the fix for KeyError "APGEN"

### LINKED ISSUE #1294: Key Error : APGEN1
Hello, 

I used Clinica to convert ADNI to BIDS and I have a pandas error : "KeyError : APGEN1" 

This error concern adni_to_bids.convert_clinical_data with participants_df = bids.create_participants_df() and more precisely value_to_read = file_to_read[participant_field_db[i]]

It's when participants.tsv are created and no tsv are generated in my case. I just download all ADNI data between ADNI 1 to ADNI 4 (on July 2024) et just run clinica without options (for all datas). 

Do you have an idea why I have this error ? I suppose that one key isn't here in clinical data for ADNI but if we have a larger problem, I would like to submit it. 

Thank you for your help !

AliceJoubert:
Closes #1294

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -405,8 +405,7 @@ def create_participants_df(
                     # See issue https://github.com/aramis-lab/clinica/issues/1294
                     if study_name == StudyName.ADNI and location == "APOERES.csv":
                         if (
-                            "APGEN1" not in file_to_read.columns
-                            or "APGEN2" not in file_to_read.columns
+                            participant_fields_db[i] not in file_to_read.columns
                             and "GENOTYPE" in file_to_read.columns
                         ):
                             # Split the 'GENOTYPE' column into 'APGEN1' and 'APGEN2'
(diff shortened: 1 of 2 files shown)


clinica/iotools/bids_utils.py
```diff
@@ -401,6 +401,17 @@ def create_participants_df(
                     file_to_read = load_clinical_csv(
                         clinical_data_dir, location.split(".")[0]
                     )
+                    if location == "APOERES.csv":
```
NicolasGensollen, line 404: This is only for adni-to-bids right ? Maybe add something like:

```python
if study_name == StudyName.ADNI and location == "APOERES.csv":
    ...
```

clinica/iotools/bids_utils.py
```diff
@@ -401,6 +401,17 @@ def create_participants_df(
...
+                        if (
```
NicolasGensollen, line 405: Maybe add a small comment with the link to the issue

NicolasGensollen (commented):
Thanks @AliceJoubert !

clinica/iotools/bids_utils.py
AliceJoubert: Yes this is for ADNI only ; the file is specific to ADNI too

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1343, opened 2024-10-22
[FIX] ADNI utils `load_clinical_csv` should avoid hidden files

### LINKED ISSUE #1163: Ignore hidden CSV files in `load_clinical_csv`

AliceJoubert:
Close #1163

### DIFF
--- a/clinica/iotools/converters/adni_to_bids/adni_utils.py
+++ b/clinica/iotools/converters/adni_to_bids/adni_utils.py
@@ -1109,7 +1109,7 @@ def load_clinical_csv(clinical_dir: Path, filename: str) -> pd.DataFrame:
 
     pattern = filename + r"(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
     files_matching_pattern = [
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1347, opened 2024-10-23
[TEST] AIBL-to-BIDS : add unit tests to functions related to creating sessions.tsv files

### LINKED ISSUE #1349: [TEST] AIBL-to-BIDS utils are lacking unit tests for `sessions.tsv` creation
The function creating sessions files for AIBL: 
https://github.com/aramis-lab/clinica/blob/48c12ab0607638bf8b2fb755c06627c6f7f19a79/clinica/iotools/converters/aibl_to_bids/utils/clinical.py#L151-L155

is currently not tested. The utility functions used in it as well.

AliceJoubert:
Close #1349 

!!! Function modifications for `_compute_ages_at_each_exam` and `_compute_exam_date_from_baseline`

### DIFF
--- a/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
+++ b/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
@@ -276,7 +276,7 @@ def _clean_exam_dates(
     exam_dates: List[Union[str, int]],
     visit_codes: List[str],
     clinical_data_dir: Path,
-) -> List[str]:
+) -> List[Union[str, int]]:
     """Clean the exam dates when necessary by trying to compute them from other sources.
 
     Parameters
--- a/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
+++ b/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
@@ -158,7 +158,7 @@ def create_sessions_tsv_file(
     Parameters
     ----------
     input_path : Path
-        The path to the input folder.
+        The path to the input folder (BIDS directory).
 
     clinical_data_dir : Path
         The path to the directory to the clinical data files.
@@ -231,7 +231,13 @@ def create_sessions_tsv_file(
         exam_dates = _clean_exam_dates(
             rid, exam_date.to_list(), visit_code.to_list(), clinical_data_dir
         )
-        age = _compute_ages_at_each_exam(patient_date_of_birth.values[0], exam_dates)
+
+        if not patient_date_of_birth.empty:
+            age = _compute_ages_at_each_exam(
+                patient_date_of_birth.values[0], exam_dates
+            )
+        else:
+            age = "n/a"
 
         visit_code[visit_code == "bl"] = "M000"
         visit_code = visit_code.str.upper()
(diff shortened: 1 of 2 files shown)


clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -263,9 +272,29 @@ def create_sessions_tsv_file(
...
 
 def _clean_exam_dates(
-    rid: str, exam_dates: List[str], visit_codes: List[str], clinical_data_dir: Path
+    rid: int,
+    exam_dates: List[Union[str, int]],
+    visit_codes: List[str],
+    clinical_data_dir: Path,
 ) -> List[str]:
-    """Clean the exam dates when necessary by trying to compute them from other sources."""
+    """Clean the exam dates when necessary by trying to compute them from other sources.
```
NicolasGensollen, line 280: This function is supposed to return a list of strings but can actually return a list with "-4" values if it fails to find the dates in the other sources.
Also, using a code value to encode missing values isn't great. It'd be better to return `List[Optional[str]]` with missing values encoded as Nones.
Even better, we could return `List[Optional[datetime.date]]` since those strings are actually dates...
WDYT ?

test/unittests/iotools/converters/aibl_to_bids/test_aibl_utils.py
```diff
@@ -1,6 +1,73 @@
...
+
+    with pytest.raises(
+        ValueError,
+        match=f"Unexpected visit code foo. Should be in format mX :"
+        "Ex: m0, m6, m12, m048...",
+    ):
+        _compute_exam_date_from_baseline("foo", [], [])
+
+
+def test_find_exam_date_in_other_csv_files():
```
NicolasGensollen, line 39: Are you planing to implement these in this PR ?

If not, please remove them otherwise we would see those tests passing while the related code isn't actually tested.

NicolasGensollen (commented):
Thanks for adding these tests @AliceJoubert !

I think the PR looks good. I made a small suggestion on how to represent the list of exam dates which we can either implement here or keep for later depending on how you want to proceed.

There are also a lot of TODOs in the code. Are you planning to address them soon ?
If not, it could be better to open dedicated issues that we can track and not forget about. These TODO comments usually get added and stay there forever as nobody remembers what they meant but nobody feels legitimate to remove them either...

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
AliceJoubert: I totally agree with you. This function is used in `create_sessions...` where -4 for missing values are then exchanged for "n/a"... I think I will keep as is for this PR and change when I refactor `create_sessions...` to avoid changing too many things right away. I'm keeping in mind this comment for when I do this :-)

AliceJoubert:
Thanks for the review @NicolasGensollen ! 

I plan on addressing the TODOs right after this PR gets merged with associated issues so they are not here to stay a long time.

Do you see anything else on this PR ?

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
NicolasGensollen: Alright, sounds good. 👍 
Could you still update the return type to be `List[Union[str, int]]` then ?

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1348, opened 2024-10-23
[FIX] Correct the `pet-linear` pipeline

### LINKED ISSUE #1345: Correct `pet-linear`
Hello, 

As I was working on `pet-linear`, I realized that the SUVR normalization step is odd. Indeed, the non-linear transform is computed between the T1w MRI in the native step and the MNI reference template. The reverse of this transform is then applied to the affinely registered PET in order to compute the SUVR in the reference region.

However, the transformation should be computed on the T1w MRI that is already registered in the MNI space (the output of `t1-linear`) and it is the forward of the transformation that should be applied.

ravih18:
Closes #1345

### DIFF
--- a/clinica/pipelines/pet/linear/pipeline.py
+++ b/clinica/pipelines/pet/linear/pipeline.py
@@ -107,7 +107,11 @@ class PETLinear(PETPipeline):
 
         # Inputs from t1-linear pipeline
         # T1w images registered
-        t1w_linear_file_pattern = T1W_LINEAR if self.parameters.get("uncropped_image", False) else T1W_LINEAR_CROPPED
+        t1w_linear_file_pattern = (
+            T1W_LINEAR
+            if self.parameters.get("uncropped_image", False)
+            else T1W_LINEAR_CROPPED
+        )
         t1w_linear_files, t1w_linear_errors = clinica_file_reader(
             self.subjects, self.sessions, self.caps_directory, t1w_linear_file_pattern
         )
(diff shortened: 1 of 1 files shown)


clinica/pipelines/pet/linear/pipeline.py
```diff
@@ -101,6 +106,20 @@ def _build_input_node(self):
             )
 
         # Inputs from t1-linear pipeline
+        # T1w images registered
+        if self.parameters["uncropped_image"]:
+            t1w_linear_file_pattern = T1W_LINEAR
+        else:
+            t1w_linear_file_pattern = T1W_LINEAR_CROPPED
```
NicolasGensollen, line 113: When calling the pipeline from the CLI, there is indeed an "uncropped_image" entry in the `parameters` dictionary. But when calling from the "Python API" (like in the functional tests for example), there is no default value for it which results in a `KeyError` at this line.
We probably should improve this, but for the sake of this PR, we can provide a default value:

```suggestion
        t1w_linear_file_pattern = T1W_LINEAR if self.parameters.get("uncropped_image", False) else T1W_LINEAR_CROPPED
```

NicolasGensollen (commented):
Thanks for spotting and fixing this bug @ravih18 !
The PR LGTM with a small suggestion to avoid a `KeyError` when calling the pipeline from the Python API.

NicolasGensollen (approved):
LGTM, thanks @ravih18 !

PR #1355, opened 2024-10-24
[ENH] Add new possibilities to pet-linear

### LINKED ISSUE #1346: Enhancement of `pet-linear` pipeline
**Is your feature request related to a problem? Please describe.**
FDG-PET from some datasets (for instance ADNI) sometimes contain negative values that can impact the behavior of `pet-linear` pipeline.

**Describe the solution you'd like**
In order to deal with it, I would like to modify the `pet-linear` pipeline and add the following steps:
- Clip PET images in input and set the value of all the voxel below 0 to 0.
- Erase the background of the image at the end of the pipeline (set the background to 0 using a dilated mask of the MNI).

ravih18:
Closes #1346


ravih18:
Finally, after several tests, we decide to remove the node that removes the background and only keep the clipping of the background.

NicolasGensollen (approved):
LGTM, thanks @ravih18 !

I'll do a quick follow up PR to:

- Refactor the function `clip_img` to separate the clipping logic from the threshold computation logic (the former is generic and could be used in other places in clinica while the latter is very specific)
- Add some unit tests

PR #1359, opened 2024-10-28
[REF] AIBL-to-BIDS : code refactoring for sessions file creation

### LINKED ISSUE #1350: [REF] AIBL-to-BIDS : refactor code for creating sessions files
Following #1349, the function :
https://github.com/aramis-lab/clinica/blob/48c12ab0607638bf8b2fb755c06627c6f7f19a79/clinica/iotools/converters/aibl_to_bids/utils/clinical.py#L151-L155

can be modified to improve code quality. 

- [ ] Avoid unnecessary loops : 
https://github.com/aramis-lab/clinica/blob/48c12ab0607638bf8b2fb755c06627c6f7f19a79/clinica/iotools/converters/aibl_to_bids/utils/clinical.py#L182-L185
- [ ] Write mapping functions instead of : 
https://github.com/aramis-lab/clinica/blob/48c12ab0607638bf8b2fb755c06627c6f7f19a79/clinica/iotools/converters/aibl_to_bids/utils/clinical.py#L218-L221
- [ ] Use simpler handling for sessions identifiers

Utility functions used in it might be touched as well : 
- [ ] Investigate for simpler handling of dates
- [ ] Handling of missing values

AliceJoubert:
Close #1350

### DIFF
--- a/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
+++ b/clinica/iotools/converters/aibl_to_bids/utils/clinical.py
@@ -1,6 +1,5 @@
-from ast import Index
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Optional, Tuple
 
 import numpy as np
 import pandas as pd
@@ -10,8 +9,6 @@ __all__ = [
     "create_scans_tsv_file",
     "create_sessions_tsv_file",
 ]
-
-from sqlalchemy.sql.operators import from_
 
 
 def create_participants_tsv_file(
@@ -153,7 +150,6 @@ def _load_specifications(
 
 
 def _mapping_diagnosis(diagnosis: int) -> str:
-    # todo : what if str(1) ?
     if diagnosis == 1:
         return "CN"
     elif diagnosis == 2:
@@ -167,7 +163,6 @@ def _mapping_diagnosis(diagnosis: int) -> str:
 def _extract_metadata_df(
     input_df: pd.DataFrame, source_id: int, bids_metadata: str, source_metadata: str
 ) -> pd.DataFrame:
-    # todo :test
     from clinica.iotools.converter_utils import viscode_to_session
 
     extract = input_df.loc[(input_df["RID"] == source_id), ["VISCODE", source_metadata]]
@@ -182,32 +177,15 @@ def _extract_metadata_df(
 
 
 def _compute_age_at_exam(
-    birth_date: Union[float, str], exam_date: Union[float, str]
-) -> float:
-    """Compute the ages of the patient at each exam date.
-
-    Parameters
-    ----------
-    birth_date : str
-        Date of birth of patient ("/%Y" format)
-
-    exam_date : str
-        Exam date ("%m/%d/%Y" format)
-
-    Return
-    ------
-    float
-        Age of the patient at exam date.
-    """
-    # todo : test
+    birth_date: Optional[str], exam_date: Optional[str]
+) -> Optional[float]:
     from datetime import datetime
 
-    if not isinstance(birth_date, float) and not isinstance(exam_date, float):
+    if birth_date and exam_date:
         date_of_birth = datetime.strptime(birth_date, "/%Y")
         exam_date = datetime.strptime(exam_date, "%m/%d/%Y")
         return exam_date.year - date_of_birth.year
-    else:
-        return np.nan
+    return None
 
 
 def create_sessions_tsv_file(
@@ -228,6 +206,7 @@ def create_sessions_tsv_file(
     clinical_specifications_folder : Path
         The path to the folder containing the clinical specification files.
     """
+    # todo :rename test
     import glob
 
     from clinica.iotools.bids_utils import (
@@ -263,15 +242,21 @@ def create_sessions_tsv_file(
 
         test.sort_index(inplace=True)
         # -4 are considered missing values in AIBL
-        test.replace([-4, "-4"], np.nan, inplace=True)
+        test.replace([-4, "-4", np.nan], None, inplace=True)
         test["diagnosis"] = test.diagnosis.apply(lambda x: _mapping_diagnosis(x))
+        test["examination_date"] = test.apply(
+            lambda x: _complete_examination_dates(
+                rid, x.session_id, x.examination_date, clinical_data_dir
+            ),
+            axis=1,
+        )
 
-        # todo : handle exam date, see function below
         # in general age metadata is present only for baseline session
         test["age"] = test["age"].ffill()
         test["age"] = test.apply(
-            lambda row: _compute_age_at_exam(row.age, row.examination_date), axis=1
+            lambda x: _compute_age_at_exam(x.age, x.examination_date), axis=1
         )
+
         # in case there is a session in clinical data that was not actually converted
         test.dropna(subset=["session_id"], inplace=True)
         test.fillna("n/a", inplace=True)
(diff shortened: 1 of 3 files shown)


clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +208,119 @@ def create_sessions_tsv_file(
...
+        sessions = pd.DataFrame(
+            {"session_id": get_bids_sess_list(bids_dir / bids_id)}
+        ).set_index("session_id", drop=False)
 
-        if not patient_date_of_birth.empty:
-            age = _compute_ages_at_each_exam(
-                patient_date_of_birth.values[0], exam_dates
+        for _, row in specifications.iterrows():
+            df = pd.read_csv(
+                glob.glob(str(clinical_data_dir / row[f"{study} location"]))[0],
```
NicolasGensollen, line 231: Maybe something like this would be a bit more robust:

```python
try:
    df = pd.read_csv(next(clinical_data_dir.glob(row[f"{study} location"])), dtype=...)
except StopIteration:
    # do something if file doesn't exist...
```

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -148,17 +149,56 @@ def _load_specifications(
...
+        return "CN"
+    elif diagnosis == 2:
+        return "MCI"
+    elif diagnosis == 3:
+        return "AD"
+    else:
+        return "n/a"
+
+
+def _extract_metadata_df(
```
NicolasGensollen, line 163: I'd vote for a more informative name like `_build_sessions_for_rid` or something like that. WDYT ?

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -148,17 +149,56 @@ def _load_specifications(
...
+    )
+    extract.drop(labels="VISCODE", inplace=True, axis=1)
+    extract.set_index("session_id", inplace=True, drop=True)
+
+    return extract
+
+
+def _compute_age_at_exam(
+    birth_date: Optional[str], exam_date: Optional[str]
+) -> Optional[float]:
```
NicolasGensollen, line 181: Isn't this `Optional[int]` instead ?

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -148,17 +149,56 @@ def _load_specifications(
     return pd.read_csv(specifications, sep="\t")
 
 
+def _mapping_diagnosis(diagnosis: int) -> str:
```
NicolasGensollen, line 152:
```suggestion
def _map_diagnosis(diagnosis: int) -> str:
```

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +208,119 @@ def create_sessions_tsv_file(
...
+            axis=1,
         )
-        cols = sessions.columns.tolist()
-        sessions = sessions[cols[-1:] + cols[:-1]]
 
-        bids_id = bids_id_factory(StudyName.AIBL).from_original_study_id(str(rid))
+        # in general age metadata is present only for baseline session
+        sessions["age"] = sessions["age"].ffill()
+        sessions["age"] = sessions.apply(
+            lambda x: _compute_age_at_exam(x.age, x.examination_date), axis=1
```
NicolasGensollen, line 258: I'm probably missing something but aren't you supposed to pass the date of birth as the first parameter ?

```python
def _compute_age_at_exam(birth_date: Optional[str], exam_date: Optional[str])
```

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +208,119 @@ def create_sessions_tsv_file(
...
+                rid, x.session_id, x.examination_date, clinical_data_dir
+            ),
+            axis=1,
         )
-        cols = sessions.columns.tolist()
-        sessions = sessions[cols[-1:] + cols[:-1]]
 
-        bids_id = bids_id_factory(StudyName.AIBL).from_original_study_id(str(rid))
+        # in general age metadata is present only for baseline session
+        sessions["age"] = sessions["age"].ffill()
```
NicolasGensollen, line 256: I don't understand this. If you have the age at the baseline session, it won't necessarily be the age at later sessions, right ?

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +208,119 @@ def create_sessions_tsv_file(
...
-    for rid in unique_rids:
-        for file in files_to_read:
-            df = pd.read_csv(file, dtype={"text": str})
-            if len(df.columns) == 1:
-                df = pd.read_csv(file, sep=";", low_memory=False)
-
-            visit_code = df.loc[(df["RID"] == rid), "VISCODE"]
-
-            for field in sessions_fields_to_read:
-                if field in list(df.columns.values) and field == "MMSCORE":
```
NicolasGensollen, line 208: I don't see these hardcoded values in your code. 
I'm probably missing something, but how do you handle these in your version ?

NicolasGensollen (commented):
Thanks @AliceJoubert !
I made a first pass and made some suggestions along the way. I also have a few questions.
Looking at the diff, it's pretty difficult to know if the old and new versions of the code are equivalent.

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
AliceJoubert: To explain what my code does quickly : 
- I read the specifications file where it is indicated for each metadata of interest : name in the study, name for BIDS, location in csv files
--> I don't really need to hard code anything for metadata extraction since I loop over what is in the spec file. The process for "extraction" is the same, which you can find in the _format_metadata_for_rid function.
- If needed I perform some mapping which is similar to what is done here but I use the BIDS name instead. For ex my "diagnosis" mapping corresponds to what was done for field "DXCURREN"

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
AliceJoubert: You're right ! I did not totally modify what was done before _ie_ storing the birth date inside the age column and compute the age based on this. Now it should be more explicit.

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
AliceJoubert: Yes totally !

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -1,6 +1,8 @@
+from multiprocessing.managers import Value
```
NicolasGensollen, line 1:
```suggestion
```

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +225,118 @@ def create_sessions_tsv_file(
     """
     import glob
```
NicolasGensollen, line 226:
```suggestion
```

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -168,263 +225,118 @@ def create_sessions_tsv_file(
...
+        csv_data = pd.read_csv(csv, low_memory=False)
+        csv_data["SESSION"] = csv_data.VISCODE.apply(lambda x: viscode_to_session(x))
+        exam_date = csv_data[(csv_data.RID == rid) & (csv_data.SESSION == session_id)]
         if not exam_date.empty and exam_date.iloc[0].EXAMDATE != "-4":
             return exam_date.iloc[0].EXAMDATE
     return None
 
 
-def _get_csv_files(clinical_data_dir: Path) -> List[str]:
+def _get_csv_files_for_alternative_exam_date(clinical_data_dir: Path) -> Tuple[Path]:
```
NicolasGensollen, line 320: We can improve this a bit. Currently this function builds a tuple of Paths from a tuple of patterns in order for the caller to iterate over the paths. We can directly return a generator object which is a bit cleaner and memory efficient (not a major concern here I admit...):

```python
from typing import Generator

def _get_csv_files_for_alternative_exam_date(clinical_data_dir: Path) -> Generator[Path]:
    """Return the paths to CSV files in which an alternative exam date could be found."""
    for pattern in (
        "aibl_mri3meta_*.csv",
        "aibl_mrimeta_*.csv",
        "aibl_cdr_*.csv",
        "aibl_flutemeta_*.csv",
        "aibl_mmse_*.csv",
        "aibl_pibmeta_*.csv",
    ):
        try:
            yield next(clinical_data_dir.glob(pattern))
        except StopIteration:
            continue
```

test/unittests/iotools/converters/aibl_to_bids/test_aibl_utils.py
```diff
@@ -1,47 +1,39 @@
...
+    [
+        (-4, "n/a"),
+        (1, "CN"),
+        (2, "MCI"),
+        (3, "AD"),
+        (0, "n/a"),
+        (None, "n/a"),
+    ],
+)
+def test_mapping_diagnosis(diagnosis, expected):
```
NicolasGensollen, line 31:
```suggestion
def test_map_diagnosis(diagnosis, expected):
```

NicolasGensollen (commented):
Just a few minor things. LGTM otherwise

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
NicolasGensollen: Yep ! Much better now ! 👍

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
NicolasGensollen: Thanks 🙏

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
AliceJoubert: Thanks for the suggestion !

AliceJoubert:

clinica/iotools/converters/aibl_to_bids/utils/clinical.py
```diff
@@ -148,283 +149,188 @@ def _load_specifications(
...
+        exam_date = csv_data[(csv_data.RID == rid) & (csv_data.SESSION == session_id)]
         if not exam_date.empty and exam_date.iloc[0].EXAMDATE != "-4":
             return exam_date.iloc[0].EXAMDATE
     return None
 
 
-def _get_csv_files(clinical_data_dir: Path) -> List[str]:
+def _get_csv_files_for_alternative_exam_date(
+    clinical_data_dir: Path,
+) -> Generator[Path, None, None]:
```
NicolasGensollen, line 319: This returns a `Generator[Path]`, right ?

AliceJoubert: Yes ! I had to modify the typing to have 3 arguments in brackets, following the [docs](https://docs.python.org/3/library/typing.html#typing.Generator). Else, it raises an error. But maybe I should not put None for the _input_ and _return_ fields.

NicolasGensollen: Since it's a simple iterator, we can do:

```python
def func(...) -> Iterator[Path]:
    ...
```

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1366, opened 2024-11-08
[ENH] Improve handling of group labels and IDs

### LINKED ISSUE #1365: Improve handling of group labels and IDs
Some of the pipelines work with groups of subjects and rely on the notion of group label and group ID.

A group label is a string composed of letters and digits only, while a group ID is a string composed of the prefix "group-" followed by a group label.

At the moment, everything is represented as plain strings and the functions in `clinica.utils.group` are checking the validity of the user provided labels.

Moreover, the transformation from group label to group ID is hardcoded as

```python
f"group-{self.parameters['group_label']}"
```

all over the code base for pipelines (sometimes close to ten times in a single function...).

By relying on a better modeling of these two concepts we could remove a lot of validation code.

Here is a proposal:

```python
class GroupLabel(UserString):
    def __init__(self, value: str):
        super().__init__(self.validate(value))

    @classmethod
    def validate(cls, value: str) -> str:
        if value.isalnum():
            return value
        raise ValueError(
            f"Group label '{value}' is not a valid group label: it must be composed only by letters and/or numbers."
        )

class GroupID(UserString):
    def __init__(self, value: str):
        super().__init__(self.validate(value))

    @classmethod
    def validate(cls, value: str) -> str:
        if not value.startswith("group-"):
            raise ValueError(f"Group ID '{value}' is not a valid group ID: it must start with 'group-'.")
        GroupLabel.validate(value.removeprefix("group-"))
        return value

    @property
    def label(self) -> GroupLabel:
        return GroupLabel(str(self).removeprefix("group-"))

    @classmethod
    def from_label(cls, label: Union[str, GroupLabel]):
        return cls(f"group-{GroupLabel(label)}")
```

Finally, in the command lines of these pipelines, the group label is not an option but a required value. The fact that it is handled in a parameters dictionary with other optional things is a bit confusing.

It should instead be passed to the Pipeline constructor, validated once and for all by casting the provided string to a `GroupLabel`, and stored in an attribute of the class.

NicolasGensollen:
Closes #1365 

Most noticeable changes:

- Implementation of `GroupLabel` and `GroupID` in module `clinica.utils.group`
- Remove all previous validation code since it is now handled by these two classes
- Implement `GroupPipeline` which subclasses `Pipeline`. All pipelines relying on groups now inherit from this class.
- The constructor of `GroupPipeline` has two mandatory arguments: `caps_directory` and `group_label`. This means that it is impossible to create an instance of `GroupPipeline` without providing a valid CAPS folder and a valid group label.
- The group label is therefore not passed within the `parameters` dictionary anymore, but directly to the constructor
- `GroupPipeline` instances have the following group-specific attributes: 

```python
group_label: GroupLabel  # user provided
group_id: GroupID = GroupID.from_label(self.group_label)
group_directory: Path = self.caps_directory / "groups" / str(self.group_id)
```


AliceJoubert (approved):

Issue #1363, opened 2024-11-07
docs are not up to date online or something up with rendering?


yarikoptic:
Sorry if it is just my tired brain... on https://aramislab.paris.inria.fr/clinica/docs/public/latest/WhatIsClinica/ I see

![image](https://github.com/user-attachments/assets/99599c3b-5398-407c-959d-5549e67797dc)

where PET is not hyperlink and subitems are not indented... thought to fix that up but found that it already was fixed up in

- e7d4a45ac3c292fa3ac3e8aea16ccf5652c184ea for PET (although may be didn't render properly?)
- eafd6754f6ffc29676f86beade1a7b2431d238bd long ago for indentation

where is the catch?

NicolasGensollen:
Hi @yarikoptic 

Not your tired brain, but rather a bug on our side I'm afraid...

- the CI pipeline responsible for building and deploying the docs was running only on `dev` and maintenance branches
- the `latest` version of the docs is basically a symlink to the `master` version of the docs, which was therefore outdated since the `master` version is currently not updated by the CI

You were thus not viewing the real "latest" version of the docs. Sorry about that and thanks a lot for bringing this up.

I've made a fix for the CI on #1364 

Best,

yarikoptic:
Great thanks. 

Formatting might still be adjustment - no indentation for sublists? 

![Screenshot_20241108-065650](https://github.com/user-attachments/assets/def617eb-d3fc-486c-9555-a43608cf0280)

NicolasGensollen:
@yarikoptic Yes indeed, sublists aren't indented properly. Do you want to submit a PR to fix that ?

PR #1367, opened 2024-11-08
Fix indentations (must have 4 spaces for mkdocs to render properly)

yarikoptic:
Also did some consistency fixes in casing and punctuation.

ref
- #1363


NicolasGensollen (approved):
LGTM, thanks @yarikoptic !

PR #1376, opened 2024-11-14
[FIX] Fix type mismatch in `clinica.iotools.bids_utils`

### LINKED ISSUE #1375: PosixPath vs. string type error in `bids_utils.py`
The error in the screenshot seems like a bug to me indeed.

The variable `file_path` comes from 

https://github.com/aramis-lab/clinica/blob/4d700d75d56f9ada51ee88b66d39966c315d89e3/clinica/iotools/bids_utils.py#L569-L573

Such that it is a `PosixPath` and not a `string`. This line should probably be:

```python
if "flutemeta" in file_path.name and study_name == StudyName.AIBL
```

instead.

_Originally posted by @NicolasGensollen in https://github.com/aramis-lab/clinica/issues/1374#issuecomment-2476312396_

NicolasGensollen:
Closes #1375

### DIFF
--- a/clinica/iotools/bids_utils.py
+++ b/clinica/iotools/bids_utils.py
@@ -578,7 +578,7 @@ def create_scans_dict(
                 # case, we just remove the erroneous content and replace it with -4 which AIBL uses as n/a value.
                 on_bad_lines = (  # noqa: E731
                     lambda bad_line: bad_line[:-3] + [-4, bad_line[-1]]
-                    if "flutemeta" in file_path and study_name == StudyName.AIBL
+                    if "flutemeta" in file_path.name and study_name == StudyName.AIBL
                     else "error"
                 )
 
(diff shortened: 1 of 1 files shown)


AliceJoubert (approved):
LGTM, thanks @NicolasGensollen !

AliceJoubert:
Don't know if that would help you in this PR but I looked up what a `bad line` would be. Here is an example : 
`bad_line = ['618', '1', 'm18', '1', '08/06/2014', ... , '1', '1', 'measured', 'AUSTIN AC CT Brain  H19s', '0']`. The code is trying to get rid of the **two** keys `'measured'` and `'AUSTIN...'` (these are _always_ the same) and replace them with **one** key '-4'.

NicolasGensollen:
Thanks @AliceJoubert 
I just wanted to fix the type error described in the linked issue.
I'm assuming the logic is correct. I reverted to draft because the conditional assignment of the lambda functions wasn't working as expected. It should be good now.

AliceJoubert:
I checked with the latest clinical data from AIBL and it works as it should ! If you want to merge @NicolasGensollen

NicolasGensollen:

PR #1379, opened 2024-11-15
[FIX] Update the version of dcm2niix required to run clinica converters

AliceJoubert:
Following users issues and some tests, it seems the version of dcm2niix can cause trouble when converting some images. We should recommend a more recent version of dcm2niix in our documentation. WDYT @NicolasGensollen ?


NicolasGensollen (approved):
I agree, especially since the users should get a warning if they use a version of dcm2niix older than `1.0.20240202`:

https://github.com/aramis-lab/clinica/blob/4d700d75d56f9ada51ee88b66d39966c315d89e3/clinica/utils/check_dependency.py#L459-L460

PR #1392, opened 2024-11-22
[REF] Refactor function `clip_img`

NicolasGensollen:
Follow-up PR of #1355 

This PR proposes to:

- Add a new `clip_nifti` function to `clinica.utils.image` for generic clipping of nifty images
- Refactor `clip_img` to use `clip_nifti` with the computed threshold
- Provide the pipeline base directory as the `output_dir` of the clipping task such that clipped images aren't written in the current folder.
- Add unit tests for function `clip_nifti` and `clip_img`

### DIFF
--- a/clinica/pipelines/pet/linear/pipeline.py
+++ b/clinica/pipelines/pet/linear/pipeline.py
@@ -298,10 +298,11 @@ class PETLinear(PETPipeline):
             name="clipping",
             interface=nutil.Function(
                 function=clip_task,
-                input_names=["input_pet"],
+                input_names=["input_pet", "output_dir"],
                 output_names=["output_image"],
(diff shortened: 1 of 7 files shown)


AliceJoubert (approved):
LGTM thanks @NicolasGensollen ! I don't remember if the clip was to be an option or not. In any case, the documentation should be updated to indicate that now we clip images :-)

NicolasGensollen:
The clipping of the background is not implemented as an option at the moment. I don't believe we want to do that (but we totally can do it if needed).
I agree with you that we need to update the doc of the pipeline to explain that backgrounds are now clipped (will do that in a dedicated PR), thanks for thinking about that !

PR #1403, opened 2024-12-04
[ENH] Skip processed visits for t1-volume-tissue-segmentation

### LINKED ISSUE #1400: Skip already processed subjects and sessions for `T1VolumeTissueSegmentation`
Similar to #1079 but for pipeline `T1VolumeTissueSegmentation`.

Adapt what was done in #1399 for `T1Linear` and `PETLinear`.

NicolasGensollen:
Closes #1400

### DIFF
--- a/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
+++ b/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
@@ -3,9 +3,8 @@ from typing import List
 
 from nipype import config
 
+from clinica.dataset import Visit
 from clinica.pipelines.engine import Pipeline
-from clinica.utils.bids import Visit
-from clinica.utils.filemanip import extract_visits
 
 cfg = dict(execution={"parameterize_dirs": False})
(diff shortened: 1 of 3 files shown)


AliceJoubert (commented):
Thanks for the work @NicolasGensollen ! It works for `t1_volume_tissue_segmentation`, I was just wondering for the whole `t1_volume` pipeline. Should we also update the list of subjects and sessions processed by the other pipelines called downstream ? Or do we need to have this implemented for each pipeline ?

NicolasGensollen:
> Thanks for the work @NicolasGensollen ! It works for `t1_volume_tissue_segmentation`, I was just wondering for the whole `t1_volume` pipeline. Should we also update the list of subjects and sessions processed by the other pipelines called downstream ? Or do we need to have this implemented for each pipeline ?

To be honest it's been a while since I opened this PR so I don't remember all the details, but I was planning to implement the same kind of logic for each sub-pipelines of t1-volume such that the skipping logic would work even when sub-pipelines are used in isolation.

If you think there is a better way, we can close this and try your idea.

NicolasGensollen:
Should I merge or do you want to take another look at it ?

PR #1409, opened 2025-01-08
PET-linear: apply transform fix

HuguesRoy:
When applying the nonlinear registration T1 to MNI, the reverse_forward_transforms should be used instead of the forward_transforms. 

As the registration is non-linear, the transformation should be applied in the correct order. 

The forward_transforms output is a list containing the transformation in natural order: the first element should be applied first, and the last element should be applied last. However, the ApplyTransform applies lists of transformations in reverse order (the last transformation in the list is applied first). Thus, the "reverse_forward_transforms" output from ANTS registration should be used.

source: https://nipype.readthedocs.io/en/latest/api/generated/nipype.interfaces.ants.html#registration


AliceJoubert (approved):
LGTM, thanks @HuguesRoy for spotting that !

ravih18:
Maybe it should be tested on a few real images to check if the normalization is still good after that fix !

PR #1416, opened 2025-01-15
[ENH] Change long_id format

### LINKED ISSUE #1415: Unpractical longitudinal filenaming
Hi Clinica devs,

The current longitudinal pipeline does not separate the session ids in the longitudinal id.

What it means is that if you have 3 sessions which are `ses-M0`, `ses-M3`, and `ses-M6`, the longitudinal id will be `long-M1M3M6`.
However if the sessions are `ses-1`, `ses-2` and `ses-3`, then you get `long-123`, which makes it hard to read.

IMHO, adding a separator would enhance the readability and usability of the pipeline. On my side, I used a point as separator which would get the format to be long-1.2.3

Have a good day,
Matthieu

AliceJoubert:
Close #1415 

- Uses '+' as a separator

- TODO : modify CI data names

### DIFF
--- a/clinica/pipelines/anatomical/freesurfer/longitudinal/correction/utils.py
+++ b/clinica/pipelines/anatomical/freesurfer/longitudinal/correction/utils.py
@@ -282,7 +282,7 @@ def get_processed_images(
     atlas: Optional[str] = None,
 ) -> list[str]:
     """
-    Extract image IDs (e.g. ["sub-CLNC01_ses-M000_long-M000M018", "sub-CLNC01_ses-M018_long-M000M018"]) of outputs
+    Extract image IDs (e.g. ["sub-CLNC01_ses-M000_long-M000+M018", "sub-CLNC01_ses-M018_long-M000+M018"]) of outputs
     already processed by T1FreeSurferLongitudinalCorrection pipeline.
     """
     image_ids = []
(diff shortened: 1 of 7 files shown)


NicolasGensollen (commented):
A non trivial example of where such a change would have consequences not captured by our tests (I think):

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/longitudinal/template/pipeline.py#L37-L40

AliceJoubert:
> A non trivial example of where such a change would have consequences not captured by our tests (I think):
> 
> https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/longitudinal/template/pipeline.py#L37-L40

Thanks for spotting that ! If we agree using a separator would be better for user comprehension, depending on the separator we could modify that regex :

AliceJoubert:
> A non trivial example of where such a change would have consequences not captured by our tests (I think):
> 
> https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/longitudinal/template/pipeline.py#L37-L40

Thanks for spotting that ! I can modify the regex depending on the separator we want to use. I made the modifications in the PR with "." but that might not be the best choice since it is widely used. We can also go with "+". WDYT ?

Issue #1413, opened 2025-01-10
Center nifti does not work how it is supposed to


MatthieuJoulot:
Dear Clinica devs,

Following my unsuccessful usage of `t1-freesurfer`, I used `center-nifti` which does not work the way it says it does. The help section says that the mandatory arguments are a `bids_directory` and an `output_directory`. However if an additional argument is not given, whether it is `-m/--modality` or `--center-all-files` nothing is done. 

It looks like there are two options:

1. remove the `--center-all-files` flag and make processing all files the default behavior (which is what I thought it would do).
2. make an other compulsory argument to say whether all the files or only selected modalities should be processed (which seems counter-intuitive and complicated to implement cleanly IMHO)

Have a good one,
Matthieu

Issue #1412, opened 2025-01-10
T1-FreeSurfer requires interaction


MatthieuJoulot:
Hi Clinica devs,

I have been using `t1-freesurfer` these days, and it turns out, it is not possible to launch a batch job if your images are not centered, because, `t1-freesurfer` is interactive when it is the case. It asks whether you want to continue launching `t1-freesurfer`. Which I can't since it is a slurm job.

This behavior is linked to a usage of `t1-freesurfer` which is not just something your average Joe does.

I think there is different approaches to handle this:

1. Give warnings in `t1-freesurfer`.
2. Have a `pet-surface` flag which would ensure a centered dataset is used, as you may hope that your advanced Joe would have read the different documentations and therefore know about the centering.
3. Be explicit about it in the --help documentation

Have a good one and see you soon 😘
Matthieu

NicolasGensollen:
Hi @MatthieuJoulot 

Thanks for reporting this issue. If I understand correctly, you launch `t1-freesurfer` on a cluster on non centered images and you have issues because Clinica gets in interactive mode in this case.

More precisely, this happens at the beginning, when building the input node, here:

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/t1/pipeline.py#L158-L162

and the interactive call is triggered here:

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/iotools/utils/data_handling/_centering.py#L119-L120

The default value of the `skip_question` parameter is set in a very misleading way here:

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/t1/cli.py#L67-L70

Where `yes` is actually a variable whose default value is `False`...

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/anatomical/freesurfer/t1/cli.py#L47

So, by default, the pipeline does NOT skip the question. To skip the question, you need to add the `-y` (or `--yes`) flag to your command:

https://github.com/aramis-lab/clinica/blob/d379ae6a87faefefe9413a002498e7ea6d5902e4/clinica/pipelines/cli_param/option.py#L202-L207

Did I understand correctly ? Have you tried adding this option ?

[...]

PR #1418, opened 2025-01-22
[FIX] `Center_Nifti` should run by default on T1

AliceJoubert:
Adresses part of #1413

The modalities to center were set as described in the documentation. By default, `center_nifti` will only process T1w images. It is possible to change this by using the `--modality` option and you can process several modalities at the same time. 

Modalities are not restricted to a set of known modalities on our side to give the user more freedom. That means the user could enter `-m TOTO` if they wanted to, however the implementation is at the moment not very robust. **This can be changed.**

For follow-up PR : 
- Review the CI data et non regression tests to handle more cases (command line tested?)
- Review logic to check for files since split in 2 functions. Might be more practical to move the logic for it to be tested and completed (issue if output folder does not exist yet).
- Check behavior for writing over
- Review threshold logic (see discussion below)

### What is left to discuss : 

From the documentation, this tool is used in clinica if the data should at some point be processed by SPM (hence the discussion from #1412 on T1-Freesurfer). As of now, `center-nifti` processes by default only files that are more than 50 mm away from the center. The threshold was empirically established specifically for SPM by previous users. It is definitely not intuitive if someone uses `center-nifti` for any other purpose than SPM processing. We could : 
- get rid of the threshold altogether and process all images (🥇 in my opinion)
- change the default behavior to processing all images and use an option to accelerate the processing by allowing the user to choose the treshold

### DIFF
--- a/clinica/iotools/utils/data_handling/_centering.py
+++ b/clinica/iotools/utils/data_handling/_centering.py
@@ -274,7 +274,7 @@ def _compute_l2_norm(file_couples: List[Tuple[Path, Path]]) -> List[float]:
 def _find_files_with_modality(
     files_dir: PathLike, modalities: Optional[Iterable[str]] = None
 ) -> List[Path]:
-    nifti_files = files_dir.glob("*.nii*")
+    nifti_files = files_dir.rglob("*.nii*")
     nifti_files_filtered = []
 
     if modalities is None:
--- a/clinica/iotools/utils/data_handling/_centering.py
+++ b/clinica/iotools/utils/data_handling/_centering.py
@@ -272,7 +272,7 @@ def _compute_l2_norm(file_couples: List[Tuple[Path, Path]]) -> List[float]:
 
 
 def _find_files_with_modality(
-    files_dir: PathLike, modalities: Optional[Iterable[str]] = None
+    files_dir: Path, modalities: Optional[Iterable[str]] = None
 ) -> List[Path]:
     nifti_files = files_dir.rglob("*.nii*")
     nifti_files_filtered = []
@@ -280,7 +280,7 @@ def _find_files_with_modality(
     if modalities is None:
         return nifti_files
     for f in nifti_files:
-        if any(elem.lower() in f.name.lower() for elem in modalities):
+        if any(modality.lower() in f.name.lower() for modality in modalities):
             nifti_files_filtered.append(f)
     return nifti_files_filtered
 
--- a/clinica/iotools/utils/data_handling/_centering.py
+++ b/clinica/iotools/utils/data_handling/_centering.py
@@ -271,6 +271,20 @@ def _compute_l2_norm(file_couples: List[Tuple[Path, Path]]) -> List[float]:
     return [np.linalg.norm(center[0] - center[1]) for center in center_coordinates]
 
 
+def _find_files_with_modality(
+    files_dir: PathLike, modalities: Optional[Iterable[str]] = None
+) -> List[Path]:
+    nifti_files = files_dir.glob("*.nii*")
+    nifti_files_filtered = []
+
+    if modalities is None:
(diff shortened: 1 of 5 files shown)


clinica/iotools/utils/data_handling/_centering.py
```diff
@@ -271,6 +271,20 @@ def _compute_l2_norm(file_couples: List[Tuple[Path, Path]]) -> List[float]:
     return [np.linalg.norm(center[0] - center[1]) for center in center_coordinates]
 
 
+def _find_files_with_modality(
+    files_dir: PathLike, modalities: Optional[Iterable[str]] = None
```
NicolasGensollen, line 275:
```suggestion
    files_dir: Path, modalities: Optional[Iterable[str]] = None
```

NicolasGensollen (commented):
Thanks @AliceJoubert !
I had a very quick look at the code which looks good.
I'll come back tomorrow to the points left to discuss.

NicolasGensollen:
> Adresses part of #1413

I'm not sure to understand how this solves the problem that @MatthieuJoulot described:

> However if an additional argument is not given, whether it is -m/--modality or --center-all-files nothing is done.

Reading the docs, it seems like without any option, all T1w images being more than 50mm from the origin would be centered. Nothing was done in this case because all your images were under this 50mm threshold ?

> Modalities are not restricted to a set of known modalities on our side to give the user more freedom. That means the user could enter `-m TOTO` if they wanted to, however the implementation is at the moment not very robust. **This can be changed.**

I agree, but I don't believe we have a robust modeling of the modalities handled by clinica at the moment. They are usually handled as plain strings. Ideally this would be an enumeration of values and the user could only provide a valid variant, but that would require some work. So, definitely not in this PR, but we can discuss about that.

> _Maybe in another PR :_ _- move logic to check for files since split in 2 functions. Need to be completed bc can cause issue if output folder does not exist yet._ _- check behavior for writing over_

I'm not sure I see exactly what you want to do, but if you feel that it would make the code better, you can totally open a follow-up PR.

> ### What is left to discuss :
> From the documentation, this tool is used in clinica if the data should at some point be processed by SPM (hence the discussion from #1412 on T1-Freesurfer). As of now, `center-nifti` processes by default only files that are more than 50 mm away from the center. The threshold was empirically established specifically for SPM by previous users. It is definitely not intuitive if someone uses `center-nifti` for any other purpose than SPM processing. We could :
> 
> * get rid of the threshold altogether and process all images (🥇 in my opinion)
> * change the default behavior to processing all images and use an option to accelerate the processing by allowing the user to choose the treshold

I might be missing some details but I would have gone with option 2 which gives more control to the user. For example, it allows to both preserve the current default behavior (setting the threshold default as 50mm) and mimic option 1 (setting the threshold at 0). Option 1 has the benefit of being simpler both in terms of code and UI though. WDYT ?

AliceJoubert:
Actually before my modifications there was no default set for the `-m` option. That means if you did not enter anything the value for modalities was `()`. The function afterwards did not handle this case so nothing was done at all, whether the images were above or below the 50mm threshold. Sorry if that was not clear. For now in this PR all I did was fix this to actually match what we say the module is supposed to do. 

I feel like this module is supposed to let the users quite free in terms of modality so that would be difficult to model. We could restrict the search of the modality key to the suffix (right before the .nii.gz extension) but that is about all we can do easily.

The `--center-all-files` problem is a separate issue so I think this will also go in another PR. Option 2 definitely gives more flexibility, I just wonder to what extent it would actually be useful for users but I have no experience on that. Since the module is supposed to be for SPM first that would make sense to keep it.

NicolasGensollen (approved):
The PR LGTM, thanks @AliceJoubert !
If you think it is ready, you can flip the switch from draft to open and we'll merge.

PR #1423, opened 2025-01-30
Fix typos in output tsv from check-missing-processing

### LINKED ISSUE #1422: Typos in output tsv from check-missing-processing
Hello,

I had an issue when running the command `clinica iotools check-missing-processing` on an ADNI BIDS and CAPS, where the output tsv had the following issues: 
- in the participant_id column was writtten "subjects" for all participants instead of the participant_id,
- in the session_id column was the participant_id instead of the session_id, 
- for certain pipelines, the column name was not correctly formatted: for instance, (0, 't1-linear') instead of t1-linear. 

Image

- OS: Ubuntu 20.04.6 LTS
- Version: 0.9.3

msolal:
Closes #1422


NicolasGensollen (approved):
LGTM ! I was able to reproduce the issue and test your fix.

Thanks a lot @msolal for spotting and fixing this bug ! 🥇

PR #1424, opened 2025-01-31
[TEST] Add non-regression tests for AIBL that compares the inside of all metadata tsv files

AliceJoubert:
Only for AIBL for now
Need to change CI data

### DIFF
--- a/test/nonregression/testing_tools.py
+++ b/test/nonregression/testing_tools.py
@@ -1,9 +1,10 @@
 # coding: utf8
 
 import os
+from enum import Enum
 from os import PathLike
 from pathlib import Path
-from typing import Dict, List, Tuple
+from typing import Callable, Dict
 
 import numpy as np
 import pandas as pd
@@ -13,7 +14,7 @@ def configure_paths(
     base_dir: Path,
     tmp_path: Path,
     name: str,
-) -> Tuple[Path, Path, Path]:
+) -> tuple[Path, Path, Path]:
     """Configure paths for tests."""
     input_dir = base_dir / name / "in"
     ref_dir = base_dir / name / "ref"
@@ -26,8 +27,8 @@ def configure_paths(
 def likeliness_measure(
     file1: PathLike,
     file2: PathLike,
-    threshold1: Tuple,
-    threshold2: Tuple,
+    threshold1: tuple,
+    threshold2: tuple,
     display: bool = False,
 ) -> bool:
     """Compares 2 Nifti inputs, with 2 different thresholds.
@@ -245,7 +246,7 @@ def _is_included(sub_ses_tsv_1: PathLike, sub_ses_tsv_2: PathLike) -> bool:
     return True
 
 
-def _sort_subject_field(subjects: List, fields: List) -> List:
+def _sort_subject_field(subjects: list, fields: list) -> list:
     """Helper function for `same_missing_modality_tsv`.
     Returns a sorted list of fields. The list is sorted by corresponding
     subject_id and by field_id if the subject_ids are equal.
@@ -361,8 +362,8 @@ def clean_folder(path: PathLike, recreate: bool = True):
 
 def list_files_with_extensions(
     path_folder: PathLike,
-    extensions_to_keep: Tuple[str, ...],
-) -> List[str]:
+    extensions_to_keep: tuple[str, ...],
+) -> list[str]:
     """List all the files with the provided extensions
     in the path_folder.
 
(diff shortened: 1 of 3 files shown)


test/nonregression/testing_tools.py
```diff
@@ -483,3 +483,51 @@ def compare_folders_structures(
...
+        )
+    for tsv in bids_ref.rglob(f"*{name}.tsv"):
+        _compare_frames(load_function(bids_out, tsv), load_function(bids_ref, tsv))
+
+
+def compare_bids_tsv(bids_out: Path, bids_ref: Path):
+    _compare_frames(_load_participant_tsv(bids_out), _load_participant_tsv(bids_ref))
+    _iteratively_compare_frames(bids_ref, bids_out, name="sessions")
+    _iteratively_compare_frames(bids_ref, bids_out, name="scans")
+    return True
```
NicolasGensollen, line 533: Your code looks good. I would have gone all the way into handling all levels in a generic way (as opposed to handle the subject level a bit differently). Maybe something more or less like that with our usual patterns:

```python
class Level(str, Enum):
    SUBJECT = "subject"
    SESSION = "session"
    SCANS = "scans"

LoaderInterface = Callable[[Path, Path], pd.DataFrame]

def _loader_factory(level: str | Level) -> LoaderInterface:
    if (level := Level(level)) == Level. SUBJECT:
        return _load_participant_tsv
    if level == Level. SESSION:
        return _load_sessions_tsv
    if level == Level.SCANS:
        return _load_scans_tsv

def _iteratively_compare_frames(bids_ref: Path, bids_out: Path, level: Level):
    loader = _loader_factory(level)
    for tsv in bids_ref.rglob(f"*{name}.tsv"):
        _compare_frames(loader(bids_out, tsv), loader(bids_ref, tsv))

def compare_bids_tsv(bids_out: Path, bids_ref: Path):
    for level in Level:
         _iteratively_compare_frames(bids_ref, bids_out, level)
```

The price to pay is that `_load_participant_tsv` will implement `LoaderInterface` although is doesn't need the second argument.

WDYT ?

NicolasGensollen (commented):
I started looking before I realized it's still a draft, sorry about that.
Still, I left my suggestion in case you want to accept it before continuing.

test/nonregression/testing_tools.py
AliceJoubert: No worries for the early review ! I would have needed that kind of thing anyway considering the way I was beginning to make conditions on tsv types. That's much clearer and I will use it, thanks !

test/unittests/test_testing_tools.py
```diff
@@ -221,3 +228,109 @@ def test_compare_folders_structures(
...
+        {
+            "filename": ["pet/foo.json", "anat/foo.json"],
+            "acq_time": ["00:00:00", "00:00:00"],
+        }
+    )
+    scans.to_csv(ses_path / "sub-001_ses-M016_scans.tsv", sep="\t", index=False)
+    return bids_path
+
+
+def test_loaders(tmp_path):
```
NicolasGensollen, line 254: Could you split this test in three ? 
First, it'd be more clean and unitary and also, you'll see which loaders are broken the first time without having to re-run multiple times.

test/unittests/test_testing_tools.py
```diff
@@ -221,3 +228,109 @@ def test_compare_folders_structures(
...
+    ],
+)
+def test_compare_bids_tsv_error(tmp_path, modified_frame, frame_path):
+    from shutil import copytree
+
+    bids_path = build_bids_tsv(tmp_path)
+    copy = tmp_path / "BIDS_copy"
+    copytree(bids_path, copy)
+    modified_frame.to_csv(copy / frame_path, sep="\t", index=False)
+    with pytest.raises(AssertionError):
```
NicolasGensollen, line 335: Do you think you could check for the error message as well ?

NicolasGensollen (commented):
Thanks @AliceJoubert !

Two minor suggestions, but LGTM otherwise.

test/unittests/test_testing_tools.py
AliceJoubert: I added a string to partly match the error message each time. Would you say it is enough ? @NicolasGensollen

NicolasGensollen: Yes, seems good enough to me. Thanks !

NicolasGensollen (approved):
Approving since the non regression tests for AIBL are passing and the error is due to a VM having no free space.

PR #1426, opened 2025-02-05
[FIX] Fix function `crop_nifti` when bounding box is outside of the image

### LINKED ISSUE #1425: Inconsistent cropped image size for ADNI FLAIR
Hello, 

I ran flair-linear on ADNI and the output images have a shape of (169, **205**, 179), different than that of the usual template size (169, **208**, 179) from the documentation. 

After investigation it appears that this comes from the size of  _input_image_ (line 433 in utils/image.py : _input_image.get_fdata()_ ) which is 3 pixel short on the y-axis compared to the coordinates of _MNI_CROP_BBOX_ : _input_image_ is of size (182, **218**, 182), and _MNI_CROP_BBOX_ has coordinates (12, 181, 13, **221**, 0, 179).

This can be problematic when using these images as inputs for neural networks, especially when working with multiple datasets with some for which the final cropped image dimensions do match the template.

Perhaps the size of the images could be checked earlier on in the pipeline and corrected if necessary to ensure a consistent final cropped size while not altering the registration to the MNI space ?

Best,

NicolasGensollen:
Closes #1425

### DIFF
--- a/clinica/utils/exceptions.py
+++ b/clinica/utils/exceptions.py
@@ -34,6 +34,10 @@ class ClinicaExistingDatasetError(ClinicaDatasetError):
         )
 
 
+class ClinicaImageDimensionError(ClinicaException):
+    """Base class for errors linked to image dimensions."""
+
+
 class ClinicaParserError(ClinicaException):
     """Base class for parser errors."""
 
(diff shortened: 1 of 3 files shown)


AliceJoubert (approved):
LGTM thanks @NicolasGensollen !

PR #1427, opened 2025-02-05
[CI] Stop the macOS pipelines

NicolasGensollen:
Since the MacOS self-hosted runner is not working properly as a service, let's comment the related pipelines and avoid the consistently broken CI.


AliceJoubert (approved):
LGTM thanks @NicolasGensollen !

PR #1428, opened 2025-02-07
[TEST] Add non-regression tests to anat pipelines

AliceJoubert:
Adds a function that checks inside the nifti files, only in T1 Linear for now (to extend if code validated).

### DIFF
--- a/test/unittests/test_testing_tools.py
+++ b/test/unittests/test_testing_tools.py
@@ -348,20 +348,18 @@ def test_compare_bids_tsv_error(tmp_path, modified_frame, frame_path, error_mess
         compare_bids_tsv(bids_path, copy)
 
 
-def _create_nifti_in_dir(dir_path: Path, zeroes: Optional[bool] = False) -> Path:
+def _create_nifti_in_dir(dir_path: Path, rds) -> Path:
     dir_path.mkdir()
-    data = (
-        np.random.RandomState(42).random((16, 16, 16, 16))
-        if not zeroes
-        else np.zeros((16, 16, 16, 16))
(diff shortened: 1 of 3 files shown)


test/unittests/test_testing_tools.py
```diff
@@ -345,3 +346,31 @@ def test_compare_bids_tsv_error(tmp_path, modified_frame, frame_path, error_mess
...
     with pytest.raises(AssertionError, match=error_message):
         compare_bids_tsv(bids_path, copy)
+
+
+def _create_nifti_in_dir(dir_path: Path, zeroes: Optional[bool] = False) -> Path:
+    dir_path.mkdir()
+    data = (
+        np.random.RandomState(42).random((16, 16, 16, 16))
+        if not zeroes
+        else np.zeros((16, 16, 16, 16))
```
NicolasGensollen, line 356: I think I'd have relied entirely on the random state to generate different nifti images (i.e. pass the random state as a function parameter). It seems a bit simpler, but we can totally keep this as is.

NicolasGensollen (commented):
Thanks @AliceJoubert !

The code LGTM.
Just wondering, have you seen a big difference in terms of computation time when running the test with and without nifti comparison ?

test/unittests/test_testing_tools.py
AliceJoubert: You're right I could use it fully, that's better

AliceJoubert:
Thank for the review @NicolasGensollen. For T1-linear only approx 3s of a difference but there are not many images in the T1 folder I think

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1430, opened 2025-02-10
[ENH] Improve the `crop_nifti`

NicolasGensollen:
Following #1426, this PR proposes to improve the `crop_nifti` function.

Currently the function behaves in a very specific way, by cropping the provided image with a hardcoded bounding box that we defined for a specific use case (T1 images with MNI reference template). Therefore, I think the name is misleading and we could make use of a more generic function.

This PR proposes to make `crop_nifti` more generic by letting users provide both the bounding box and the reference image. If no bounding box is provided, then no cropping is performed. If no reference image is provided, the input image is used as a reference. In other words `crop_nifti(img)` with no other argument is the identity function.

The old `crop_nifti` has been renamed `crop_nifti_using_t1_mni_template` and is nothing more than a partial function on the new `crop_nifti`, calling it with the MNI_BBOX and the MNI cropped template. 

Implementation details: I introduced two classes called `NiftiImage` and `NiftiImage3D` which behave as wrappers around a path to a nifti image. I'm still unsure whether we should keep them or not, but I think it could help us reduce boilerplate code we have in various places to check that a nifti is 3D, or that a nifti file exists, and so on.

### DIFF
--- a/clinica/utils/image.py
+++ b/clinica/utils/image.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
+from functools import cached_property
 from os import PathLike
 from pathlib import Path
 from typing import Callable, Optional, Tuple, Union
@@ -10,6 +11,8 @@ from nibabel.nifti1 import Nifti1Image
 
 __all__ = [
     "HemiSphere",
+    "NiftiImage",
+    "NiftiImage3D",
     "compute_aggregated_volume",
     "get_new_image_like",
     "merge_nifti_images_in_time_dimension",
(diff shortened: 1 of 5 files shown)


clinica/utils/image.py
```diff
@@ -403,26 +411,95 @@ def _get_mni_template_flair() -> Path:
...
+
+    def load(self) -> nib.Nifti1Image:
+        return nib.load(self.path)
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        return self.load().shape
+
+    @property
+    def bbox(self) -> Bbox3D:
```
AliceJoubert, line 437: Just wondering if this should not be a property of NiftiImage3D

clinica/utils/image.py
```diff
@@ -403,26 +411,95 @@ def _get_mni_template_flair() -> Path:
...
+        image_path = Path(image_path)
+        if not image_path.is_file():
+            raise FileNotFoundError(f"File {image_path} does not exist.")
+        try:
+            nib.load(image_path)
+        except ImageFileError:
+            raise IOError(f"File {image_path} is not a nifti image or is corrupted.")
+        self.path = Path(image_path)
+
+    def load(self) -> nib.Nifti1Image:
```
AliceJoubert, line 429: Would that be possible to store the data that was loaded somewhere ? If we want to do for example image.shape() then image.affine() we would need to load the image twice which is not really necessary

AliceJoubert (approved):
Thanks for the work @NicolasGensollen ! LGTM, I just have 2 comments about the NiftiImage class

clinica/utils/image.py
NicolasGensollen: It totally should, it's a mistake of mine. Thanks for spotting it !

clinica/utils/image.py
NicolasGensollen: Yes, it's a good idea. 👍

NicolasGensollen: Proposed an implementation in 3d4163e

PR #1433, opened 2025-02-14
[ENH] In `center-nifti`, give the option to the user to choose the centering threshold

### LINKED ISSUE #1413: Center nifti does not work how it is supposed to

AliceJoubert:
Closes #1413 following #1418 

Changes the option of `--center_all_files` to `--threshold` above which images are to be centered.

In other PR : 
- Add non regression tests that checks inside images and txt file
- Change CI data if necessary (might be interesting to have different thresholds/modalities to test)
- Use the new NiftiImage class in `_get_world_coordinate_of_center`

### DIFF
--- a/clinica/iotools/utils/center_nifti.py
+++ b/clinica/iotools/utils/center_nifti.py
@@ -49,7 +49,10 @@ def center_nifti(
     )
     from clinica.utils.stream import cprint, log_and_raise
 
-    cprint("Clinica is now centering the requested images.", lvl="info")
+    cprint(
+        f"Clinica is now centering the requested images with a threshold of {centering_threshold} mm.",
+        lvl="info",
+    )
 
     centered_files = center_all_nifti(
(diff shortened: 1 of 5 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1437, opened 2025-02-20
[FIX][DOC] Modify the name of variables in the client and documentation to fit actual possibilities for `create-subjects-visits`

AliceJoubert:
The variable for the input directory described with the `--help` option were incoherent with the documentation. Both the client and the documentation were inexact about it since the tool can process both BIDS and CAPS datasets.


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1438, opened 2025-02-20
[TEST] Modify non-regression tests for `center-nifti`

AliceJoubert:
Following #1433, non-regression needed to be updated and completed with nifti comparison.

### To discuss :

In the non-regression tests we are testing `center_all_nifti` and not `center-nifti`.  Is there a reason for this ? We are missing the txt file which can be quite useful.


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1439, opened 2025-02-20
[TEST] Add nifti comparison to FLAIR linear pipeline

AliceJoubert:
- Uses function `compare_niftis` in the FLAIR linear pipeline.
- `compare_niftis` was made a little more flexible by giving access to the metric threshold used as a requirement for similarity validation

### To discuss
We have images that don't pass the test right now. This might be partly due to the recent changes in the cropping function but some uncropped images tests aren't validated either. The threshold for the similarity metric is 0.9, I don't know if it makes sense to go lower or if we would need to update all the faulty images.

### DIFF
--- a/test/nonregression/pipelines/anat/test_t1_linear.py
+++ b/test/nonregression/pipelines/anat/test_t1_linear.py
@@ -65,4 +65,4 @@ def run_flair_linear(
     pipeline.run(plugin="MultiProc", plugin_args={"n_procs": 4}, bypass_check=True)
 
     compare_folders(out_caps, ref_caps, output_dir)
-    compare_niftis(out_caps, ref_caps)
+    compare_niftis(out_caps, ref_caps, threshold=0.9)
--- a/test/nonregression/pipelines/anat/test_t1_linear.py
+++ b/test/nonregression/pipelines/anat/test_t1_linear.py
@@ -52,13 +52,17 @@ def run_flair_linear(
 ) -> None:
     from clinica.pipelines.t1_linear.anat_linear_pipeline import AnatLinear
 
+    out_caps = output_dir / "caps"
+    ref_caps = ref_dir / "caps"
+
     pipeline = AnatLinear(
         bids_directory=fspath(input_dir / "bids"),
-        caps_directory=fspath(output_dir / "caps"),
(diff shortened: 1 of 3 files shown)


NicolasGensollen (commented):
Thanks @AliceJoubert !
The code LGTM.

Concerning the threshold issue, did you try to run the pipeline multiple time and verify that you were having the same output ? I'm asking because this test will only make sense if we have a fully deterministic mechanism.

If there is some randomness, we need to understand why and see whether we can control it (setting a random seed for example). If we cannot, then see what threshold would work and if it makes sense to have it.

If there is no randomness, then it means that the pipeline drifted from what it once produced. If that is the case, we should take a closer look before updating the reference images. The references were produced a long time ago, so it'll be hard to find out why things drifted. It could simply due to new versions of dependencies for example...

AliceJoubert:
I ran the regression tests several times to see if the similarity metrics changed but the values are the same each time so there is no variability. If I am not wrong the random seed in anatomical (T1 and FLAIR) pipelines works the same way and is set to 0 by default. 

That means it drifted from the original results, or I need to check the metric

NicolasGensollen:
What if you just test for equality of the images without using a metric ?

You can use this function if needed:

https://github.com/aramis-lab/clinica/blob/31b2838ecc2f3c5336c5d142fc05322d71cf60dc/clinica/utils/testing_utils.py#L470-L474

NicolasGensollen:
I meant equality between outputs of subsequent runs of the pipeline.

AliceJoubert:
> I meant equality between outputs of subsequent runs of the pipeline.

Yes, I just took the lazy way and used the tests to get the metrics without having to code anything for it. 

> What if you just test for equality of the images without using a metric ?
> 
> You can use this function if needed:
> 
> https://github.com/aramis-lab/clinica/blob/31b2838ecc2f3c5336c5d142fc05322d71cf60dc/clinica/utils/testing_utils.py#L470-L474

Did not try that but with the images I have shown I guess it is not relevant anymore. Though I did not realize we had this module there, I could totally reuse some of these functions for non-regression tests !

NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1440, opened 2025-02-21
[FIX] Fix broken non regression test

NicolasGensollen:
Fix non regression test broken recently by #1433.


NicolasGensollen:
Oups, didn't realize it was done in #1438 

Closing...

PR #1443, opened 2025-02-21
[ENH] Separate converters and iotools in separate modules

### LINKED ISSUE #1447: Better architecture for `iotools` and `converters`
This issue is related to an envisioned architectural change consisting of having `converters` and `iotools` in their own modules instead of sharing the same `clinica.iotools` module.

The base idea is to have:

- the CLI `clinica convert` using code from the `clinica.converters` module
- the CLI `clinica iotools` using code from the `clinica.iotools` module
- the CLI `clinica run` using code from the `clinica.pipelines` module

NicolasGensollen:
## Description

Closes #1447 

This PR investigates the architectural change of having converters and iotools in their own modules instead of sharing the same `clinica.iotools` module.

The base idea is to have:

- the CLI `clinica convert` using code from the `clinica.converters` module 
- the CLI `clinica iotools` using code from the `clinica.iotools` module
- the CLI `clinica run` using code from the `clinica.pipelines` module

Objects used in multiple modules are moved in another top-level module (at the moment it is `clinica.utils` although this may evolve to something better one day...).

This is in our roadmap as the first step toward maybe splitting the converters of Clinica into their separate package one day.

## Things to do

- [x] Fix all the broken unit tests
- [x] Make sure non regression tests are passing


clinica/converters/genfi_to_bids/README.md
```diff
@@ -0,0 +1,86 @@
+<!-- markdownlint-disable MD046 -->
```
AliceJoubert, line 1: Is that file expected to be there ?

clinica/iotools/describe.py
```diff
@@ -5,7 +5,9 @@
 from rich.console import Console
 from rich.table import Table
 
-from clinica.iotools.bids_dataset_description import BIDSDatasetDescription
+from clinica.converters.bids_dataset_description import (
+    BIDSDatasetDescription,  # TODO: move that
```
AliceJoubert, line 9: Is the TODO still relevant ?

AliceJoubert (commented):
LGTM @NicolasGensollen, that will be much clearer. How would you feel about moving the files related to bids for the converters in a `bids_modelisation` (or similar) folder ?

clinica/converters/genfi_to_bids/README.md
NicolasGensollen: Hmmm, no it wasn't supposed to be committed but ended up getting caught by git. Will remove it. Thanks !

clinica/iotools/describe.py
NicolasGensollen: Yes, it's a TODO that we should address in a follow-up PR (see #1448).

NicolasGensollen:
> How would you feel about moving the files related to bids for the converters in a `bids_modelisation` (or similar) folder ?

I think something like that would be a good plan. 👍  
If it's OK with you, let's work on that specifically in another PR targeting issue #1448

clinica/converters/genfi_to_bids/README.md
NicolasGensollen: Done in 0be6e70

Issue #1444, opened 2025-02-27
Broken non regression tests


NicolasGensollen:
It seems like we have two broken tests at the moment: https://github.com/aramis-lab/clinica/actions/runs/13538941039/job/37904968563

```
=========================== short test summary info ============================
FAILED nonregression/iotools/test_run_utils.py::test_center_nifti - AssertionError: Following text files are different : 

 centered_nifti_list_20250225-151004.txt
FAILED nonregression/pipelines/anat/test_t1_linear.py::test_flair_linear - AssertionError: Following images do not meet the similarity criteria : 

sub-ADNI941S6580_ses-M000_space-MNI152NLin2009cSym_desc-Crop_res-1x1x1_FLAIR.nii.gz
sub-ADNI941S6580_ses-M000_space-MNI152NLin2009cSym_res-1x1x1_FLAIR.nii.gz
sub-02_ses-M000_run-123_space-MNI152NLin2009cSym_desc-Crop_res-1x1x1_FLAIR.nii.gz
sub-02_ses-M000_run-123_space-MNI152NLin2009cSym_res-1x1x1_FLAIR.nii.gz
= 2 failed, 12 passed, 1 skipped, 30 deselected, 134 warnings in 3546.52s (0:59:06) =
```

@AliceJoubert I think it is due to the recent work done in #1438 #1439 and #1441

I haven't looked in details but I think there are two distinct issues: one with the text file comparison and one with the nifti comparison (where it seems like the function isn't comparing the right images).

Do you think you can take a closer look ? (I you're busy with other things I can try to find some time to look into it).

AliceJoubert:
I can look at it today

PR #1445, opened 2025-02-27
[FIX] Sort list of centered files to be written in the text file produced by `center-nifti`

AliceJoubert:
Partly adresses #1444


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

PR #1453, opened 2025-03-07
[ENH] Rework architecture specific to BIDS and CAPS dataset

### LINKED ISSUE #1448: Improve architecture of `clinica.iotools` and `clinica.converters`
This is a reminder to perform some cleaning and improvements of code when #1443 will be merged.

This is especially true around things related to BIDS which are still defined in multiple places (`clinica.utils.bids`, `clinica.converters._utils` etc...).

It's very likely that the end goal is to put all of this logic in `clinicaIO`, but in the mean time we can first work on grouping these things within Clinica to later ease the extraction.

NicolasGensollen:
Closes #1448 

Following #1443, this PR continues the re-organization of the code relative to `iotools` and `converters`. More precisely, it targets the code handling BIDS and CAPS datasets which is scattered across the code base (some in `clinica.converters`, some in `clinica.iotools`, some in `clinica.utils`...).

The base idea of this PR is to regroup all this code in a separate module called `clinica.datasets` (name can change if needed). This module has all the logic related to BIDS and CAPS datasets (validation, queries, writing, parsing...).
The logic specific to BIDS datasets is in the `clinica.datasets.bids` submodule while the logic specific to CAPS is in `clinica.dataset.caps`. Logic handling both is in `clinica.datasets`.

Going through these pieces of code, it seems we have a lot of small functions doing more or less the same things, and sometimes in ways that we don't need anymore. An example is the functions trying to guess whether a folder is a BIDS or CAPS dataset by doing some heuristic checks. Since the introduction of the MANDATORY `dataset_description.json` file for both types of  datasets, we don't need to guess anymore, we have the information written in the metadata, so we "just" need to parse it.

This is still ongoing work, lots of tests are broken at the moment...


NicolasGensollen:
Alright, looks like all tests are passing again. 
@AliceJoubert sorry again for the large PR. 
Let me know what you think when you have some time to make a review. Again, I've made some choices but we can discuss them if you feel there are better alternatives.

Issue #1450, opened 2025-02-28
Improve Test Coverage Description


AliceJoubert:
## Description

Improve Clinica test coverage. For non-regression tests : add comparison of images and produced side files.

### Non-regression in converters : _(should we add Bids Validator ?)_
- [ ] ADNI : ::todo when refactored
- [x] UKB : #1526
- [x] AIBL : #1424 
- [x] GENFI : #1468
- [x] HABS : #1471
- [x] NIFD : #1472
- [x] OASIS : #1465
- [x] OASIS3 : #1479
- [x] IXI : #1463

### Non-regression in pipelines : 
- [x] Anatomical linear : 
    - #1428,
    - #1439,
    - #1446
- [ ] Anatomical volume
- [ ] Anatomical FreeSurfer
- [ ] Statistics
- [ ] DWI
- [ ] PET

### Non-regression in IOtools : 
- [x] CreateSubjectsVisits
- [x] Check Missing Modalities
- [ ] Check Missing Processing
- [x] Merge TSV
- [x] Center Nifti : 
   - #1438,
   - #1441
- [ ] Describe

### Unit testing
- Improve coverage overall

PR #1465, opened 2025-03-11
[TEST] OASIS-to-BIDS : add non-regression tests for metadata files

AliceJoubert:
Extends non regression tests to metadata files as part of https://github.com/aramis-lab/clinica/issues/1450.

### DIFF
--- a/test/nonregression/testing_tools.py
+++ b/test/nonregression/testing_tools.py
@@ -530,7 +530,9 @@ def _loader_factory(level: Union[str, Level]) -> LoaderInterface:
 def _compare_frames(df1: pd.DataFrame, df2: pd.DataFrame, object: str):
     from pandas.testing import assert_frame_equal
 
-    assert_frame_equal(df1, df2, check_like=True, obj=object)
+    assert_frame_equal(
+        df1, df2, check_like=True, obj=object, check_dtype=False, check_exact=True
+    )
 
 
 def _iteratively_compare_frames(bids_ref: Path, bids_out: Path, level: Level):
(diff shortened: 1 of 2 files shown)


test/nonregression/testing_tools.py
```diff
@@ -530,7 +530,9 @@ def _loader_factory(level: Union[str, Level]) -> LoaderInterface:
 def _compare_frames(df1: pd.DataFrame, df2: pd.DataFrame, object: str):
     from pandas.testing import assert_frame_equal
 
-    assert_frame_equal(df1, df2, check_like=True, obj=object)
+    assert_frame_equal(
+        df1, df2, check_like=True, obj=object, check_dtype=False, check_exact=True
```
NicolasGensollen, line 534: Why not checking the data type ?

NicolasGensollen (commented):
Thanks for pursuing this work @AliceJoubert !

I just have a small question. Looks good otherwise.

test/nonregression/testing_tools.py
AliceJoubert: I assumed it is not a big deal if the type for a value is different (float or int for example) if the value is the same (verified with `check_exact` normally). If you think it is I will update the converter accordingly

NicolasGensollen: But, you had to make this change because you got an error (like an int mistaken for a float) for OASIS ?

Ideally, I think we should make sure to have consistent data types. If this is painful to do, we can merge like this.
It's already much much better than no testing at all ! 👍

AliceJoubert: Yes it's for OASIS. I looked quickly how to change it so we can have the same datatypes but it's not so easy, which is why I did not do it. If we can function this way for now it would be quite convenient !

NicolasGensollen: Alright ! Let's move on like this then !

NicolasGensollen (approved):

PR #1470, opened 2025-03-14
[FIX] Change querying pattern in `bids_pet_nii` when `tracer` exists but `reconstruction` is None

### LINKED ISSUE #1469: Wrong pattern in `bids_pet_nii`
The function 
https://github.com/aramis-lab/clinica/blob/f7a82994761bcd85f70cfd346cd5bfbce959d2b4/clinica/utils/input_files.py#L585-L588

Returns the pattern to pass to the clinica file reader in order to get the PET scans from a BIDS folder. It takes two optional parameters: the tracer and the reconstruction method. If there is no value provided for one or two of these parameters, the function is supposed to return a pattern matching all possible values for the entity:

https://github.com/aramis-lab/clinica/blob/f7a82994761bcd85f70cfd346cd5bfbce959d2b4/clinica/utils/input_files.py#L618-L625

This works as expected for most cases but not in the special case where the user provides a tracer but not a reconstruction method. In this situation, the returned pattern is `*{tracer}_pet.nii*` while it should be `*{tracer}*_pet.nii*` in order to match any reconstruction method.

AliceJoubert:
Close #1469

### DIFF
--- a/clinica/utils/input_files.py
+++ b/clinica/utils/input_files.py
@@ -611,15 +611,17 @@ def bids_pet_nii(
 
     description = f"PET data"
     trc = ""
+    rec = ""
+
     if tracer is not None:
         tracer = Tracer(tracer)
         trc = f"_trc-{tracer.value}"
         description += f" with {tracer.value} tracer"
-    rec = ""
+        rec = "*"
     if reconstruction is not None:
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
LGTM, thanks for fixing this quickly @AliceJoubert !

PR #1471, opened 2025-03-14
[TEST] HABS-to-BIDS : Add comparison of metadata files in non-regression tests

AliceJoubert:
Extended non-regression tests for metadata files as part of #1450


NicolasGensollen (approved):
Thanks @AliceJoubert !
LGTM, if the non regression tests for converters are passing, feel free to merge.

PR #1474, opened 2025-03-17
[CI] OASIS-to-BIDS : Add option --subjects-list and its functionality for oasis-to-bids converter

### LINKED ISSUE #1467: Add option `--subjects-list` for `oasis-to-bids` converter
Some converters like `adni-to-bids` have an option to provide a list of subjects to convert:

https://github.com/aramis-lab/clinica/blob/f7a82994761bcd85f70cfd346cd5bfbce959d2b4/clinica/converters/adni_to_bids/cli.py#L15

Which seems to be missing from the `oasis-to-bids` converter:

https://github.com/aramis-lab/clinica/blob/f7a82994761bcd85f70cfd346cd5bfbce959d2b4/clinica/converters/oasis_to_bids/cli.py#L9-L20

It would be nice to add this option for various reasons. First, this would give us a more consistent API across converters. Second, it would allow users to convert only a subset of the OASIS subjects even though this is less useful than for ADNI.

The subjects to convert are computed here:

https://github.com/aramis-lab/clinica/blob/f7a82994761bcd85f70cfd346cd5bfbce959d2b4/clinica/converters/oasis_to_bids/_converter.py#L212-L216

Basically, everything found in the raw data will be converted. In order to implement this option, this list will probably be filtered by the user-provided list of subjects.

Adam-Ismaili-92:
Closes #1467

### DIFF
--- a/clinica/converters/oasis_to_bids/_utils.py
+++ b/clinica/converters/oasis_to_bids/_utils.py
@@ -17,65 +17,61 @@ __all__ = [
 ]
 
 
-def _define_subjects_list(
-    source_dir: Path,
-    subjs_list_path: Optional[Path] = None,
-) -> List[Path]:
-    import re
+def _get_subjects_from_data(source_dir: Path) -> list[str]:
+    return [
+        folder.name
+        for folder in source_dir.iterdir()
+        if not folder.name.startswith(".")
+    ]
 
-    from clinica.utils.stream import cprint
+
+def _get_subjects_from_list(subjects_list_path: Path) -> list[str]:
+    return subjects_list_path.read_text().splitlines()
+
+
+def _filter_oasis_subjects(source_dir: Path, subjects_list: list[str]) -> list[Path]:
+    import re
 
     rgx = re.compile(r"OAS1_\d{4}_MR1")
 
-    if subjs_list_path:
-        cprint("Loading a subjects lists provided by the user...")
-        return list(
-            filter(
-                lambda path: path.is_dir(),
-                list(
-                    map(
-                        Path,
-                        [
-                            source_dir / subj
-                            for subj in list(
-                                filter(
-                                    rgx.fullmatch,
-                                    subjs_list_path.read_text().splitlines(),
-                                )
-                            )
-                        ],
-                    )
-                ),
-            )
-        )
-
-    cprint(f"Using the subjects contained in the OASIS dataset at {source_dir}")
     return list(
         filter(
             lambda path: path.is_dir(),
-            list(
-                map(
-                    Path,
-                    [
-                        source_dir / subj
-                        for subj in list(
-                            filter(
-                                rgx.fullmatch,
-                                [folder.name for folder in source_dir.iterdir()],
-                            )
-                        )
-                    ],
+            [
+                source_dir / subj
+                for subj in filter(
+                    rgx.fullmatch,
+                    subjects_list,
                 )
-            ),
+            ],
         )
     )
 
 
 def get_subjects_list(
-    source_dir: Path,
-    subjs_list_path: Optional[Path] = None,
-) -> List[Path]:
-    return _define_subjects_list(source_dir, subjs_list_path)
+    source_dir: Path, subjs_list_path: Optional[Path] = None
+) -> list[Path]:
+    """Gets the list of paths to the subjects folders.
+
+    Parameters
+    ----------
+    source_dir : Path
+        The path to the input dataset folder.
+
+    subjs_list_path : Optional[Path]
+        The path to the subjects list file.
+
+    Returns
+    -------
+    list[Path] :
+        List of paths to the subjects folders.
+    """
+    if subjs_list_path:
+        return _filter_oasis_subjects(
+            source_dir, _get_subjects_from_list(subjs_list_path)
+        )
+
+    return _filter_oasis_subjects(source_dir, _get_subjects_from_data(source_dir))
 
 
 def _convert_cdr_to_diagnosis(cdr: Union[int, str]) -> str:
--- a/clinica/converters/oasis_to_bids/_utils.py
+++ b/clinica/converters/oasis_to_bids/_utils.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Iterable, List, Optional, Union
+from typing import Iterable, Optional, Union
 
 import nibabel as nb
 import numpy as np
@@ -17,16 +17,12 @@ __all__ = [
 ]
 
 
-def _get_subjects_from_data(source_dir: Path) -> list[str]:
+def _get_subjects_list_from_data(source_dir: Path) -> list[str]:
     return [
         folder.name
         for folder in source_dir.iterdir()
         if not folder.name.startswith(".")
     ]
-
-
-def _get_subjects_from_list(subjects_list_path: Path) -> list[str]:
-    return subjects_list_path.read_text().splitlines()
 
 
 def _filter_oasis_subjects(source_dir: Path, subjects_list: list[str]) -> list[Path]:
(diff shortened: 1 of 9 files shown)


clinica/converters/oasis_to_bids/_utils.py
```diff
@@ -1,18 +1,80 @@
...
 __all__ = [
+    "get_subjects_list",
     "create_sessions_df",
     "write_sessions_tsv",
     "write_scans_tsv",
     "get_first_image",
 ]
 
 
+def _define_subjects_list(
```
AliceJoubert, line 17: If you follow my suggestions (see below) you will see there is a common filter between the two options within the function. In both cases you have a list that can be filtered with the regex.

You could change the approach to the problem as such : 
```python
def _get_subjects_from_data(source_dir:Path)->list(str):
    return [folder.name for folder in source_dir.iterdir()]

def _get_subjects_from_list(subjects_list_path:Path)->list(str):
    return subjs_list_path.read_text().splitlines()

def _filter_oasis_subjects(subjects_list:list(str))->list(str):
   # use regex to filters subjects
   return ...

def get_subjects_list(source_dir:Path, subjs_list_path: Optional[Path] = None)->list(Path):
    if subjs_list_path:
       return [source_dir/subject for subject in _filter_oasis_subjects(_get_subjects_from_list(source_dir)) if source_dir/subject.is_dir()]
    return [source_dir/subject for subject in _filter_oasis_subjects(_get_subjects_from_data(source_dir))]
```


This way, you can test each and every function properly. In any case if you don't choose this option it is necessary to unit test your functions !

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 ! I have a few suggestions, let me know what you think.

clinica/converters/oasis_to_bids/_utils.py
Adam-Ismaili-92: I've added the corrections you suggested, I forgot to add unit tests indeed, I'll do it right away.

clinica/converters/oasis_to_bids/_utils.py
```diff
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Iterable, Union
+from typing import Iterable, List, Optional, Union
```
AliceJoubert, line 2: Do you still need the `List` import ?

clinica/converters/oasis_to_bids/_utils.py
```diff
@@ -16,6 +17,63 @@
...
 
+def _get_subjects_from_data(source_dir: Path) -> list[str]:
+    return [
+        folder.name
+        for folder in source_dir.iterdir()
+        if not folder.name.startswith(".")
+    ]
+
+
+def _get_subjects_from_list(subjects_list_path: Path) -> list[str]:
```
AliceJoubert, line 28: Now that this function looks like this, I think there is a function in IXI-to-BIDS that looks exactly the same : 

https://github.com/aramis-lab/clinica/blob/353ede19ba53d2e70e0f0b148eb88fbf3be6b2ed/clinica/converters/ixi_to_bids/_utils.py#L44-L45

There is no use doing a duplicate function so maybe you could use the one from IXI. Since it would be then used by several converters it would need to be moved up in the modules hierarchy and be placed in the `utils` for all converters.

test/unittests/converters/test_oasis_to_bids_utils.py
```diff
@@ -310,6 +310,55 @@ def test_get_image_with_good_orientation(tmp_path: Path) -> None:
     assert len(image.shape) == 3  # the image dimension should be 3D
 
 
+def test_get_subjects_list_from_data(tmp_path: Path) -> None:
+    from clinica.converters.oasis_to_bids._utils import get_subjects_list
+
+    source_dir = tmp_path / "dataset"
+    source_dir.mkdir()
+
+    for filename in [
```
AliceJoubert, line 319: I am nitpicking but you can use a tuple `()` instead of a list here as you do not need list complexity

test/unittests/converters/test_oasis_to_bids_utils.py
```diff
@@ -310,6 +310,55 @@ def test_get_image_with_good_orientation(tmp_path: Path) -> None:
     assert len(image.shape) == 3  # the image dimension should be 3D
 
 
+def test_get_subjects_list_from_data(tmp_path: Path) -> None:
```
AliceJoubert, line 313: You can add the case where `"OAS1_0001_MR1"` is not a directory since it is what you are filtering based on this in your functions

AliceJoubert, line 313:
```suggestion
def test_get_subjects_list_from_data(tmp_path) -> None:
```

In test functions you do not need to define the type of `tmp_path` as it is a pytest fixture (same below). You would need to define types in helper functions if you ever write one though

test/unittests/converters/test_oasis_to_bids_utils.py
```diff
@@ -310,6 +310,55 @@ def test_get_image_with_good_orientation(tmp_path: Path) -> None:
...
+        "OAS1_0003_MR2",
+        "OAS1_004_MR1",
+        "foo",
+    ]:
+        (source_dir / filename).mkdir()
+
+    assert get_subjects_list(source_dir) == [source_dir / "OAS1_0001_MR1"]
+
+
+def test_get_subjects_list_from_file(tmp_path: Path) -> None:
```
AliceJoubert, line 331:
```suggestion
def test_get_subjects_list_from_file(tmp_path) -> None:
```

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 ! Just a few more details but for the code part it should be good to go.

clinica/converters/oasis_to_bids/_utils.py
Adam-Ismaili-92: No, it has been removed in the newest commit

clinica/converters/oasis_to_bids/_utils.py
Adam-Ismaili-92: You're indeed right, the duplicate has been deleted and the original function has been moved up in the hierarchy.

test/unittests/converters/test_oasis_to_bids_utils.py
Adam-Ismaili-92: I didn't know that lists were more complex than tuples, the suggestion is indeed interesting and applied in the last commit.

test/unittests/converters/test_oasis_to_bids_utils.py
Adam-Ismaili-92: Thanks for the suggestion, I added a similar case where one file is correctly named but is not a directory in the last commit.

test/unittests/converters/test_ixi_to_bids_utils.py
```diff
@@ -30,12 +30,12 @@ def test_get_subjects_list_from_data(tmp_path):
 
 
 def test_get_subjects_list_from_file(tmp_path):
-    from clinica.converters.ixi_to_bids._utils import _get_subjects_list_from_file
+    from clinica.converters._utils import get_subjects_list_from_file
```
AliceJoubert, line 33: Can you move this one to `test_converters_utils`?

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 . Just one last detail before merging

test/unittests/converters/test_ixi_to_bids_utils.py
Adam-Ismaili-92: No problem, it has been done in the next commit normally.

AliceJoubert (approved):
LGTM ! Thanks for the work @Adam-Ismaili-92

PR #1476, opened 2025-03-25
[FIX][DOC] Fix options for all pipelines in `Interacting With Clinica`

AliceJoubert:


Adam-Ismaili-92 (approved):
Simple change in the code, respecting the coding style.
PR functionally tested, it is replacing well the "markdown codes" near the options by "listing dots".

PR #1479, opened 2025-03-25
[TEST] OASIS3-to-BIDS : Add comparison of metadata files in non-regression tests

AliceJoubert:
Extended non-regression tests for metadata files as part of #1450


Adam-Ismaili-92 (approved):
Simple line of code added.
Non regression tests passed for OASIS3-to-BIDS after adding metadata files comparison.

PR #1480, opened 2025-03-26
[FIX] Remove the `--yes` option from the `t1-volume` pipeline as it is not used anymore

AliceJoubert:
Following #1419, this option should not be in the client for the pipeline anymore.


Adam-Ismaili-92 (approved):
Option well removed.
Clear instructions added to documentation.

Issue #1478, opened 2025-03-25
Harmonize option `--subjects_list` over converters


AliceJoubert:
Following #1467, converters for ADNI, OASIS1 and IXI should have the `-sl` option. Implementation was already different between ADNI and IXI, and OASIS1 added another one. Maybe this would be the time to think of a way to proceed the same way for each converter if possible.

### To discuss : 

- Way of discriminating subjects :
  - ADNI : based on presence in clinical data
  - OASIS1/IXI : based on name in folder

- Structure : 

Could be similar to what is done in IXI : 

```python
# Use these functions for all converters

def _get_subjects_list_from_file(subjs_list_path: Path) -> List[str]:
    return subjs_list_path.read_text().splitlines()

def define_participants(
    data_directory: Path,
    subjs_list_path: Optional[Path] = None,
) -> List[str]:
    """
    Defines the actual list of participants based on the (provided) subjects list filtered
    using existing data.

    Parameters
    ----------
    data_directory : Path to the raw data directory.
    subjs_list_path : [Optional] Path to a text file containing a list of specific subjects to extract.

    Returns
    -------
    The list of ids that were either present in the data directory or asked for with a specific text file, provided
    associated images actually exists.

    """

    list_from_data = _get_subjects_list_from_data(data_directory)
    if subjs_list_path is None:
        return sorted(list_from_data)
    cprint("Loading a subjects list provided by the user...")
    list_from_file = _get_subjects_list_from_file(subjs_list_path)
    list_filtered = [subject for subject in list_from_file if subject in list_from_data]
    invalid_subjects = list(set(list_from_file) - set(list_filtered))
    if invalid_subjects:
        cprint(
            f"The subjects : {' , '.join(invalid_subjects)} do not have any associated data inside the directory {data_directory}"
            f" and can not be converted."
        )
    return sorted(list_filtered)
```

```python
# Have a function specific to the converter filtering on the name/clinical data/... EX for IXI

def _get_subjects_list_from_data(data_directory: Path) -> List[str]:
    return list(
        dict.fromkeys(
            re.match(r"IXI\d{3}", path.name).group(0)
            for path in data_directory.rglob(pattern="IXI*.nii.gz")
            if re.match(r"IXI\d{3}", path.name)
        )
    )
```

- The output list can also be checked using the modelisation of BIDS IDs we did with method `.validate` : we will need to decide if we raise an error or just warn the user if an id is wrong

[...]

PR #1482, opened 2025-03-27
[BUG] OASIS-to-BIDS informative logs made more relevant based on --subjects_list option

### LINKED ISSUE #1322: [BUG] OASIS-to-BIDS informative logs are not relevant
Currently, the converter only indicates : 
- Clinical data conversion
- The difference between the result and clinical data (see issue #249) -> too much noise !
- Missing metadata for all subjects in **clinical data** -> too much noise too !

Adam-Ismaili-92:
Closes #1322

### DIFF
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -602,4 +602,4 @@ def compare_bids_ids_lists(
     list[BIDSSubjectID] :
         Bids ids difference.
     """
-    return list(set(expected_subjects).symmetric_difference(set(obtained_subjects)))
+    return list(set(expected_subjects) - set(obtained_subjects))
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -57,6 +57,7 @@ def create_participants_df(
     clinical_data_dir: Path,
     bids_ids: list[str],
     delete_non_bids_info: bool = True,
+    subjects: Optional[Path] = None,
 ) -> pd.DataFrame:
     """Create the file participants.tsv.
 
@@ -94,6 +95,7 @@ def create_participants_df(
     prev_sheet = 0
     index_to_drop = []
     subjects_to_drop = []
+    subjects_list = []
     study_name = StudyName(study_name)
     location_name = f"{study_name.value} location"
 
@@ -182,6 +184,13 @@ def create_participants_df(
         ]
     participant_df.reset_index(inplace=True, drop=True)
 
+    # Extracting subjects list from the input subj list file
+    if subjects:
+        import re
+
+        rgx = re.compile(r"OAS1_\d{4}_MR1")
+        subjects_list = filter(rgx.fullmatch, get_subjects_list_from_file(subjects))
+
     # Adding participant_id column with BIDS ids
     for i in range(0, len(participant_df)):
         value = bids_id_factory(study_name).from_original_study_id(
@@ -190,7 +199,11 @@ def create_participants_df(
         bids_id = [s for s in bids_ids if value in s]
         if len(bids_id) == 0:
             index_to_drop.append(i)
-            subjects_to_drop.append(value)
+
+            if (
+                subjects and any(value[-4:] == subj[5:9] for subj in subjects_list)
+            ) or (not subjects):
+                subjects_to_drop.append(value)
         else:
             participant_df.at[i, "participant_id"] = bids_id[0]
 
@@ -200,6 +213,11 @@ def create_participants_df(
                 "The following subjects of dataset directory were not found in your BIDS folder :\n"
                 + ", ".join(subjects_to_drop)
             ),
+            lvl="info",
+        )
+    else:
+        cprint(
+            msg=("All subjects of dataset directory were found in your BIDS folder."),
             lvl="info",
         )
     # Delete all the rows of the subjects that are not available in the BIDS dataset
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -23,7 +23,7 @@ __all__ = [
     "viscode_to_session",
     "load_clinical_csv",
     "get_subjects_list_from_file",
-    "compare_bids_ids",
+    "compare_bids_ids_lists",
 ]
 
 
@@ -584,17 +584,17 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
     return subjects_list_path.read_text().splitlines()
 
 
-def compare_bids_ids(
-    bids_ids_1: list[BIDSSubjectID], bids_ids_2: list[BIDSSubjectID]
+def compare_bids_ids_lists(
+    expected_subjects: list[BIDSSubjectID], obtained_subjects: list[BIDSSubjectID]
 ) -> list[BIDSSubjectID]:
     """Gets the difference between bids ids lists.
 
     Parameters
     ----------
-    bids_ids_1 : list[BIDSSubjectID]
+    expected_subjects : list[BIDSSubjectID]
         First bids ids list.
 
-    bids_ids_2 : list[BIDSSubjectID]
+    obtained_subjects : list[BIDSSubjectID]
         Second bids ids list.
 
     Returns
@@ -602,4 +602,4 @@ def compare_bids_ids(
     list[BIDSSubjectID] :
         Bids ids difference.
     """
-    return list(set(bids_ids_1).symmetric_difference(set(bids_ids_2)))
+    return list(set(expected_subjects).symmetric_difference(set(obtained_subjects)))
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -10,7 +10,7 @@ import pandas as pd
 
 from clinica.utils.filemanip import UserProvidedPath
 
-from .study_models import StudyName
+from .study_models import BIDSSubjectID, StudyName
 
 __all__ = [
     "create_participants_df",
@@ -93,6 +93,7 @@ def create_participants_df(
 
     from clinica.utils.stream import cprint
 
+    from .oasis_to_bids._utils import _get_subjects_list_from_file
     from .study_models import bids_id_factory
 
     fields_bids = ["participant_id"]
@@ -100,9 +101,8 @@ def create_participants_df(
     prev_sheet = 0
     index_to_drop = []
     subjects_to_drop = []
-    subjects_list = []
+    subjects_set = {}
     study_name = StudyName(study_name)
-    subjects_pattern = get_subject_id_pattern(study_name)
     location_name = f"{study_name.value} location"
 
     participants_specs = pd.read_csv(
@@ -192,44 +192,25 @@ def create_participants_df(
 
     participant_df.reset_index(inplace=True, drop=True)
 
-    # Extracting subjects list from the input subjects list file
-    if subjects and subjects_pattern:
-        import re
-
-        rgx = re.compile(subjects_pattern)
-        subjects_list = filter(rgx.fullmatch, get_subjects_list_from_file(subjects))
+    if subjects:
+        subjects_set = set(_get_subjects_list_from_file(subjects))
 
     # Adding participant_id column with BIDS ids
     for i in range(0, len(participant_df)):
-        value = bids_id_factory(study_name).from_original_study_id(
-            participant_df["alternative_id_1"][i]
-        )
-        bids_id = [s for s in bids_ids if value in s]
+        bids_id_from_participant_df = bids_id_factory(
+            study_name
+        ).from_original_study_id(participant_df["alternative_id_1"][i])
 
-        # If the conversion to BIDS has failed for the processed subject
-        if len(bids_id) == 0:
-            # If a subjects list is provided by the user
-            if subjects:
-                # If the subject is from the OASIS dataset
-                if study_name == StudyName.OASIS:
-                    # The 4 ID digits from the subject name converted to BIDS
-                    bids_subj_digits = value[-4:]
-
-                    # List of 4 ID digits from the subject names in the subjects list
-                    raw_subj_digits_list = [subj[5:9] for subj in subjects_list]
-
-                    # If one matching groups of digits is found, indicate that the conversion has failed
-                    if bids_subj_digits in raw_subj_digits_list:
-                        index_to_drop.append(i)
-                        subjects_to_drop.append(value)
-
-            # If a subjects list is not provided, indicate that the conversion has failed
-            else:
-                index_to_drop.append(i)
-                subjects_to_drop.append(value)
+        if bids_id_from_participant_df in bids_ids:
+            participant_df.at[i, "participant_id"] = bids_id_from_participant_df
 
         else:
-            participant_df.at[i, "participant_id"] = bids_id[0]
+            index_to_drop.append(i)
+
+            if (
+                subjects and bids_id_from_participant_df in subjects_set
+            ) or not subjects:
+                subjects_to_drop.append(bids_id_from_participant_df)
 
     if len(subjects_to_drop) > 0:
         cprint(
@@ -630,48 +611,3 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
         List of subjects.
     """
     return subjects_list_path.read_text().splitlines()
-
-
-def get_subject_id_pattern(study_name: StudyName) -> str:
-    """Gets subjects pattern for each variant of StudyName enumeration.
-
-    Parameters
-    ----------
-    study_name : StudyName
-        StudyName enumeration.
-
-    Returns
-    -------
-    str :
-        Subjects pattern.
-    """
-
-    if study_name == StudyName.ADNI:
-        return r"\d{3}_S_\d{4}"
-
-    elif study_name == StudyName.AIBL:
-        return r"\d+"
-
-    # elif study_name == StudyName.GENFI :
-    #   TO COMPLETE
-
-    # elif study_name == StudyName.HABS :
-    #   TO COMPLETE
-
-    elif study_name == StudyName.IXI:
-        return r"IXI\d+-[a-zA-Z]+-\d{4}-(DTI|MRA|PD|T1|T2)"
-
-    elif study_name == StudyName.NIFD:
-        return r"\d{1}_S_\d{4}"
-
-    elif study_name == StudyName.OASIS:
-        return r"OAS1_\d{4}_MR1"
-
-    elif study_name == StudyName.OASIS3:
-        return r"OAS3\d{4}_(CT|MR|PIB|AV45|FDG)_d\d{4}"
-
-    # elif study_name == StudyName.UKB :
-    #   TO COMPLETE
-
-    else:
-        return None
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -192,7 +192,7 @@ def create_participants_df(
 
     participant_df.reset_index(inplace=True, drop=True)
 
-    if subjects:
+    if subjects and study_name == StudyName.OASIS:
         subjects_set = set(_get_subjects_list_from_file(subjects))
 
     # Adding participant_id column with BIDS ids
--- a/clinica/converters/_utils.py
+++ b/clinica/converters/_utils.py
@@ -24,6 +24,7 @@ __all__ = [
     "load_clinical_csv",
     "get_subjects_list_from_file",
     "get_subject_id_pattern",
+    "compare_bids_ids",
 ]
 
 
@@ -100,8 +101,6 @@ def create_participants_df(
     prev_location = ""
     prev_sheet = 0
     index_to_drop = []
-    subjects_to_drop = []
-    subjects_set = {}
     study_name = StudyName(study_name)
     location_name = f"{study_name.value} location"
 
@@ -192,9 +191,6 @@ def create_participants_df(
 
     participant_df.reset_index(inplace=True, drop=True)
 
-    if subjects and study_name == StudyName.OASIS:
-        subjects_set = set(_get_subjects_list_from_file(subjects))
-
     # Adding participant_id column with BIDS ids
     for i in range(0, len(participant_df)):
         bids_id_from_participant_df = bids_id_factory(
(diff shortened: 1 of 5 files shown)


clinica/converters/_utils.py
```diff
@@ -182,6 +184,13 @@ def create_participants_df(
         ]
     participant_df.reset_index(inplace=True, drop=True)
 
+    # Extracting subjects list from the input subj list file
+    if subjects:
+        import re
+
+        rgx = re.compile(r"OAS1_\d{4}_MR1")
```
NicolasGensollen, line 191: This regex is specific to OASIS but, since this function is used by several converters, you can end up here for another converter like adni-to-bids or aibl-to-bids which will be problematic.

So, maybe move this under the "elif study_name == StudyName.OASIS:" above ?

Another solution would be to have a pattern factory which would return the pattern for each variant of the StudyName enumeration, such that this would become something like this:

```python
rgx = re.compile(get_subject_id_pattern(StudyName.OASIS)
subjects_list = ...
```

WDYT ?

clinica/converters/_utils.py
```diff
@@ -190,18 +199,27 @@ def create_participants_df(
         bids_id = [s for s in bids_ids if value in s]
         if len(bids_id) == 0:
             index_to_drop.append(i)
-            subjects_to_drop.append(value)
+
+            if (
+                subjects and any(value[-4:] == subj[5:9] for subj in subjects_list)
```
NicolasGensollen, line 204: This is hard to understand and it looks very fragile. Could you explain what you want to do here ?

clinica/converters/_utils.py
```diff
@@ -57,6 +57,7 @@ def create_participants_df(
     clinical_data_dir: Path,
     bids_ids: list[str],
     delete_non_bids_info: bool = True,
+    subjects: Optional[Path] = None,
```
NicolasGensollen, line 60: Could you add this new argument to the docstring with a proper description ?

clinica/converters/_utils.py
```diff
@@ -182,6 +184,13 @@ def create_participants_df(
         ]
     participant_df.reset_index(inplace=True, drop=True)
 
+    # Extracting subjects list from the input subj list file
```
NicolasGensollen, line 187:
```suggestion
    # Extracting subjects list from the input subjects list file
```

NicolasGensollen (commented):
Thanks @Adam-Ismaili-92 !

The `create_participants_df` is an example of what code shouldn't look like: a giant untested function with super long loops and tons of variables and mutable objects....
In this context, it's hard to know whether the code we add is breaking stuffs or not. So, it's not an easy terrain to work in and to review.

I made some comments to clarify a few things but it would be great if you could add a unit test to confirm the behavior when a list of subjects is provided.
I think @AliceJoubert recently added some tests here:

https://github.com/aramis-lab/clinica/blob/25e79021e0e0feac208a3adac262961f5c6f0853/test/unittests/converters/test_converter_utils.py#L178

might be a good place to start...

Also, we really need to break down this function in smaller chunks with proper tests (it's already on the todo list, but we should go for it at some point).

clinica/converters/_utils.py
Adam-Ismaili-92: Since the name of the subject retrieved by the bids_id_factory function has the pattern "sub-OASIS1\d{4}" (the variable storing the name is called "value"), and the name of the subjects entered in the input file is of the form "OAS1-\d{4}-MR1", I compare the 4 common digits between the two patterns. I then iterate over my list of subjects by comparing the 4 digits present on each entry with the last 4 digits of my "value" variable. 

Thus, if the BIDS conversion has failed for the processed subject (upper if condition), and a subjects list file is provided by the user, and one matching entry is found for the processed subject name (= so the subject name is registered in the subjects list but has not been converted), the subject name is then added to the log indicating that it has not been found in the BIDS folder.

It is a thing only working for the case of OASIS-to-BIDS converter nonetheless, I should add a condition to better frame this part, and commentaries to better explain it.

clinica/converters/_utils.py
Adam-Ismaili-92: I have chosen the second solution of creating a pattern factory which would return the pattern of each variant of StudyName enumeration, for more generalisability.

But some regex are missing since I do not have proper access to all datasets yet.

Adam-Ismaili-92:
> Thanks @Adam-Ismaili-92 !
> 
> The `create_participants_df` is an example of what code shouldn't look like: a giant untested function with super long loops and tons of variables and mutable objects.... In this context, it's hard to know whether the code we add is breaking stuffs or not. So, it's not an easy terrain to work in and to review.
> 
> I made some comments to clarify a few things but it would be great if you could add a unit test to confirm the behavior when a list of subjects is provided. I think @AliceJoubert recently added some tests here:
> 
> https://github.com/aramis-lab/clinica/blob/25e79021e0e0feac208a3adac262961f5c6f0853/test/unittests/converters/test_converter_utils.py#L178
> 
> might be a good place to start...
> 
> Also, we really need to break down this function in smaller chunks with proper tests (it's already on the todo list, but we should go for it at some point).

Thanks for your reviews and suggestions. I will try to add tests in the next commits.

clinica/converters/_utils.py
```diff
@@ -78,6 +80,10 @@ def create_participants_df(
         If True delete all the rows of the subjects that are not available in the BIDS dataset.
         Default=True.
 
+    subjects : Path, optional
+        Path to the input subjects list provided by the user.
+        Default=None.
+
```
NicolasGensollen, line 86: I think this function is getting way too complicated for what it's doing. In terms of management of `bids_id`, they are extracted from the participants dataframe, one row at the time, and the row can be dropped if:

- the `bids_id` is not in the list `bids_ids` passed as a parameter
- the `bids_id` is not in the list provided as a file through the parameter `subjects`

In addition, all these IDs can be plain strings or proper `BIDSSubjectID` which adds to the confusion.

I think we need to simplify things by:

- Building a single set of `BIDSSubjectID` from parameters `bids_ids` and `subjects`
- Simply check, for each row of the participant data frame, if the ID is in this filtering set or not.

clinica/converters/_utils.py
```diff
@@ -94,7 +100,9 @@ def create_participants_df(
     prev_sheet = 0
     index_to_drop = []
     subjects_to_drop = []
+    subjects_list = []
```
NicolasGensollen, line 103: Here we handle the `bids_ids` parameter of the function:

Ideally, the parameter `bids_ids` would already be a `list[BIDSSubjectID]` and not a `list[str]`. This is something we could improve later, but we can still convert this list of strings to a list of proper objects:

```python
bids_ids: list[BIDSSubjectID] = [
    bids_id_factory(study_name)(bids_id) for bids_id in bids_ids
]
```

clinica/converters/_utils.py
```diff
@@ -94,7 +100,9 @@ def create_participants_df(
...
     study_name = StudyName(study_name)
+    subjects_pattern = get_subject_id_pattern(study_name)
```
NicolasGensollen, line 105: Let's get rid of this (see another comment to see why):

```suggestion
```

clinica/converters/_utils.py
```diff
@@ -175,33 +183,67 @@ def create_participants_df(
...
         ]
+
     participant_df.reset_index(inplace=True, drop=True)
 
+    # Extracting subjects list from the input subjects list file
+    if subjects and subjects_pattern:
+        import re
+
+        rgx = re.compile(subjects_pattern)
+        subjects_list = filter(rgx.fullmatch, get_subjects_list_from_file(subjects))
```
NicolasGensollen, line 200: Here we handle the `subjects` parameter of the function:

```python
if subjects:
    subjects_list = [
        bids_id_factory(study_name).from_original_study_id(subject)
        for subject in get_subjects_list_from_file(subjects)
    ]
```

Then, we combine `bids_ids` and `subjects_list` into a single filtering set:

```python
bids_id_to_keep: set[BIDSSubjectID] = set(bids_ids + subjects_list)
```

clinica/converters/_utils.py
```diff
@@ -175,33 +183,67 @@ def create_participants_df(
...
+
     # Adding participant_id column with BIDS ids
     for i in range(0, len(participant_df)):
         value = bids_id_factory(study_name).from_original_study_id(
             participant_df["alternative_id_1"][i]
         )
         bids_id = [s for s in bids_ids if value in s]
+
```
NicolasGensollen, line 208: Here, we filter each row according to our filtering set:

```python
bids_id_from_participant_df = bids_id_factory(study_name).from_original_study_id(
    participant_df["alternative_id_1"][i]
)
if bids_id_from_participant_df in bids_id_to_keep:
    participant_df.at[i, "participant_id"] = bids_id_from_participant_df
else:
    index_to_drop.append(i)
    subjects_to_drop.append(bids_id_from_participant_df)
```

clinica/converters/_utils.py
```diff
@@ -588,3 +630,48 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
         List of subjects.
     """
     return subjects_list_path.read_text().splitlines()
+
+
+def get_subject_id_pattern(study_name: StudyName) -> str:
```
NicolasGensollen, line 635: We can delete this function and rely on `BIDSSubjectID` (see `clinica.converters.study_model`) which already has the logic to convert IDs from their raw values to BIDS values and vice versa. 
Sorry for the bad suggestion, I hadn't thought about it hard enough...

NicolasGensollen (commented):
Thanks @Adam-Ismaili-92 !

After taking a closer look, I think we need to simplify things a bit. The code seems way too complicated for what it's doing and I fear it's just going to get worse...

I made a suggestion on how to simplify the management of `bids_ids` in this function. I haven't tested anything though, so there might be some adjustments to make.

Also pinging @AliceJoubert since she knows this part of the codebase pretty well and might be able to correct me if I overlooked some sketchy details...

clinica/converters/_utils.py
AliceJoubert: Considering `bids_ids` is a list of the subjects present in the BIDS dataset I would say we can replace it upstream with the list resulting of the new function `get_subjects_list` that was added in oasis `_utils` in #1474
https://github.com/aramis-lab/clinica/blob/0a0b9f02f12473eda125fa68b1bcd2f49e5438bb/clinica/converters/oasis_to_bids/_converter.py#L59-L73

Wa can make the change in this PR for the participants only and extend to scans and sessions in a next PR. That would mean something like this in the function above : 

```python
cprint("Converting clinical data...", lvl="info")  
self._create_participants_tsv(clinical_data_dir, bids_dir, bids_ids = get_subjects_list(bids_dir, subjects)) 
bids_ids = get_subjects_from_bids_dataset(bids_dir) 
self._create_sessions_tsv(clinical_data_dir, bids_dir, bids_ids) 
self._create_scans_tsv(bids_dir) 
```

Adam-Ismaili-92: Thanks for your suggestions @AliceJoubert and @NicolasGensollen.

I also agree with the first point of Alice, and the fact that it should be handled in a dedicated PR.

I also agree to use a single subject filtering collection and a SubjectID classes for efficiency matters, but I have a few questions with the code you proposed.

get_subjects_list is adapted to process folder names from the original dataset, not the BIDS one.
When filtering, the function will check if folder names match the form : OAS1-\d{4}-MR1
But, if I'm not mistaken, the subject folders names inside the BIDS dataset are of the form : sub-OASIS1\d{4}
Thus a mismatch will arise and bids_ids will be empty.

So how do you think I should handle the problem? Should I adapt get_subjects_list? Or create a new specific getter?

NicolasGensollen: First of all, there are several functions with very similar names and which are doing very similar things, which could be confusing. For example:

https://github.com/aramis-lab/clinica/blob/4ae78a490cd07dd76b01de2f554f16d8f821051f/clinica/converters/_utils.py#L577-L590

Takes the path to a file containing a list of subject IDs and return the list of IDs, but as plain strings.

And these ones, converter-specific, but with slightly different behaviors:

https://github.com/aramis-lab/clinica/blob/4ae78a490cd07dd76b01de2f554f16d8f821051f/clinica/converters/adni_to_bids/_utils.py#L78-L85

https://github.com/aramis-lab/clinica/blob/4ae78a490cd07dd76b01de2f554f16d8f821051f/clinica/converters/oasis_to_bids/_utils.py#L48-L65

It could be the time to make all of this more uniform and maybe improve some names to make it clear what these functions are doing (not in this PR of course !).

Second, I think that we should handle the IDs using the dedicated class we added recently. For example:

https://github.com/aramis-lab/clinica/blob/4ae78a490cd07dd76b01de2f554f16d8f821051f/clinica/converters/study_models.py#L204-L214

Because, these classes handle both formats, the one from the study and the BIDS ID.

So maybe we could have something like this:

```python
def get_subjects_list_from_file(subjects_list_path: Path) -> list[BIDSSubjectID]:
    ...
``` 

Then, with this list, you can switch from BIDS to original IDs if you need:

```python
bids_ids = get_subjects_list_from_file(some_file)
raw_ids = [x.to_original_study_id() for x in bids_ids]
```

WDYT ?

AliceJoubert: My bad, I did not mention it. You can use the implemented modelisation of bids ids for OASIS : 

https://github.com/aramis-lab/clinica/blob/4ae78a490cd07dd76b01de2f554f16d8f821051f/clinica/converters/study_models.py#L204-L226

where the method `from_original_study_id` transforms the ids from the raw data to the bids ids

Adam-Ismaili-92: In the most recent commit, I tried to handle IDs by using the dedicated class. 

I also used both functions **_get_subjects_list_from_file** and **get_subjects_list** to get the list of **bids_ids** either from the data or from the subjects list file provided by the user, in the correct type (the dedicated class).

However, I could not modify the generic function **get_subjects_list_from_file** directly in this PR, it created significant regression errors that would be difficult to handle in a single PR.

I second your proposition to tackle the multiple occurrences of **_get_subjects_list_from_file/data**, **get_subjects_list** in an another PR.

WDYT @AliceJoubert @NicolasGensollen ?

AliceJoubert:
I took a closer look at the PR from the beginning. Let's go back to what we want from this PR for a minute. The main goal was to inform the user about potential data loss, which basically boils down to comparing what is inside the BIDS folder (result) and what subjects the user has asked for. 

- This does not need to happen inside the `create_participants_tsv` function and I would rather suggest that we have a specific function that does this. All the function needs to do is make the difference between requested subjects and subjects inside the BIDS, then warn the user accordingly. That means we do not need to create specific cases inside `create_participants_tsv` and write it specifically for OASIS.  

- As for the function `get_subjects_list` I think it would be best to restrict its usage to raw folders to avoid confusion. Especially as we started a modelisation of BIDS and have dedicated functions already performing this : 

https://github.com/aramis-lab/clinica/blob/9a987a733ba9acabc737e2db9e0aa68246f87a3f/clinica/dataset/bids/_queries.py#L11-L31

- @Adam-Ismaili-92 something wrong might have happened when rebasing as I see the function `mapping_diagnosis` from your previous PR being added here in your new commits. 


_This PR shed light on many potential issues we will resolve in follow up PRs :_

_--> Extend to ADNI, IXI... any that has the subject option_
_--> Moving the participants.tsv function specific to OASIS in oasis utils and test in the appropriate file, then cut out parts that are not necessary in OASIS_
_--> Same for ADNI_
_--> Unify the `get_subjects_list` functions (see #1478) and be more explicit with the names_
_--> Make sure we are using existing models such as `BIDSSubjectID`_

clinica/converters/_utils.py
```diff
@@ -588,3 +582,24 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
...
+
+    bids_ids_2 : list[BIDSSubjectID]
+        Second bids ids list.
+
+    Returns
+    -------
+    list[BIDSSubjectID] :
+        Bids ids difference.
+    """
+    return list(set(bids_ids_1).symmetric_difference(set(bids_ids_2)))
```
AliceJoubert, line 605:
```suggestion
    return list(set(expected_subjects) - set(obtained_subjects))
```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,53 @@ def convert(
...
+            Default=None.
+        """
+        from clinica.converters._utils import compare_bids_ids
+        from clinica.converters.study_models import OASISBIDSSubjectID
+        from clinica.dataset.bids._queries import get_subjects_from_bids_dataset
+        from clinica.utils.stream import cprint
+
+        from ._utils import get_subjects_list
+
+        raw_ids: list[OASISBIDSSubjectID] = [
```
AliceJoubert, line 84:
```suggestion
        raw_ids = [
```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,53 @@ def convert(
...
+            OASISBIDSSubjectID(OASISBIDSSubjectID.from_original_study_id(subject.name))
+            for subject in get_subjects_list(source_dir, subjs_list_path=subjects)
+        ]
+        bids_ids: list[OASISBIDSSubjectID] = get_subjects_from_bids_dataset(bids_dir)
+
```
AliceJoubert, line 89:
```suggestion
        bids_ids = [OASISBIDSSubjectID(subject) for subject get_subjects_from_bids_dataset(bids_dir)]

```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,53 @@ def convert(
...
+        ids_difference = compare_bids_ids(raw_ids, bids_ids)
+
+        if len(ids_difference) > 0:
+            cprint(
+                msg=(
+                    "The following subjects were not found in your BIDS folder :\n"
```
AliceJoubert, line 95:
```suggestion
                    "The following subjects were not converted to your BIDS folder :\n"
```

AliceJoubert (commented):
Thanks for taking into account what we discussed @Adam-Ismaili-92. Here are a few more suggestions but it should soon be good to merge

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -81,20 +129,19 @@ def _create_participants_tsv(
         from clinica.converters._utils import create_participants_df
         from clinica.converters.study_models import StudyName
 
+        from ._utils import mapping_diagnosis
```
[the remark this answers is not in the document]
Adam-Ismaili-92: I didn't find why, and I'm a little afraid to break something if I try to do git manipulations to correct this. But it doesn't seem to prevent the merge for the PR or to cause conflicts.

clinica/converters/_utils.py
Adam-Ismaili-92: Those are not the same types of differences, the first one (using symmetric_difference) is keeping the different objects from both sets, while the second (using -) is only keeping the different objects inside expected_subjects. For my tests and the correct functioning of the function, I need to seep the objects from both sets.

clinica/converters/_utils.py
```diff
@@ -588,3 +582,24 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
         List of subjects.
     """
     return subjects_list_path.read_text().splitlines()
+
+
+def compare_bids_ids_lists(
```
AliceJoubert, line 587: Since the function is doing something specific (_ie_ an asymmetrical difference) I think it would be nice to be more explicit in the name of the function and the docstring. 

For example, the name can be something like `comparing_expected_vs_obtained_bids_ids` (feel free to adapt if you have something better).

clinica/converters/_utils.py
```diff
@@ -588,3 +582,24 @@ def get_subjects_list_from_file(subjects_list_path: Path) -> list[str]:
...
+    expected_subjects: list[BIDSSubjectID], obtained_subjects: list[BIDSSubjectID]
+) -> list[BIDSSubjectID]:
+    """Gets the difference between bids ids lists.
```
AliceJoubert, line 590:
```suggestion
    """Gets the BIDS ids that were expected but not found in the obtained subjects list.
```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,54 @@ def convert(
...
+            Default=None.
+        """
+        from clinica.converters._utils import compare_bids_ids_lists
+        from clinica.converters.study_models import OASISBIDSSubjectID
+        from clinica.dataset.bids._queries import get_subjects_from_bids_dataset
+        from clinica.utils.stream import cprint
+
+        from ._utils import get_subjects_list
+
+        raw_ids = [
```
AliceJoubert, line 84: In other parts of the codebase for converters you can see either `source_id` and `bids_id` to indicate ids under the format of the study and the same ids under the BIDS format, respectively. Since it is not exactly what we have here and to avoid confusion, I suggest we stick with the denomination used inside the function  : 
```suggestion
        expected_ids = [
```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,54 @@ def convert(
...
+            OASISBIDSSubjectID(OASISBIDSSubjectID.from_original_study_id(subject.name))
+            for subject in get_subjects_list(source_dir, subjs_list_path=subjects)
+        ]
+        bids_ids = [
```
AliceJoubert, line 88:
```suggestion
        obtained_ids = [
```

clinica/converters/oasis_to_bids/_converter.py
```diff
@@ -55,6 +55,54 @@ def convert(
...
+            OASISBIDSSubjectID(subject)
+            for subject in get_subjects_from_bids_dataset(bids_dir)
+        ]
+
+        if ids_difference := compare_bids_ids_lists(raw_ids, bids_ids):
```
AliceJoubert, line 93:
```suggestion
        if ids_difference := compare_bids_ids_lists(expected_ids, obtained_ids):
```

AliceJoubert (commented):
Last small remarks I have on this PR and it should be good to go @Adam-Ismaili-92

AliceJoubert:
Both failing tests are not due to this PR, so I am merging. Thanks for the work @Adam-Ismaili-92 !

Issue #1481, opened 2025-03-26
Difficulties running `t1-volume` with SPM standalone set-up


AliceJoubert:
I tried to run `t1-volume` locally on my MacOS (M3 Sonoma) with SPM standalone v25.01, Matlab runtime vR2024b and clinica in dev version (0.9.3+).  I encountered several difficulties : 

### Required paths
Even though I planned on using only the standalone, clinica requested for `SPM_HOME` in addition to `MCR_HOME` and `SPMSTANDALONE_HOME`.

### Checking for dependencies
When building the pipeline `T1VolumeTissueSegmentation`,  dependencies are checked with the function `check_software` : https://github.com/aramis-lab/clinica/blob/64dd0c670ca894222ee6d7e92d4ca68c385eb618/clinica/utils/check_dependency.py#L704-L709
Basically, it performs : 
(1) checking for environment variables presence
(2) checking for software version through nypipe interface command line version property.

In our pipeline here the default for `software` is only `spm` whatever the set-up : with MATLAB + SPM12 or with MCR + SPM standalone.
In step (1) function `check_spm` verifies first if the pipeline can be run with the standalone : 

https://github.com/aramis-lab/clinica/blob/64dd0c670ca894222ee6d7e92d4ca68c385eb618/clinica/utils/check_dependency.py#L379-L383

However `software` is not updated so in step (2) the version check is performed for `spm` whatever the result of step (1). So, it runs even though `spm12` is not installed with this set-up : 

```bash
...
  File "/Users/alice.joubert/clinica/clinica/utils/check_dependency.py", line 518, in get_software_version
    return _get_spm_version()
           ^^^^^^^^^^^^^^^^^^
  File "/Users/alice.joubert/clinica/clinica/utils/check_dependency.py", line 540, in _get_spm_version
    return Version(spm.SPMCommand().version)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alice.joubert/miniforge3/envs/clinica/lib/python3.12/site-packages/packaging/version.py", line 200, in __init__
    match = self._regex.search(version)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```

I would expect the check to be made with :  https://github.com/aramis-lab/clinica/blob/64dd0c670ca894222ee6d7e92d4ca68c385eb618/clinica/utils/check_dependency.py#L543-L550

[...]

PR #1488, opened 2025-04-01
[FIX] Fix SPM standalone version check

NicolasGensollen:
Addresses part of #1481 

WIP...


AliceJoubert (approved):
Thanks for the work @NicolasGensollen, now it works as I would have expected

Issue #1485, opened 2025-03-31
Improve tests for client instantiation


AliceJoubert:
Recently several bugs linked to the client were reported (see #1480, #1484). These could have been avoided if the client was extensively tested. There is a [module](https://github.com/aramis-lab/clinica/blob/4fe58feabe9e3d7978348f3b090ae0c71fb89486/test/instantiation/test_cli.py) in instantiation tests but it could be completed. For example, options are not tested at all.

PR #1489, opened 2025-04-02
[FIX] Fix broken imports in CLI

### LINKED ISSUE #1484: Broken imports in iotools client
Following BIDS and CAPS architecture rework in #1453 there are broken imports in 2 iotools : https://github.com/aramis-lab/clinica/blob/4fe58feabe9e3d7978348f3b090ae0c71fb89486/clinica/iotools/cli.py#L92 and https://github.com/aramis-lab/clinica/blob/4fe58feabe9e3d7978348f3b090ae0c71fb89486/clinica/iotools/cli.py#L110

NicolasGensollen:
Closes #1484 

This PR also proposes to add some unit tests related to the CLI client in order to avoid these kind of bugs (see issue #1485). The proposed solution consists of mocking the internal function and calling the CLI in the test with `CliRunner`. The tests then check the exit code as well as whether the mock was called and the arguments passed to it. Assuming the internal function is tested separately as well, this should be enough I believe. This testing strategy can be used for other CLI in order to close fully #1485.

### DIFF
--- a/clinica/iotools/cli.py
+++ b/clinica/iotools/cli.py
@@ -95,11 +95,8 @@ def check_missing_modalities(
     output_prefix: str = "missing_mods",
 ) -> None:
     """Check missing modalities in a BIDS dataset."""
-    from clinica.utils.inputs import check_bids_folder
-
     from .data_handling import compute_missing_mods
 
-    check_bids_folder(bids_directory)
     compute_missing_mods(bids_directory, output_directory, output_prefix)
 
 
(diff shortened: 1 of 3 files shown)


AliceJoubert (approved):
LGTM, thanks @NicolasGensollen ! For the client tests, do you think we should test all options combinations or is it not really necessary ? In these cases for example this would be with and without the option `--output_file` for each function

NicolasGensollen:
> LGTM, thanks @NicolasGensollen ! For the client tests, do you think we should test all options combinations or is it not really necessary ? In these cases for example this would be with and without the option `--output_file` for each function

Thanks for the review @AliceJoubert ! I think we should try to cover as many options as possible. Since everything is mocked, these tests are very fast, so we can test numerous combinations. I added some tests for the option `--output_prefix` for `check_missing_modalities`. I think the rest is only arguments, already covered by the added tests, right ?

PR #1490, opened 2025-04-02
[REF] Search for SPM standalone execution file with pattern instead of hardcoding for spm12

AliceJoubert:
Adresses part of #1481 regarding `_get_platform_dependant_matlab_command_for_spm_standalone`.

Rq : not sure it will do well with CI set-up

### DIFF
--- a/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
+++ b/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
@@ -135,7 +135,7 @@ class T1VolumeTissueSegmentation(Pipeline):
             zip_list_files,
         )
 
-        use_spm_standalone_if_available()
+        # use_spm_standalone_if_available() #todo : why needed at this point ?
 
         # Get <subject_id> (e.g. sub-CLNC01_ses-M000) from input_node
         # and print begin message
--- a/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
+++ b/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
@@ -135,7 +135,7 @@ class T1VolumeTissueSegmentation(Pipeline):
             zip_list_files,
         )
 
-        # use_spm_standalone_if_available() #todo : why needed at this point ?
+        use_spm_standalone_if_available()  # todo : why needed at this point ?
 
         # Get <subject_id> (e.g. sub-CLNC01_ses-M000) from input_node
         # and print begin message
--- a/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
+++ b/clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
@@ -135,8 +135,6 @@ class T1VolumeTissueSegmentation(Pipeline):
             zip_list_files,
         )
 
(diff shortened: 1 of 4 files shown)


clinica/utils/spm.py
```diff
@@ -114,16 +114,27 @@ def configure_nipype_interface_to_work_with_spm() -> None:
     mlab.MatlabCommand.set_default_paths(f"{get_spm_home()}")
 
 
+def _get_real_spm_standalone_file(spm_standalone_home: Path) -> str:
+    try:
+        return next(spm_standalone_home.rglob("run_spm*.sh")).name
```
NicolasGensollen, line 119: I'm wondering whether it could happen that multiple "run_spmXX.sh" files are present ? I doubt this could be the case but it might be more robust to handle the case, WDYT ? We could handle it by raising an error with the list of files found.
Also, I think it could be interesting to log the file used (INFO or DEBUG) (probably done in the caller rather than here).

clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
```diff
@@ -135,7 +135,7 @@ def _build_core_nodes(self):
             zip_list_files,
         )
 
-        use_spm_standalone_if_available()
+        use_spm_standalone_if_available()  # todo : why needed at this point ?
```
NicolasGensollen, line 138: Good question, I think this is a leftover from legacy code. With the new improved way of checking third party softwares, this is indeed not needed since the Nipype configuration is done when trying to get the version. I think we can remove it and test whether the pipeline runs with both SPM + Matlab and SPM + MCR. WDYT ?

NicolasGensollen (commented):
Thanks @AliceJoubert !

The PR looks good to me. I just have a small comment on the function `_get_real_spm_standalone_file`.

clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
AliceJoubert: Okay I will test that separately

clinica/utils/spm.py
AliceJoubert: I don't think that would happen but I can handle the case anyway !

clinica/utils/spm.py
```diff
@@ -114,16 +114,32 @@ def configure_nipype_interface_to_work_with_spm() -> None:
...
 def _get_platform_dependant_matlab_command_for_spm_standalone(
     spm_standalone_home: Path, mcr_home: Path
 ) -> str:
     import platform
 
+    from clinica.utils.stream import cprint
+
+    spm_file = _get_real_spm_standalone_file(spm_standalone_home)
+    cprint(
+        f"Using the following spm file to build the matlab dependent command : {spm_file}",
```
NicolasGensollen, line 135:
```suggestion
        f"Using the following spm file to build the matlab dependent command : {spm_standalone_home / spm_file}",
```

clinica/pipelines/t1_volume_tissue_segmentation/t1_volume_tissue_segmentation_pipeline.py
NicolasGensollen: If you have tested it and it works as expected, I believe this is ready to be merged.

AliceJoubert:
I could test with both Matlab + SPM12 and SPM12 standalone set-up and it works so I removed the duplicated line looking for the environment variables.

NicolasGensollen:
Thanks @AliceJoubert !

Merging then !

PR #1492, opened 2025-04-02
[REF] OASIS-to-BIDS : moving parts of the code that performs data modification from _converter to _utils

### LINKED ISSUE #1458: [REF] OASIS-to-BIDS : move parts of the code that performs data modification from `_converter` to `_utils`
In OASIS-to-BIDS converter, some lines of codes are still performing modifications on the  dataframes inside the `_converter.py` module. 

https://github.com/aramis-lab/clinica/blob/8180aef7a08c561fe7ad39a9dd589013008d42b6/clinica/converters/oasis_to_bids/_converter.py#L91-L98

It would be better to have these made in the `_utils.py` in an appropriate function so it can be tested.

Adam-Ismaili-92:
Closes #1458

### DIFF
--- a/clinica/converters/oasis_to_bids/_utils.py
+++ b/clinica/converters/oasis_to_bids/_utils.py
@@ -13,6 +13,7 @@ __all__ = [
     "write_sessions_tsv",
     "write_scans_tsv",
     "get_first_image",
+    "initialize_participants_df",
     "get_image_with_good_orientation",
 ]
 
@@ -224,6 +225,29 @@ def get_first_image(input_folder: Path) -> Path:
             f"No file ending in .img found in {input_folder}.",
             FileNotFoundError,
         )
+
+
+def initialize_participants_df(participants_df: pd.DataFrame) -> None:
+    """Initialize participants metadata in the corresponding dataframe.
+
+    Parameters
+    ----------
+    participants_df : pd.DataFrame
+        Dataframe containing participants metadata.
+
+    Returns
+    -------
+    pd.DataFrame :
+        participants_df.
+    """
+
+    # Replace the values of the diagnosis_bl column
+    participants_df["diagnosis_bl"].replace([0.0, np.nan], "CN", inplace=True)
+    participants_df["diagnosis_bl"].replace([0.5, 1.0, 1.5, 2.0], "AD", inplace=True)
+
+    participants_df = participants_df.fillna("n/a")
+
+    return participants_df
 
 
 def get_image_with_good_orientation(image_path: Path) -> nb.Nifti1Image:
--- a/clinica/converters/oasis_to_bids/_utils.py
+++ b/clinica/converters/oasis_to_bids/_utils.py
@@ -13,7 +13,7 @@ __all__ = [
     "write_sessions_tsv",
     "write_scans_tsv",
     "get_first_image",
-    "initialize_participants_df",
+    "mapping_diagnosis",
     "get_image_with_good_orientation",
 ]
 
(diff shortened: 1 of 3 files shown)


clinica/converters/oasis_to_bids/_utils.py
```diff
@@ -226,6 +227,29 @@ def get_first_image(input_folder: Path) -> Path:
         )
 
 
+def initialize_participants_df(participants_df: pd.DataFrame) -> None:
```
AliceJoubert, line 230:
```suggestion
def initialize_participants_df(participants_df: pd.DataFrame) -> pd.DataFrame:
```

clinica/converters/oasis_to_bids/_utils.py
```diff
@@ -226,6 +227,29 @@ def get_first_image(input_folder: Path) -> Path:
...
+        Dataframe containing participants metadata.
+
+    Returns
+    -------
+    pd.DataFrame :
+        participants_df.
+    """
+
+    # Replace the values of the diagnosis_bl column
+    participants_df["diagnosis_bl"].replace([0.0, np.nan], "CN", inplace=True)
```
AliceJoubert, line 245: Now reading this it feels wrong to put `"CN"` (which means Cognitively Normal) or any other diagnosis when we actually do not know anything about it. I personally would prefer to keep a nan. Same for any other value that is not hard-coded here

AliceJoubert, line 245: Maybe we could use another function to make the correspondance between values : 

```python
def _mapping_diagnosis(diagnosis_bl:float)->str:
     if diagnosis_bl == 0 : 
          return "CN"
     if diagnosis_bl in (0.5, 1.0, 1.5, 2.0):
          return "AD"
     return "n/a"
```

That is very easy to test and can be used as follows : 

```python
    participants_df["diagnosis_bl"] = participants_df["diagnosis_bl"].apply(lambda x:_mapping_diagnosis(x))
```

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 ! I have a few comments about the code :

clinica/converters/oasis_to_bids/_utils.py
Adam-Ismaili-92: Thanks for your suggestion. I am a bit lost however. What are you referring to when you say "Same for any other value that is not hard-coded here"? Do you mean that I should remove the replacements by diagnosis and keep the original value for any value that is not a float? Or should I remove all the replacements by diagnosis?

Adam-Ismaili-92: Thanks for your suggestions. I have one question however. Will it not replace the np.nan by "n/a"? Is it the expected behavior?

clinica/converters/oasis_to_bids/_utils.py
Adam-Ismaili-92: Thanks for your suggestion. It was an oversight on my part.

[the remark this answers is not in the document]
Adam-Ismaili-92: Thanks for your suggestion. The function is more understandable and the name is indeed matching better.

clinica/converters/oasis_to_bids/_utils.py
AliceJoubert: Yes, it is expected as we follow BIDS specifications for converters. Unknowns values are to be written "n/a". You can see it specified on [this page](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#participants-file). Hence why there is often `.fillna("n/a")` in converters

test/unittests/converters/test_oasis_to_bids_utils.py
```diff
@@ -365,6 +365,18 @@ def test_get_subjects_list_from_file(tmp_path) -> None:
...
+def test_mapping_diagnosis(tmp_path) -> None:
+    from clinica.converters.oasis_to_bids._utils import mapping_diagnosis
+
+    values = [0.0, 0.5, 1.0, np.nan, 2.0, 1.5, np.nan]
+
+    mapped_values = [mapping_diagnosis(val) for val in values]
+
+    expected_mapped_values = ["CN", "AD", "AD", "n/a", "AD", "AD", "n/a"]
+
+    assert mapped_values == expected_mapped_values
```
AliceJoubert, line 377: You can use `parametrize` when you need to test several cases following the same testing scenario :

```suggestion
@pytest.mark.parametrize("input, expected", [(0, "CN"), (0.5, "AD"), (1, "AD"), (1.5, "AD"), (2, "AD"), (np.nan, "n/a"), (4, "n/a")])
def test_mapping_diagnosis(input, expected) -> None:
    from clinica.converters.oasis_to_bids._utils import mapping_diagnosis
    assert expected == mapping_diagnosis(input)
```

Also, you don't need to pass `tmp_path` to the function if you won't use it in the test

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 ! Just one last comment about your unit test

test/unittests/converters/test_oasis_to_bids_utils.py
Adam-Ismaili-92: Thanks for your suggestion, I will use parametrize instead.

AliceJoubert:
LGTM @Adam-Ismaili-92 , will merge

PR #1497, opened 2025-04-07
[FIX] Fix imports in `pet-surface`

AliceJoubert:
After creation of `pipelines.pet` module some functions were moved to `pet.utils` from `utils.pet` but imports were not changed for `pet-surface`. This pipeline is also expected to be in this module (cf #1237) but in the meantime these imports should be fixed.


NicolasGensollen (approved):
LGTM, thanks for fixing these @AliceJoubert !

PR #1499, opened 2025-04-07
[FIX] Fix imports in  `pet-volume` after creation of pipelines.pet

davidaquilue:
Following PR #1497 there were some left-over imports that imported get_suvr_mask from `utils.pet`.

Adding the PR as these imports should be fixed while the pipeline is included in the new module (cf #1237).

### DIFF
--- a/clinica/pydra/pet_volume/pipeline.py
+++ b/clinica/pydra/pet_volume/pipeline.py
@@ -72,9 +72,9 @@ def build_core_workflow(name: str = "core", parameters: dict = {}) -> Workflow:
     """
     from pydra.tasks import petpvc
 
+    from clinica.pipelines.pet.utils import get_suvr_mask
     from clinica.pydra.shared_workflows.smoothing import build_smoothing_workflow
     from clinica.pydra.utils import sanitize_fwhm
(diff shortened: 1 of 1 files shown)


clinica/pydra/pet_volume/pipeline.py
```diff
@@ -74,7 +74,7 @@ def build_core_workflow(name: str = "core", parameters: dict = {}) -> Workflow:
 
     from clinica.pydra.shared_workflows.smoothing import build_smoothing_workflow
     from clinica.pydra.utils import sanitize_fwhm
-    from clinica.utils.pet import get_suvr_mask
+    from clinica.pipelines.pet.utils import get_suvr_mask
```
NicolasGensollen, line 77: Could you move this line just before the import of `build_smoothing_workflow` ? This will make the code formatter happy and turn the CI green.

NicolasGensollen (commented):
Thanks for fixing this up @davidaquilue !

Just a small thing to pass the CI, but LGTM otherwise.

NicolasGensollen (approved):
Thanks @davidaquilue !

The CI failure is unrelated to this PR, so I'm gonna merge.

PR #1500, opened 2025-04-08
[REF] Proposition for AIBL Json creation

AliceJoubert:
This PR proposes a solution to write the json file associated to a PET image for the AIBL dataset using the header of the dicom image. 

Discussion (for later):
-  can this be used for other than AIBL ?
- would using a class be beneficial ? I am starting to get a lot of : if modality do something


clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+            ["ReconMethodParameterValues", ("ReconstructionMethod",), [""]],
+            ["ReconFilterType", ("ConvolutionKernel",), "n/a"],
+            ["ReconFilterSize", ("ConvolutionKernel",), np.nan],  # todo : corresp ?
+            ["AttenuationCorrection", ("AttenuationCorrectionMethod",), "n/a"],
+        ],
+    ).set_index(keys="BIDSname", drop=False)
+
+
+def _get_dcm_value_from_header(
+    keys_list: Tuple[str], dicom_header: Union[FileDataset, DataElement]
```
NicolasGensollen, line 121:
```suggestion
    keys_list: tuple[str], dicom_header: Union[FileDataset, DataElement]
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+        return None
+
+    result = None
+    if len(keys_list) == 1:
+        result = dicom_header.get(Tag(keys_list[0]))
+    elif header := dicom_header.get(Tag(keys_list[0])):
+        # Handles nested tags
+        _get_dcm_value_from_header(keys_list[1:], header[0])
+    if result:
+        return result.value
```
NicolasGensollen, line 151: Wouldn't something a bit simpler like this work ?

```python
if not keys_list:
    return None
header = dicom_header.get(Tag(keys_list[0])):
if len(keys_list) == 1:
    return header
return _get_dcm_value_from_header(keys_list[1:], header[0])
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+
+    # todo
+    # if result:
+    #     if type(result.value) == MultiValue:
+    #         # Handles case where result is MultiValue, which is not JSON serializable
+    #         return list(result.value)
```
NicolasGensollen, line 157: Maybe implement this in another function wrapping `_get_dcm_value_from_header` and handling the result value.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+    else:
+        # todo : prints in debug ?
+        for index, data in metadata.iterrows():
+            if dcm_value := _get_dcm_value_from_header(data.DCMtag, dicom_header):
+                metadata.loc[index, "Value"] = dcm_value
+
+
+def _set_scan_start(dcm_result: pd.DataFrame) -> None:
+    dcm_result.loc["ScanStart", "Value"] = "0"
```
NicolasGensollen, line 181: I'm not sure to understand why you don't put 0 directly as the default value in the data frame ?

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+    # if result:
+    #     if type(result.value) == MultiValue:
+    #         # Handles case where result is MultiValue, which is not JSON serializable
+    #         return list(result.value)
+    #     return result.value
+
+    return None
+
+
+def _update_from_image_dicoms(metadata: pd.DataFrame, dcm_dir: Path) -> None:
```
NicolasGensollen, line 163: Or something similar, just to clarify a bit more that it is the metadata which is updated.

```suggestion
def _update_metadata_from_image_dicoms(metadata: pd.DataFrame, dcm_dir: Path) -> None:
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+import pandas as pd
+from pydicom.dataset import DataElement, FileDataset
+from pydicom.multival import MultiValue
+from pydicom.tag import Tag
+
+from clinica.utils.stream import cprint
+
+# todo : __all__
+
+# todo : can this be used for other than AIBL ?
```
NicolasGensollen, line 19: Indeed, it's also what I'm wondering when reading all this.
We can keep it here and see later whether this could be useful for other converters, and move it if it turns out to be.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def _set_injection_start(dcm_result: pd.DataFrame) -> None:
+    if (injection := dcm_result.loc["InjectionStart", "Value"]) and (
+        acquisition := dcm_result.loc["TimeZero", "Value"]
+    ):
+        dcm_result.loc["InjectionStart", "Value"] = injection - float(acquisition)
+
+
+def _get_admin_mode_from_start_time(dcm_result: pd.DataFrame) -> None:
```
NicolasGensollen, line 191: If you have, for all these small functions, some context or knowledge about why you update things this way, it could be nice to write it in a small docstring.

NicolasGensollen (commented):
Thanks @AliceJoubert !

The PR looks good so far. At first, I wasn't a huge fan of the giant data frame mutated by all these functions, but I don't have a better idea at the moment. So let's stick with that !

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: Indeed ! Thanks

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: I think I did this because the default values in the beginning in the dataframe are the ones indicated by BIDS, and the 0 added here is a choice we make (that I would have to explain)

NicolasGensollen: Alright, makes sense then !

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+    _set_time_from_ms_to_seconds(metadata, "FrameDuration")
+    _get_admin_injection_time_to_zero(metadata)
+
+    _set_decay_time(metadata)
+    _check_decay_correction(metadata)
+
+    _update_injected_mass(metadata)
+
+
+def get_json_data(dcm_dir: Path, modality: str) -> pd.DataFrame:
```
NicolasGensollen: Why not use `Modality` instead of `str` ? 
If user-facing (which I doubt it is), you could use `str | Modality`, otherwise, the caller will already have handled the str to Modality cast.

```suggestion
def get_json_data(dcm_dir: Path, modality: Modality) -> pd.DataFrame:
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+    modality = Modality(modality)
```
NicolasGensollen:
```suggestion
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+
+def _update_injected_mass(dcm_result: pd.DataFrame) -> None:
+    injected = dcm_result.loc["InjectedRadioactivity", "Value"]
+    specific = dcm_result.loc["SpecificRadioactivity", "Value"]
+    if injected and specific != "n/a":
+        dcm_result.loc["InjectedMass", "Value"] = injected / float(specific)
+        dcm_result.loc["InjectedMassUnits", "Value"] = "mole"
+
+
+def _get_default_for_modality(modality: Modality) -> Optional[pd.DataFrame]:
```
NicolasGensollen: Following your naming conventions, I would have called this function `_get_dicom_tags_and_defaults`. WDYT ?

NicolasGensollen: I don't understand why the return type is `Optional[pd.DataFrame]`, it should be `pd.DataFrame`, right ?

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+        cprint(
+            msg=f"No DICOM found at {dcm_dir}, the image json will be filled with default values",
+            lvl="warning",
+        )
+    else:
+        for index, data in metadata.iterrows():
+            if dcm_value := _get_dcm_value_from_header(data.DCMtag, dicom_header):
+                cprint(msg=f"Value for {data.DCMtag} is : {dcm_value}", lvl="debug")
+                metadata.loc[index, "Value"] = dcm_value
+            elif getLevelName(getLogger("clinica").level) == "DEBUG":
```
NicolasGensollen: I don't understand why you're checking that. You just need to log a debug message, you don't care what's the logger's logging level, you just log it, and the logger will handle it according to its logging level. Does that make sense ?

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+    header = dicom_header.get(keys_list[0])
+    if len(keys_list) == 1:
+        return header
+    if header:
+        return _get_dcm_value_from_header(keys_list[1:], header[0])
+    return None
+
+
+def _get_dcm_value_from_header(
+    keys_list: tuple[str], dicom_header: Union[FileDataset, DataElement]
```
NicolasGensollen: Maybe the variable name `keys_list` could be improved. Since you call it with `data.DCMtag`, I'd expect something like `dcm_tags`. 

Also, it seems like this value is a length 1 or more tuple of strings and can be None, so it should be `dcm_tags: Optional[tuple[str, ...]] = None`.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+                "InstitutionalDepartmentName",
+                ("InstitutionalDepartmentName",),
+                "n/a",
+            ],
+        ],
+    ).set_index(keys="BIDSname", drop=False)
+
+
+def _check_dcm_value(
+    dicom_data: Union[None, str, list, float],
```
NicolasGensollen: This is getting as input the output from `_fetch_dcm_data_from_header`, so it seems like the types are not totally aligned, right ?

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def _format_time(timestamp: str) -> str:
+    """Format the timestamps that comes under the format HHMMSS or HHMMSS.0000 to HH:MM:SS."""
+    from time import strftime, strptime
+
+    return strftime("%H:%M:%S", strptime(timestamp.split(".")[0], "%H%M%S"))
+
+
+def _substract_formatted_times(reference: str, action: str) -> float:
```
NicolasGensollen: The variable name "action" tricked me into thinking this was a parameter passed to `strptime`. Maybe use the term `event` instead. Either:

- `_substract_formatted_times(reference: str, event: str) -> float:`
- `_substract_formatted_times(reference_time: str, event_time: str) -> float:`

I have a slight preference for the second which seems clearer even though a bit redundant.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+
+def _set_time_from_ms_to_seconds(dcm_result: pd.DataFrame, field: str) -> None:
+    """Set time to seconds for FrameTimeStart and FrameDuration which are encoded as list(float), given in ms."""
+    if start := dcm_result.loc[field, "Value"]:
+        dcm_result.loc[field, "Value"] = start / 1000
+
+
+def _get_admin_injection_time_to_zero(dcm_result: pd.DataFrame) -> None:
+    """The administration mode should be defined as bolus if < 10 min and infusion if > 30 min."""
+    # todo : confirm ?
```
NicolasGensollen: Was this confirmed ?

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+            )
+
+
+def _set_time_from_ms_to_seconds(dcm_result: pd.DataFrame, field: str) -> None:
+    """Set time to seconds for FrameTimeStart and FrameDuration which are encoded as list(float), given in ms."""
+    if start := dcm_result.loc[field, "Value"]:
+        dcm_result.loc[field, "Value"] = start / 1000
+
+
+def _get_admin_injection_time_to_zero(dcm_result: pd.DataFrame) -> None:
```
NicolasGensollen: The name feels weird. Maybe `_set_mode_of_administration_from_injection_time` ??

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+    """The administration mode should be defined as bolus if < 10 min and infusion if > 30 min."""
+    # todo : confirm ?
+    if (injection := dcm_result.loc["InjectionStart", "Value"]) != "n/a":
+        if abs(injection) / 60 < 10:
```
NicolasGensollen: Maybe split to avoid repeating the ".loc(....)" three times:

```python
dcm_result.loc["ModeOfAdministration", "Value"] = _get_mode_of_administration_from_injection_time(
    injection_time
)

def _get_mode_of_administration_from_injection_time(injection_time: int) -> str:
    if abs(injection_time) / 60 < 10:
        return "bolus"
    ...
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+    if corrected == "START" or corrected == "ADMIN":
+        dcm_result.loc["ImageDecayCorrected", "Value"] = True
+    else:
+        dcm_result.loc["ImageDecayCorrected", "Value"] = False
+
+
+def _set_decay_time(dcm_result: pd.DataFrame) -> None:
+    correction = dcm_result.loc["ImageDecayCorrectionTime", "Value"]
+
+    if correction == "START":
```
NicolasGensollen, line 280: Small suggestion to simplify a bit:

```python
decay_time = None
if correction == "START":
    decay_time = dcm_result.loc["ScanStart", "Value"]
elif correction == "ADMIN":
    decay_time = dcm_result.loc["InjectionStart", "Value"]
dcm_result.loc["ImageDecayCorrectionTime", "Value"] = decay_time
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+            dcm_result.loc["ModeOfAdministration", "Value"] = "bolus"
+        elif abs(injection) / 60 > 30:
+            dcm_result.loc["ModeOfAdministration", "Value"] = "infusion"
+        else:
+            dcm_result.loc["ModeOfAdministration", "Value"] = "bolus-infusion"
+
+
+def _check_decay_correction(dcm_result: pd.DataFrame) -> None:
+    corrected = dcm_result.loc["ImageDecayCorrected", "Value"]
+    if corrected == "START" or corrected == "ADMIN":
```
NicolasGensollen: Small suggestion to simplify:

```python
dcm_result.loc["ImageDecayCorrected", "Value"] = corrected in ("START", "ADMIN")
```

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+import numpy as np
+import pandas as pd
+from pydicom.dataset import DataElement, FileDataset
+from pydicom.multival import MultiValue
+
+from clinica.utils.stream import cprint
+
+from .bids import Modality
+
+# todo : __all__
```
NicolasGensollen: Still needs to be done

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,342 @@
...
+
+# todo : can this be used for other than AIBL ?
+# todo : would using a class be beneficial ? I am starting to get a lot of : if modality do something
```
NicolasGensollen: At the moment it's not that shocking (I think there are two places where you condition on the modality). But I agree that using classes would be a totally valid design as well. All your small functions mutating the dataframe would be rewritten as methods mutating an internal dataframe and you could indeed use inheritance to handle the modality:

```python
class DicomMetaData:
     def __init__(self):
         self._meta: pd.DataFrame = pd.DataFrame(
             columns=["BIDSname", "DCMtags", "Value"],
             data=self._get_default_data()
         ).set_index(keys="BIDSname", drop=False)

     def _get_default_data(self) -> list:
         return [
         # Hardware
             ["DeviceSerialNumber", ("DeviceSerialNumber",), "n/a"],
             ["ManufacturersModelName", ("ManufacturerModelName",), "n/a"],
             ["SoftwareVersions", ("SoftwareVersions",), "n/a"],
             ["BodyPart", ("BodyPartExamined",), "n/a"],
             ["MagneticFieldStrength", ("MagneticFieldStrength",), "n/a"],
             # Institution
             ["InstitutionName", ("InstitutionName",), "n/a"],
             ["InstitutionAddress", ("InstitutionAddress",), "n/a"],
             [
                 "InstitutionalDepartmentName",
                 ("InstitutionalDepartmentName",),
                 "n/a",
             ],
         ]

    def process(self, dcm_dir: Path):
        # do the generic stuffs
        ...

class PetDicomMetaData(DicomMetaData):
    def _get_default_data(self) -> list:
        return super()._get_default_data() + [......]

    def process(self, dcm_dir: Path):
        super().process(dcm_dir)
        # do the pet specific stuffs
        ...
```

If we are unsure, we can leave it like this and rethink about that when adding other modalities or adapting for other converters. I imagine this is when things will get clearer...

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: I think I wanted to indicate it would depend on modality but the function signature would show that

AliceJoubert: Yes I forgot to remove it

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: I think that was because I thought it was unnecessary to go through the "else" every time especially since the option "DEBUG" is rarely used. But that does not gain much time anyway, I can remove it

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: It is not explicitly `None` but it can be an empty tuple `()` which amounts to the same in the way it is handled. I can change to `Optional`

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: You're right, I am just unsure how to handle the output types for the recursive function

clinica/converters/aibl_to_bids/utils/json.py
NicolasGensollen: If it cannot be None, then we shouldn't add Optional, `tuple[str, ...]` is fine. Sorry for the confusion.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,340 @@
...
+                "InstitutionalDepartmentName",
+                ("InstitutionalDepartmentName",),
+                "n/a",
+            ],
+        ],
+    ).set_index(keys="BIDSname", drop=False)
+
+
+def _check_dcm_value(
+    dicom_data: Optional[Any],
```
NicolasGensollen, line 174: `Any` is used to say that the variable can be of any type, including None. So technically, this is equivalent to `dicom_data: Any`. However, it makes the code a bit more clear about the fact that dicom_data can be None, so we can leave it like this.

clinica/converters/aibl_to_bids/utils/json.py
```diff
@@ -0,0 +1,340 @@
...
+    if dicom_data:
+        if type(dicom_data) == MultiValue:
+            return list(dicom_data)
+        return dicom_data
+    return None
+
+
+def _fetch_dcm_data_from_header(
+    dicom_header: Union[FileDataset, DataElement], dicom_tags: Optional[tuple[str, ...]]
+) -> Optional[Union[Any, DataElement]]:
```
NicolasGensollen: Here, however, the union of `Any` with something is always `Any`, so let's go with:

```suggestion
) -> Optional[Any]:
```

test/unittests/converters/aibl_to_bids/test_aibl_json.py
```diff
@@ -0,0 +1,218 @@
...
+    _update_metadata_from_image_dicoms(pd.DataFrame(), tmp_path)
+    mock_cprint.assert_called_once_with(
+        msg=f"No DICOM found at {tmp_path}, the image json will be filled with default values",
+        lvl="warning",
+    )
+
+
+# todo : mock to test for success scenario
+
+# def test_update_metadata_from_image_dicoms_success(tmp_path, mocker):
```
NicolasGensollen: Is this test supposed to be restored or should we remove it ?

clinica/converters/aibl_to_bids/utils/json.py
NicolasGensollen: Let's close this thread if we both agree on the short term solution.

clinica/converters/aibl_to_bids/utils/json.py
NicolasGensollen: Closing, see latest review comments.

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: I checked with Ninon and actually we cannot really use this to determine this field. I will just indicate that I could not find the associated dicom key and probably open an issue about it later, as for all the other fields I could not find

clinica/converters/aibl_to_bids/utils/json.py
AliceJoubert: I saw [here](https://github.com/python/mypy/issues/3138) it was not so easy so just in case I kept the `Optional`

test/unittests/converters/aibl_to_bids/test_aibl_json.py
AliceJoubert: No we should keep it but I can't make it work and need to figure out why

PR #1507, opened 2025-04-11
[FIX] Genfi-to-BIDS : modification of the source path of the image in scans tsv files

### LINKED ISSUE #1504: Genfi-to-BIDS `scans` write paths of source files from the root
In the Genfi-to-BIDS converter, the scans tsv files indicate the path to the image source file. When comparing tsv files in non regression tests it compares what is inside the CI data to files that are created from the CI machines. For example : 

```bash
[left]: [/mnt/data_ci/GenfiToBids/in/unorganized/GRN001-01-MR00/2/DICOM/2.25.84396454062804760632627598104588575299-2-132-usuls5.dcm, ...]
[right]: [/Users/alice.joubert/clinica_data_ci/data_ci/GenfiToBids/in/unorganized/GRN001-01-MR00/2/DICOM/2.25.84396454062804760632627598104588575299-2-36-107025a.dcm, ...]
```

We should modify this to get the path of the image starting from the given BIDS directory. Here, that would be for example `/Users/alice.joubert/clinica_data_ci/data_ci/GenfiToBids/in/unorganized/`.

AliceJoubert:
Closes #1504

I will do the modification of the CI right after this one is merged and I rebased my branch for my other PR #1484. The nonregression tests should fail only because the scans files are now named `scans` and not `scan` in the CI reference.

### DIFF
--- a/clinica/converters/genfi_to_bids/_converter.py
+++ b/clinica/converters/genfi_to_bids/_converter.py
@@ -96,6 +96,7 @@ def convert(
         participants=results["participants"],
         sessions=results["sessions"],
         scans=results["scans"],
+        source=path_to_dataset,
     )
     write_modality_agnostic_files(
         study_name=StudyName.GENFI,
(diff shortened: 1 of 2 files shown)


NicolasGensollen (approved):
LGTM, thanks @AliceJoubert !

Let's follow your plan, merging this one...

PR #1511, opened 2025-04-24
[REF] ADNI-to-BIDS : Keeping parameter mod_to_update instead of force-new-extraction option

### LINKED ISSUE #1461: ADNI-to-BIDS : misleading parameter `mod_to_update` vs option `--force-new-extraction`
In ADNI-to-BIDS client, the user can use the option `--force-new-extraction` to write over existing files if the converter is ran over the same folder several times. At some point in the code the parameter becomes `mod_to_update`:

https://github.com/aramis-lab/clinica/blob/7b148a746f7a97c713c2cbb0e7e31d1ce3eef9b6/clinica/converters/adni_to_bids/_converter.py#L212-L222

The user will get warning logs using the `mod_to_update` parameter :

```console
2025-03-10 18:22:13,311:WARNING:There already exist images : [PosixPath('/Users/alice.joubert/data/ADNI/BIDSres/sub-ADNI023S4115/ses-M114/anat/sub-ADNI023S4115_ses-M114_T1w.nii.gz')].
The parameter 'mod_to_update' is set to False so that they cannot be overwritten.
```

This is misleading and could be easily avoided by keeping one or the other.

Adam-Ismaili-92:
Closes #1461 

Documentation needs to be changed since the option use has changed.

### DIFF
--- a/clinica/converters/adni_to_bids/cli.py
+++ b/clinica/converters/adni_to_bids/cli.py
@@ -20,10 +20,10 @@ from ._utils import ADNIModality
     help="Convert clinical data only.",
 )
 @click.option(
-    "-u",
-    "--mod_to_update",
+    "-f",
+    "--force-new-extraction",
     is_flag=True,
-    help="Defines updating mode. Force overwriting and new data extraction if set to True.",
+    help="Force new data extraction.",
 )
 @click.option(
     "-m",
@@ -45,7 +45,7 @@ def cli(
     xml_path: Optional[str] = None,
     subjects_list: Optional[str] = None,
     clinical_data_only: bool = False,
-    mod_to_update: bool = False,
+    force_new_extraction: bool = False,
     modalities: Optional[Iterable[Union[str, ADNIModality]]] = None,
     n_procs: Optional[int] = None,
 ) -> None:
@@ -58,9 +58,9 @@ def cli(
 
     from ._converter import convert
 
-    if clinical_data_only and mod_to_update:
+    if clinical_data_only and force_new_extraction:
         raise ClinicaParserError(
-            "Arguments `clinical_data_only` and `mod_to_update` are mutually exclusive."
+            "Arguments `clinical_data_only` and `force_new_extraction` are mutually exclusive."
         )
     convert(
         dataset_directory,
@@ -70,7 +70,7 @@ def cli(
         subjects=subjects_list,
         modalities=modalities or list(ADNIModality),
         xml_path=xml_path,
-        mod_to_update=mod_to_update,
+        force_new_extraction=force_new_extraction,
         n_procs=n_procs,
     )
 
--- a/clinica/converters/adni_to_bids/cli.py
+++ b/clinica/converters/adni_to_bids/cli.py
@@ -20,10 +20,10 @@ from ._utils import ADNIModality
     help="Convert clinical data only.",
 )
 @click.option(
-    "-f",
-    "--force-new-extraction",
+    "-u",
+    "--mod_to_update",
     is_flag=True,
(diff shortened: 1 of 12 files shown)


clinica/converters/adni_to_bids/cli.py
```diff
@@ -58,9 +58,9 @@ def cli(
 
     from ._converter import convert
 
-    if clinical_data_only and force_new_extraction:
+    if clinical_data_only and mod_to_update:
```
NicolasGensollen, line 61: Unrelated to this PR, but this should be moved out of the client. Otherwise, users relying on the Python API through the `convert()` function won't get the same behavior.

clinica/converters/adni_to_bids/cli.py
```diff
@@ -20,10 +20,10 @@
     help="Convert clinical data only.",
 )
 @click.option(
-    "-f",
-    "--force-new-extraction",
+    "-u",
+    "--mod_to_update",
```
NicolasGensollen, line 24: Doing that is a breaking change since it changes the command line interface. Users relying on the option `-f` will get their code broken.

Technically this is possible but we should put a deprecation cycle in place for a couple of releases to warn users about the change and let them adapt their code accordingly.

Generally speaking, we avoid breaking the public API as much as possible and we do it only for good reasons. Here, I don't really think this qualifies as a good reason because "mod_to_update" is very obscure (to me at least). I actually think "force-new-extraction" was much clearer.

That being said, I totally agree with the initial issue and the fact that the logging messages need to use the correct argument name. I just think we should update "mod_to_update" into "force_new_extraction" in the underlying code rather than the opposite in the public API.

WDYT ?

NicolasGensollen (commented):
Thanks for working on that @Adam-Ismaili-92 !

clinica/converters/adni_to_bids/cli.py
Adam-Ismaili-92: I agree with your suggestion, it would indeed avoid processes that are too complex for the task at hand.

It just seemed simpler to me to replace "force_new_extraction" with "mod_to_update" to avoid modifications that would go beyond the converter we're dealing with and would modify and potentially break more general code.

I will try to think of a solution that would allow "mod_to_update" to be replaced by "force_new_extraction" instead.

NicolasGensollen (commented):
Thanks @Adam-Ismaili-92 !

You need to modify upstream code as well in order to have consistent naming everywhere. 

It starts here:

https://github.com/aramis-lab/clinica/blob/9a987a733ba9acabc737e2db9e0aa68246f87a3f/clinica/converters/adni_to_bids/_converter.py#L221-L229

Where we should have `force_new_extraction=force_new_extraction`.

And then, there is a bit of work to do because all modality converters need to be updated, for example:

https://github.com/aramis-lab/clinica/blob/9a987a733ba9acabc737e2db9e0aa68246f87a3f/clinica/converters/adni_to_bids/modality_converters/_dwi.py#L11-L18

As well as downstream code, for example:

https://github.com/aramis-lab/clinica/blob/9a987a733ba9acabc737e2db9e0aa68246f87a3f/clinica/converters/adni_to_bids/_utils.py#L644-L650

Does that make sense ?

Adam-Ismaili-92:
> force_new_extraction

Yes it makes sense. I haven't done those modifications at first thinking the name "mod_to_update" could make sense farer in the code, but if it doesn't I am willing to make the modifications.

NicolasGensollen (approved):

NicolasGensollen:
The two failures are already on dev and are unrelated, I'm gonna merge and we will investigate those later.

PR #1468, opened 2025-03-13
[TEST][FIX] GENFI-to-BIDS : Add comparison of metadata files in non-regression tests

AliceJoubert:
Added non-regression tests for metadata files.

Fixes : 
- To follow BIDS specs : name of metadata files to "scans" instead of "scan" ; name of field "filename" instead of "bids_filename" in the scans metadata file
- Modification of dataframe operations for participants metadata file because there were duplicate lines

Remark : 
Functions are really long, it would be beneficial to split these (ex `write_bids`) in order to unit test it.

PR #1512, opened 2025-04-28
[REF] GENFI -to-BIDS : Use source_path.parent instead of source_path in scans metadata

### LINKED ISSUE #1509: Genfi-to-BIDS scans tsv file writing is not deterministic
In converter Genfi-to-BIDS, the source path for the dicom image seems to not be deterministic. Maybe that is cause by the following function : 

https://github.com/aramis-lab/clinica/blob/68b2234d9a8ff62fded1437ca51e98dbd2479be9/clinica/converters/genfi_to_bids/_utils.py#L39-L83

where duplicates are dropped without re-arranging the index first to keep the order.

This causes the "source_path" in scans.tsv files to be possibly different for the same command line and input files.

AliceJoubert:
Closes #1509 

The issue caused failing non-regression tests when comparing scans.tsv files. Instead of trying to get the same dicom file every time upstream as described in 1509, I chose to modify the path written in the scans.tsv file. To modify that properly I would have had to unit test every intermediate function which is not my goal right now, though it has been added to the to-do list.

Current non-regression tests for GENFI can be expected to fail with this PR but should not when #1468 is rebased

### DIFF
--- a/clinica/converters/genfi_to_bids/_utils.py
+++ b/clinica/converters/genfi_to_bids/_utils.py
@@ -67,7 +67,6 @@ def _filter_dicoms(df: DataFrame) -> DataFrame:
         "localiser",
         "localizer",
     ]
-
     df = df.drop_duplicates(subset=["source"])
     df = df.assign(
         series_desc=lambda x: x.source_path.apply(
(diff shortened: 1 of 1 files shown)


clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -859,7 +858,7 @@ def write_bids(
                 / str(metadata.session_id)
                 / f"{metadata.participant_id}_{metadata.session_id}_scan.tsv"
             )
-            metadata.source_path = metadata.source_path.relative_to(source)
+            metadata.source_path = metadata.source_path.relative_to(source).parent
```
NicolasGensollen, line 861: I'm not sure to understand. This isn't really fixing the non deterministic issue but avoiding it by writing the parent folder instead, right ?
I'm afraid we will lose the information about which dicom was used like this.

PR #1518, opened 2025-05-22
[REF] Factorizing ```install_nifti```

### LINKED ISSUE #826: Factorize `install_nifti`
Both converters `habs2bids` and `nifd2bids` have their own implementation of `install_nifti`:

https://github.com/aramis-lab/clinica/blob/83d6d9958b3b3e2ffe82e5fedf09cd70b3109a44/clinica/iotools/converters/habs_to_bids/habs_to_bids.py#L178-L185

https://github.com/aramis-lab/clinica/blob/83d6d9958b3b3e2ffe82e5fedf09cd70b3109a44/clinica/iotools/converters/nifd_to_bids/nifd_utils.py#L320-L328

I believe these could be unified and factored out (in `bids_utils` maybe ?).

Also, the name of the function could be changed to something more explicit.

Adam-Ismaili-92:
Closes #826

### DIFF
--- a/clinica/converters/habs_to_bids/_converter.py
+++ b/clinica/converters/habs_to_bids/_converter.py
@@ -195,14 +195,15 @@ def _parse_imaging_data(paths: list[tuple[str, str]]) -> Optional[pd.DataFrame]:
     return df
 
 
+"""
 def _install_nifti(zipfile: str, filename: str, bids_path: str) -> None:
-    """Install a NIfTI file from a source archive to the target BIDS path."""
     import fsspec
 
     fo = fsspec.open(zipfile)
     fs = fsspec.filesystem("zip", fo=fo)
     with fsspec.open(bids_path, mode="wb") as f:
         f.write(fs.cat(filename))
+"""
 
 
 def _write_bids(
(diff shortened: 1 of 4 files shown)


clinica/converters/habs_to_bids/_converter.py
```diff
@@ -280,10 +271,10 @@ def _write_bids(
         )
 
         for filename, row in dataframe.iterrows():
-            _install_nifti(
-                zipfile=str(sourcedata / row.source_zipfile),
-                filename=row.source_filename,
-                bids_path=str(bids_basedir / filename),
+            install_nifti(
+                sourcedata_dir=str(sourcedata / row.source_zipfile),
```
NicolasGensollen, line 275: Why are you casting to strings when your function can handle both strings and Path objects ? Passing Path objects would avoid an unnecessary double casting and would be much clearer.

NicolasGensollen (commented):
Thanks for working on this @Adam-Ismaili-92 !

The PR LGTM, just one small suggestion.

clinica/converters/habs_to_bids/_converter.py
Adam-Ismaili-92: I kept the same logic used in the previous code (casting to str when the function is called), and made my function being able to handle both strings and Paths in case it is used with one or the other, but it is true that this casting becomes unnecessary. I will remove it.

NicolasGensollen (approved):

Issue #1505, opened 2025-04-11
Function to write metadata tsv files in Genfi-to-BIDS could be split and tested


AliceJoubert:
In Genfi-to-BIDS, writing the different metadata types of files is done in only one function : 

https://github.com/aramis-lab/clinica/blob/4247631b0b01c72f53eb15439afa4eb4a6b48ac0/clinica/converters/genfi_to_bids/_utils.py#L783-L873

Rather than having one big function which is not tested, it would be better to use this one as an interface to three separate functions specific to one type of file to test them more easily.

PR #1538, opened 2025-08-13
[REF] GENFI-To-BIDS : Separation of ```write_bids``` into 3 functions and tests added

Adam-Ismaili-92:
Closing #1505

### DIFF
--- a/clinica/converters/genfi_to_bids/_utils.py
+++ b/clinica/converters/genfi_to_bids/_utils.py
@@ -824,42 +824,22 @@ def _drop_duplicate_line_with_nans(participants: pd.DataFrame) -> pd.DataFrame:
     return participants
 
 
-def write_bids(
-    to: Path,
-    participants: DataFrame,
-    sessions: DataFrame,
-    scans: DataFrame,
-    source: Path,
-) -> None:
-    """This function writes the BIDS
+def _write_description_and_participants(to, fs, participants):
+    """This function writes the dataset_description.json and participants.tsv files of the BIDS architecture.
 
     Parameters
     ----------
     to: Path
         The path where the BIDS should be written
 
+    fs: LocalFileSystem
+        The local filesystem handler
+
     participants: DataFrame
         DataFrame containing the data for the participants.tsv
-
-    sessions: DataFrame
-        DataFrame containing the data for the sessions.tsv
-
-    scans: DataFrame
-        DataFrame containing the data for the scans.tsv
-
-    source : Path
-        Path to the source imaging data
     """
-    import os
-
-    from fsspec.implementations.local import LocalFileSystem
-
-    from clinica.converters._utils import run_dcm2niix, write_to_tsv
+    from clinica.converters._utils import write_to_tsv
     from clinica.dataset import BIDSDatasetDescription
-
-    cprint("Starting to write the BIDS.", lvl="info")
-    fs = LocalFileSystem(auto_mkdir=True)
-    # Ensure BIDS hierarchy is written first.
 
     participants = (
         participants.reset_index()
@@ -876,6 +856,23 @@ def write_bids(
         with fs.open(to / "participants.tsv", "w") as participant_file:
             write_to_tsv(participants, participant_file)
 
+
+def _write_sessions(to, fs, sessions):
+    """This function writes the session.tsv files of the BIDS architecture.
+
+    Parameters
+    ----------
+    to: Path
+        The path where the BIDS should be written
+
+    fs: LocalFileSystem
+        The local filesystem handler
+
+    sessions: DataFrame
+        DataFrame containing the data for the sessions.tsv
+    """
+    from clinica.converters._utils import write_to_tsv
+
     for participant_id, data_frame in sessions.groupby("participant_id"):
         sessions = data_frame.droplevel(
             ["participant_id", "modality", "bids_filename", "run_num"]
@@ -884,6 +881,25 @@ def write_bids(
         sessions_filepath = to / str(participant_id) / f"{participant_id}_sessions.tsv"
         with fs.open(sessions_filepath, "w") as sessions_file:
             write_to_tsv(sessions, sessions_file)
+
+
+def _write_scans(to, source, scans):
+    """This function writes the scans.tsv files of the BIDS architecture.
+
+    Parameters
+    ----------
+    to: Path
+        The path where the BIDS should be written
+
+    source : Path
+        Path to the source imaging data
+
+    scans: DataFrame
+        DataFrame containing the data for the scans.tsv
+    """
+    import os
+
+    from clinica.converters._utils import run_dcm2niix
 
     scans = scans.reset_index().set_index(["bids_full_path"], verify_integrity=True)
 
@@ -920,6 +936,47 @@ def write_bids(
                     metadata.number_of_parts,
                     metadata.run_num,
                 )
+
+
+def write_bids(
+    to: Path,
+    participants: DataFrame,
+    sessions: DataFrame,
+    scans: DataFrame,
+    source: Path,
+) -> None:
+    """This function writes the BIDS
+
+    Parameters
+    ----------
+    to: Path
+        The path where the BIDS should be written
+
+    participants: DataFrame
+        DataFrame containing the data for the participants.tsv
+
+    sessions: DataFrame
+        DataFrame containing the data for the sessions.tsv
+
+    scans: DataFrame
+        DataFrame containing the data for the scans.tsv
+
+    source : Path
+        Path to the source imaging data
+    """
+    import pandas as pd
+    from fsspec.implementations.local import LocalFileSystem
+
+    cprint("Starting to write the BIDS.", lvl="info")
+    fs = LocalFileSystem(auto_mkdir=True)
+    # Ensure BIDS hierarchy is written first.
+
+    _write_description_and_participants(to, fs, participants)
+
+    _write_sessions(to, fs, sessions)
+
+    _write_scans(to, source, scans)
+
     _correct_fieldmaps_name(to)
     _delete_real_and_imaginary_files(to)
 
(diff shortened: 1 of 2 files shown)


clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -824,43 +824,23 @@ def _drop_duplicate_line_with_nans(participants: pd.DataFrame) -> pd.DataFrame:
...
 
-def write_bids(
-    to: Path,
-    participants: DataFrame,
-    sessions: DataFrame,
-    scans: DataFrame,
-    source: Path,
-) -> None:
-    """This function writes the BIDS
+def _write_description_and_participants(to, fs, participants):
```
```suggestion
def _write_description_and_participants(to:Path, fs:LocalFileSystem, participants:DataFrame):
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -876,6 +856,23 @@ def write_bids(
         with fs.open(to / "participants.tsv", "w") as participant_file:
             write_to_tsv(participants, participant_file)
 
+
+def _write_sessions(to, fs, sessions):
```
```suggestion
def _write_sessions(to:Path, fs:LocalFileSystem, sessions:DataFrame):
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -885,6 +882,25 @@ def write_bids(
         with fs.open(sessions_filepath, "w") as sessions_file:
             write_to_tsv(sessions, sessions_file)
 
+
+def _write_scans(to, source, scans):
```
```suggestion
def _write_scans(to:Path, source:Path, scans:DataFrame):
```

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: Actually I noticed the use of both `DataFrame` and `pd.DataFrame` in this file (see the imports at the beginning of the file). Do you think you can harmonize that in this PR ?

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -885,6 +882,25 @@ def write_bids(
...
+    """This function writes the scans.tsv files of the BIDS architecture.
```
AliceJoubert: This function also convert the images from dicom to niftis. It is debatable as to whether or not we should keep it this way but at least the name of the function should reflect that.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -920,6 +936,47 @@ def write_bids(
...
+
+
+def write_bids(
+    to: Path,
+    participants: DataFrame,
+    sessions: DataFrame,
+    scans: DataFrame,
+    source: Path,
+) -> None:
+    """This function writes the BIDS
```
```suggestion
    """This function writes the BIDS folder with both images and metadata files.
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
             }
         ),
     )
+
+
+def test_write_description_and_participants(tmp_path):
+    import pandas as pd
```
```suggestion
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+    from fsspec.implementations.local import LocalFileSystem
```
AliceJoubert: If you use it several times you can put it at the top of the file

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+    participants = pd.DataFrame(
+        {
+            "participant_id": ["sub-C9ORF002", "sub-MAPT003"],
+            "session_id": [None, None],
+            "modality": [None, None],
+            "run_num": [None, None],
+            "bids_filename": [None, None],
+            "source": [None, None],
+        }
+    )
```
AliceJoubert: It would be interesting to use actual values to see how it performs. You can also include special cases handled by the function, for example cases with `np.nan` : 
https://github.com/aramis-lab/clinica/blob/8a79ccbecb0efe54ffe3ae5b34c479b8dbf4c624/clinica/converters/genfi_to_bids/_utils.py#L869-L871

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+
+    # Check participants.tsv content
+    df_out = pd.read_csv(tsv_file, sep="\t")
+    assert "participant_id" in df_out.columns
+    assert set(df_out["participant_id"]) == {"sub-C9ORF002", "sub-MAPT003"}
+
+
+def test_write_sessions(tmp_path):
+    import pandas as pd
+    from fsspec.implementations.local import LocalFileSystem
```
AliceJoubert: Same remarks as previously

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+        {
+            "participant_id": [participant_id, participant_id],
+            "modality": [None, None],
+            "bids_filename": [None, None],
+            "run_num": [None, None],
+            "session_id": ["ses-M000", "ses-M070"],
+        }
+    ).set_index(
+        ["participant_id", "modality", "bids_filename", "run_num", "session_id"]
+    )
```
AliceJoubert: Once again, don't hesitate to use actual values to see how the function behaves

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+    tsv_path = tmp_path / participant_id / f"{participant_id}_sessions.tsv"
+    assert tsv_path.exists()
+
+    # Check sessions.tsv content
+    df_out = pd.read_csv(tsv_path, sep="\t")
+    session_ids = (
+        set(df_out["session_id"])
+        if "session_id" in df_out.columns
+        else set(df_out.index.astype(str))
+    )
```
AliceJoubert: Can you give me an example where  you need this case ?

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+    assert session_ids == {"ses-M000", "ses-M070"}
+
+
+def test_write_scans(tmp_path, monkeypatch):
+    import subprocess
+
+    import pandas as pd
```
```suggestion
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+    from clinica.converters.genfi_to_bids._utils import _write_scans
+
+    # Make dcm2niix succeed without running anything to bypass it
+    class FakeCompleted:
+        def __init__(self):
+            self.returncode = 0
+            self.stdout = b""
+            self.stderr = b""
+
+    monkeypatch.setattr(subprocess, "run", lambda *a, **k: FakeCompleted())
```
AliceJoubert: Technically it works so I am not against it. I am just wondering about the homogeneity of the code. If you looked around a bit in the code for tests you might have seen `mock` or `mocker` is being used, like that :

https://github.com/aramis-lab/clinica/blob/8a79ccbecb0efe54ffe3ae5b34c479b8dbf4c624/test/unittests/utils/test_check_dependency.py#L401-L412

The logic is a bit different. If I understand correctly in your code you are mocking the `subprocess` command while the example I show you is mocking the function in clinica. In you case that would be the `run_dcm2niix` function. I feel like it is more straightforward to mock the function that is directly used in _write_scans instead of an underlying subprocess but I might be wrong. Tell me what you think !

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+        "index",
+        "filename",
+        "modality",
+        "run_num",
+        "source_path",
+        "manufacturer",
+        "number_of_parts",
+    ]
+
+    print(list(df_out.columns))
```
```suggestion
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -144,3 +144,155 @@ def test_drop_duplicate_line_with_nans():
...
+
+    assert list(df_out.columns) == expected_cols
```
AliceJoubert: Maybe use a `set` here to avoid order issues

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Indeed, this function might need to be further reformatted. I chose to just change the name of the function to better reflect what it does. Tell me if you think further reformatting should be done in the PR.

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: It was done to avoid getting an error or an empty set if the column named "session_id" could not be retrieved, but just putting ```assert set(df_out["session_id"]) == {"ses-M000", "ses-M070"}``` might work just fine.

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: Yes, my intention was to mock the ```run_dcm2niix``` function, since it is checked in the function ```_write_scans_and_niftis``` before proceeding to the rest of it. I think the ```subprocess```is mocking effectively and directly the ```run_dcm2niix```function, but for harmonization reasons I can use ```mock``` or ```mocker```.

test/unittests/converters/test_genfi_to_bids_utils.py
AliceJoubert: Since you would want to catch the case where you can't get the "session_id" column it's best to not put that condition indeed

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: No it is fine in this PR. However we should think about it in another PR.

AliceJoubert (approved):

PR #1541, opened 2025-08-22
[DOC] Duplicates information about dataset description in our BIDS specifications

### LINKED ISSUE #1540: No documentation on the dataset_description
Hi clinica devs,

Clinica pipelines will not work unless a `dataset_description.json`. However, what is expected in that file is not described in the documentation, meaning the user has to go find it in the BIDS spec (not that easy).
This is problematic, especially since clinica seems to expect the field `DatasetType` which is not mandatory.

Have a good day

AliceJoubert:
Closes #1540


docs/snippets/dataset_description.md
Adam-Ismaili-92, line 1: Wouldn’t having exactly the same explanatory text in the ```dataset_description.json``` file for both the CAPS and BIDS sections be a bit redundant and heavy? 
Maybe it would be better, in the CAPS section, to simply redirect to the explanation already provided in the BIDS section, for example? Tell me what you think.

Adam-Ismaili-92:
Thank you for your work @AliceJoubert ! I await for your response before doing any further steps.

docs/snippets/dataset_description.md
AliceJoubert: I can do that if you prefer. I will be fair, both solutions are not very satisfying to me but this whole section (Clinica environment) is bound to be reworked. For ex I was planning on adding a section on Nipoppy so that will probably be reworked at the same time. For now, I can just link to the BIDS file.

PR #1542, opened 2025-08-22
[REF] Support SPM25 for `t1-volume`

AliceJoubert:
SPM25 standalone only
ADD test 
Verify documentation

Don't wait too much for other pipelines

### DIFF
--- a/clinica/pipelines/t1_volume/info.json
+++ b/clinica/pipelines/t1_volume/info.json
@@ -8,7 +8,7 @@
         {
             "type": "software",
             "name": "spm",
-            "version": ">=12"
+            "version": ">=25"
         }
     ]
 }
--- a/clinica/pipelines/t1_volume_create_dartel/info.json
+++ b/clinica/pipelines/t1_volume_create_dartel/info.json
@@ -8,7 +8,7 @@
         {
             "type": "software",
             "name": "spm",
-            "version": ">=12"
+            "version": ">=25"
         }
     ]
 }
--- a/clinica/pipelines/t1_volume_dartel2mni/info.json
+++ b/clinica/pipelines/t1_volume_dartel2mni/info.json
@@ -8,7 +8,7 @@
         {
             "type": "software",
             "name": "spm",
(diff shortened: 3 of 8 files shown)


Adam-Ismaili-92 (changes requested):
Thanks @AliceJoubert for your work !

The tests were mainly focused on **non-regression** ones.

The tests have only been conducted on **Linux**, third party softwares (MATLAB, MCR) still need to be properly downloaded on Mac for the remaining tests to be conducted.

On Linux :
- **SPM12** works well (throwing an error recommending SPM25 installation)
- **SPM12 Standalone** works well (throwing an error recommending SPM25 installation)
- **SPM25** sends assertion errors or loads indefinitely
- **SPM25 Standalone** sends an error probably linked to an installation or compatibility issue with MCR : `”OSError: This docstring was not generated by Nipype!”`

Adam-Ismaili-92 (approved):
Unit tests show that SPM25 Standalone is recognized well by Clinica. 
However, some weird errors regarding checks done on SPM arise during unit tests completion, which might need to be updated. Those errors are similar between MacOS and Linux : 
Capture d’écran 2025-11-06 à 17 40 06

Through the use of CI data, almost all non-regression tests on T1-volume pass successfully with SPM25 Standalone, with just one exception : `test-t1-volume-tissue-segmentation`, which sends different outputs based on the type of OS used.

- On MacOS, `test-t1-volume-tissue-segmentation` sends an error indicating that a file  named `wsub-ADNI011S4105_ses-M000_T1w.nii`, which was supposed to be generated, was not found :
Capture d’écran 2025-11-06 à 17 53 59

- On Linux, `test-t1-volume-tissue-segmentation` sends an error regarding a failed assertion in the `likeliness` measure :
Capture d’écran 2025-11-06 à 17 14 36

Those errors need to be addressed in their specific issues and PRs.

To sum up, this PR works well in integrating SPM25 Standalone support to Clinica's T1-volume.

LGTM ! Thanks for your work @AliceJoubert !

PR #1547, opened 2025-09-17
[REF] Support SPM25 for `pet-volume`

AliceJoubert:
TODO : doc
Check interfaces
Modify CI modules


Adam-Ismaili-92 (changes requested):
Thanks @AliceJoubert for your work !

The tests were mainly focused on **non-regression** ones.

The tests have only been conducted on **Linux**, third party softwares (MATLAB, MCR) still need to be properly downloaded on Mac for the remaining tests to be conducted.

On Linux :
- **SPM12** works well (throwing an error recommending SPM25 installation)
- **SPM12 Standalone** works well (throwing an error recommending SPM25 installation)
- **SPM25** sends an error that should have been handled in the PR #1543 : `“AttributeError: 'DynamicTraitedSpec' object has no attribute 'in_image’”`
- **SPM25 Standalone** sends an error probably linked to an installation or compatibility issue with MCR : `”OSError: This docstring was not generated by Nipype!”`

Adam-Ismaili-92 (approved):
Unit tests show that SPM25 Standalone is well recognized by Clinica.
However, some weird errors regarding checks done on SPM arise during unit tests completion, which might need to be updated. Those errors are similar between MacOS and Linux :
Capture d’écran 2025-11-06 à 17 40 06

Through the use of the current CI data, one test doesn't pass with SPM25 Standalone : `test-pet-volume`, which sends the same output on Linux and MacOS. The error regards a failed assertion on the `likeliness` measure :
Capture d’écran 2025-11-06 à 19 32 57

Those errors need to be addressed in their specific issues and PRs.

To sum up, this PR works well in integrating SPM25 Standalone support to Clinica's PET-volume.

LGTM ! Thanks for your work @AliceJoubert !

PR #1563, opened 2025-11-18
[FIX] IXI-to-BIDS : Setting `participant_id` as the first column of `/participants.tsv`

### LINKED ISSUE #1560: [BUG] IXI-to-BIDS : Incorrect column order in `/participants.tsv`
**Describe the bug**
After applying the [BIDS validator](https://bids-standard.github.io/bids-validator/) on the architecture resulting from the BIDS conversion of the CI data, an error occurs indicating that the file `/participants.tsv` has incorrect column order.

**To Reproduce**
1. Get the IXI raw subset from the CI data
2. Convert the IXI raw subset into the BIDS format
3. Go to the [BIDS validator](https://bids-standard.github.io/bids-validator/)
4. Select the IXI subset in BIDS format
4. See the error

**Expected behavior**
This error should be thrown :
`Some TSV columns are in the incorrect order:`
`/participants.tsv (Initial column 1 found at index 11.)`

**Screenshots**
Image

**Desktop:**
 - OS: MacOS Sequoia 15.7.1
 - Browser: Safari
 - Clinica version: 0.10.1

**Additional context**
This error could be explained by the fact that the column named `participant_id` is at the end of the `.tsv` file, while it should be at the beginning.

Adam-Ismaili-92:
Closes #1560

### DIFF
--- a/test/unittests/converters/test_ixi_to_bids_utils.py
+++ b/test/unittests/converters/test_ixi_to_bids_utils.py
@@ -664,10 +664,14 @@ def test_write_sessions(tmp_path):
 def test_write_participants(tmp_path):
     from clinica.converters.ixi_to_bids._utils import write_participants
 
-    clinical = formatted_clinical_data_builder().drop("participant_id", axis=1)
-    expected = formatted_clinical_data_builder()
-    write_participants(tmp_path, clinical, ["IXI001", "IXI002"])
-    expected.drop(["acq_time", "session_id"], axis=1, inplace=True)
+    write_participants(
+        tmp_path,
+        formatted_clinical_data_builder().drop("participant_id", axis=1),
+        ["IXI001", "IXI002"],
+    )
+    expected = formatted_clinical_data_builder().drop(
+        ["acq_time", "session_id"], axis=1
+    )
     expected = pd.concat(
         [expected, pd.DataFrame({col: ["n/a"] for col in expected.columns})]
     ).reset_index(drop=True)


clinica/converters/ixi_to_bids/_utils.py
```diff
@@ -399,6 +399,9 @@ def write_participants(
             clinical_data.loc[participant, "source_id"] = participant
     if not bids_dir.exists():
         bids_dir.mkdir()
+    clinical_data = clinical_data[
+        ["participant_id"] + clinical_data.columns.drop("participant_id").tolist()
+    ]
```
AliceJoubert: I would rather have you specify the column number above (line 391) rather than remove/add again the column.

Adam-Ismaili-92: It will not remove the column, but reorder it. 
`clinical_data.columns` returns the list of column **names**, `drop` removes the name `participant_id` from this list, and `["participant_id"] + list` places `participant_id` at the front.
Finally, the columns are reordered by using `clinical_data[["participant_id", ...]]`.

I could still find a way to specifically use the index of the column `participant_id`, but I don't think it is really necessary. However, I agree that I can place the reordering part right after the `.assign` at the line 391.

AliceJoubert: I read quickly the first time indeed, but the point still stands. At L391 the column `participant_id` is created, so you can modify that line to create it at the right position.

What is currently done is `df.assign(name=values)`. What you can do instead is a  `df.insert(0, name, values)`. This way you combine the creation of the column and the placing at the right position.

Adam-Ismaili-92:
> > > Thanks for the work @Adam-Ismaili-92 . I asked for a small modification. Otherwise, I have some questions :
> > > 
> > > * Did you run the BIDS validator again afterwards?
> > > * Did you run the non-regression tests locally ?
> > > * What do you think of adding the BIDS validator to the non-regression tests for converters ?
> > 
> > 
> > 
> > * I did run the BIDS validator afterwards, it works.
> > * I did run the non-regression tests locally, it works with the new correction.
> > * I think it is a good idea, allowing us to check consistently for BIDS architecture compliance produced by Clinica's converters. However, I am not so sure if it is a good idea to integrate it right now, as a lot of converters probably need to be upgraded to match current BIDS specifications, making the non regression tests automatically fail for an undefined amount of time.
> 
> Ok for the tests. The goal would be to integrate it one converter after the other. It could be ran for IXI only for example. WDYT ?

I think it is a clever idea to integrate it one converter after the other. I agree to this approach.

clinica/converters/ixi_to_bids/_utils.py
Adam-Ismaili-92: The issue is that the column `participant_id` already exists at the point of the `df.assign` ; It is not created by the command, just filled. This makes it impossible to replace the command by `df.insert`, otherwise throwing this kind of error : `ValueError: cannot insert participant_id, already exists`. 

Do you want me to drop the column before inserting, or to preserve the current logic ?
I could also find the place in the code where it is created to insert it at the first position.

AliceJoubert: I am curious as to how you tested that option because df.assign does create the column. Do you mind sharing that ? (privately if you would prefer)

Adam-Ismaili-92: > I am curious as to how you tested that option because df.assign does create the column. Do you mind sharing that ? (privately if you would prefer)

As we discussed, it was a problem of unit test needing to be updated. `df.assign` does indeed create a new column. It has been replaced by a `df.insert` as suggested, and the unit test has been updated.

test/unittests/converters/test_ixi_to_bids_utils.py
```diff
@@ -664,8 +664,8 @@ def test_write_sessions(tmp_path):
 def test_write_participants(tmp_path):
     from clinica.converters.ixi_to_bids._utils import write_participants
 
-    clinical = formatted_clinical_data_builder()
-    expected = clinical.copy()
+    clinical = formatted_clinical_data_builder().drop("participant_id", axis=1)
+    expected = formatted_clinical_data_builder()
     write_participants(tmp_path, clinical, ["IXI001", "IXI002"])
     expected.drop(["acq_time", "session_id"], axis=1, inplace=True)
```
AliceJoubert:
```suggestion
    
    write_participants(tmp_path, formatted_clinical_data_builder().drop("participant_id", axis=1), ["IXI001", "IXI002"])
    expected = formatted_clinical_data_builder().drop(["acq_time", "session_id"], axis=1, inplace=True)
```

Adam-Ismaili-92: The changes should be done in the current commit.

PR #1607, opened 2025-12-18
[ENH] GENFI-to-BIDS : `ses-M[months]` replaced with `ses-[visit]` for session ids

### LINKED ISSUE #1555: [ENH] GENFI-to-BIDS : `ses-M[months]`should be replaced with `ses-[visit]` for session folders names
Currently, the session folders produced by the converter GENFI-to-BIDS follow the BIDS standard naming convention `ses-M[months_from_baseline]`.

However, this naming convention is not really practical for research-based or values matching operations, since GENFI clinical data possess a column `Visit` respecting another standard :
- Entry values are numbers composed of **one** or **two** decimals
- First decimal (none or 1) corresponds to the GENFI phases (GENFI-1 or GENFI-2)
- Second decimal corresponds to the visit number

Thus, the current naming convention should be replaced by another which is also BIDS-compliant : `ses-[visit]`, matching the naming convention used in the clinical data.

Adam-Ismaili-92:
Closes #1555

### DIFF
--- a/docs/Converters/GENFItoBIDS.md
+++ b/docs/Converters/GENFItoBIDS.md
@@ -92,7 +92,7 @@ where:
     └── ...
     ```
 
-    Sessions naming respects the naming convention used in GENFI clinical data. In particular, for `ses-XY`, `X` represents the GENFI phase, and `Y` represents the visit number.
+    Sessions respect the naming convention used in GENFI clinical data. In particular, for `ses-XY`, `X` represents the GENFI phase, and `Y` represents the visit number.
 
 - `OPTIONS`:
     - `--clinical-data-dir/-cdd` is the path to the clinical data directory.


clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -380,75 +380,35 @@ def _compute_source_id_and_source_ses_id(df: pd.DataFrame) -> pd.DataFrame:
...
 
-    for col in ("acq_date", "baseline"):
-        df[col] = df[col].apply(lambda x: datetime.strptime(x, "%Y%m%d"))
-    return df.assign(
-        ses_month=lambda x: 12
-        * (x.acq_date.apply(lambda y: y.year) - x.baseline.apply(lambda y: y.year))
-        + (x.acq_date.apply(lambda y: y.month) - x.baseline.apply(lambda y: y.month)),
-        session_id=lambda x: x.ses_month.map(lambda y: f"ses-M{y:03d}"),
+    df = df.assign(
+        session_id=lambda x: x.source_ses_id.map(lambda y: f"ses-{y}"),
```
AliceJoubert: Just 2 small questions for my own understanding : 
- what is the type of `source_ses_id` in the original data ? (numerical, str, whatever?)
- Basically if visit_code = x then session_id = ses-x ?

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 . Do you think you can add unit tests for intermediary functions you touched (_compute_session, _compute version)? Can you also add a little information about visits in the documentation ? I feel like it is important to inform users how this specific converter works as it is now different from other converters we provide.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: - `source_ses_id` is a string in the original data.
- Yes, it is exactly that.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -380,75 +380,35 @@ def _compute_source_id_and_source_ses_id(df: pd.DataFrame) -> pd.DataFrame:
...
-    pd.Dataframe
-        Contains the baseline date.
-    """
-    df_1 = (
-        df[["source_id", "source_ses_id", "acq_date"]]
-        .groupby(["source_id", "source_ses_id"])
-        .min()
+        genfi_version=lambda x: x.source_ses_id.astype("int").apply(
+            lambda y: f"GENFI{len(str(y))}"
+        ),
```
AliceJoubert: If I understood correctly the goal is to get : 
- "01" --> 1
- "12" --> 2
___

If first decimal is never going to change you could do : 
```python
genfi_version=lambda x: f"GENFI{int(x.source_ses_id[0])+1}"
```
___ 

Or keep the check on the length of the string but write that in a separate function to make sure it is never greater than 2 : 

```python
def _check_version_from_visit(visit:str) -> Union[str, None] :
    if length:=len(int(visit)) > 2:
        raise(ValueError, "...")
    return length
```

and use it like that : 
```python
genfi_version=lambda x: f"GENFI{_check_version_from_visit(x.source_ses_id)}")
```

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: Then : 
- From the issue you wrote for the PR, the `second decimal corresponds to the visit number`. Is it normal to take both decimals here ?
- Why do you need to set `source_ses_id` as an `int` ?
- In any case, you can simplify your code with : 
```python
    df = df.assign(session_id=lambda x: "ses-"+x.source_ses_id)
```

Adam-Ismaili-92: - Yes, it is normal since both of the decimals are needed to compute the visit code as it is used in GENFI clinical data.
- I need to set it as an `int` afterwards to not break the codes logic, since this column converted to `int` was used several times later.
- Ok, I will try to add this suggestion.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: It was an old logic created before this PR. I can indeed change it to `genfi_version=lambda x: f"GENFI{int(x.source_ses_id[0])+1}"` since GENFI will have more than 2 versions in the near future.

AliceJoubert: Does this logic still apply if there are more versions of GENFI ? If yes that means version 3 would indicated "2X", version 4 "3X"...

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: Ok. Please update your issue then so it is clear you take both digits of the `source_ses_id` (not `visit`) to compute the session number, for future reference.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -380,75 +380,34 @@ def _compute_source_id_and_source_ses_id(df: pd.DataFrame) -> pd.DataFrame:
...
-    return df.assign(
-        ses_month=lambda x: 12
-        * (x.acq_date.apply(lambda y: y.year) - x.baseline.apply(lambda y: y.year))
-        + (x.acq_date.apply(lambda y: y.month) - x.baseline.apply(lambda y: y.month)),
-        session_id=lambda x: x.ses_month.map(lambda y: f"ses-M{y:03d}"),
+    df = df.assign(
+        session_id=lambda x: "ses-" + x.source_ses_id,
+        source_ses_id=lambda x: x.source_ses_id.astype("int"),
     )
+    return df
```
AliceJoubert, line 410:
```suggestion
    return df.assign(
        session_id=lambda x: "ses-" + x.source_ses_id,
        source_ses_id=lambda x: x.source_ses_id.astype("int"),
    )
```

AliceJoubert (commented):
Thanks for the work and the details @Adam-Ismaili-92, that is clearer. If you can add unit tests for these functions now that would be perfect !

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: No problem, the issue should have been updated.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Yes, it should work exactly like that.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: For which functions should I add a unit test ? Just `_compute_session_numbers` (the one with the code that you quoted above) ?

AliceJoubert: The ones you touched and don't have a unit test yet : both compute_session and compute_visit

AliceJoubert: Ideally these unit tests should be written before touching a function so you can see what happens when you change the code

Adam-Ismaili-92: The unit tests have been added. Thanks also for the advice

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -29,6 +29,56 @@ def test_identify_modality_is_nan():
...
+
+@pytest.fixture
+def input_df_compute_session_numbers() -> pd.DataFrame:
+    return pd.DataFrame(
+        {
+            "source_id": ["GRN001", "GRN001", "C9ORF001", "C9ORF001"],
+            "genfi_version": ["GENFI1", "GENFI2", "GENFI2", "GENFI3"],
+            "source_ses_id": ["01", "11", "17", "21"],
+        }
+    )
```
AliceJoubert: Good to use a fixture. I don't think you need 2 different ones for your tests though since they are very similar. Actually they should be the same I think. You can use a generic name like `_generate_clinical_data()`.

If it was done properly (unrelated to your work, definitely not your fault) we could have only one fixture of that sort for the whole file. It's good if the dataframe values match reality as much as possible (like `source_ses_id` being strings) as you did in your fixtures.

Adam-Ismaili-92:
> Thanks for the tests, just one last change and it should be good.

The modifications on the use of a single fixture instead of multiple similar ones have been done.

Should I also modify the documentation to explain in more details how the converter works with the new session names ?

PR #1608, opened 2025-12-29
[FIX] Fix command setting with MAC for SPM

### LINKED ISSUE #1614: [BUG] `T1-Volume` writes the result of SPM step in the wrong directory
With the new set-up SPM25 standalone + MCR, the pipeline `T1-Volume` does not complete anymore on Mac (M3), though it works well on Linux. After investigation, it seems that the cause of this is the command line set to work with SPM25 for Mac, which cd into SPM directory : 
https://github.com/aramis-lab/clinica/blob/55f70a296339506c2405d695ec2f148ad2c44904/clinica/utils/spm.py#L127-L146

In some cases  it writes the end result in there rather than inside Nipype temporary working directory. As a result, the result is not found by the pipeline and causes it to fail. What is surprising is that it does not happen every time and did not cause an issue for `PET-Volume`.

This behavior was introduced in PR#31 for `PET-Surface` but it looks like it is not needed anymore.

AliceJoubert:
Closes #1614

How to test : SPM25 standalone + MCR set-up with pipelines T1-Volume / PET-Volume

### DIFF
--- a/test/unittests/utils/test_spm.py
+++ b/test/unittests/utils/test_spm.py
@@ -106,14 +106,17 @@ def test_use_spm_standalone_if_available_error(tmp_path):
             use_spm_standalone_if_available()
 
 
-def test_get_platform_dependant_matlab_command_for_spm_standalone():
+def test_get_matlab_command_for_spm_standalone(mocker):
     from clinica.utils.spm import (
         _get_matlab_command_for_spm_standalone,
     )
 
+    mocker.patch(
+        "clinica.utils.spm._get_real_spm_standalone_file", return_value="run_spm25.sh"
+    )
     assert (
         _get_matlab_command_for_spm_standalone(Path("/foo/bar"), Path("/foo/bar/baz"))
-        == "/foo/bar/run_spm12.sh /foo/bar/baz script"
+        == "/foo/bar/run_spm25.sh /foo/bar/baz script"
     )
 
 


Adam-Ismaili-92 (commented):

Every tests have been done on MacOS.

- Unit tests surrounding `T1-volume` and `PET-volume` work well.
- `clinica run t1-volume` and `clinica run pet-volume` commands have been tested individually and also work well.
- However, non-regression tests on `T1-volume` and `PET-volume` doesn't seem to work, throwing an error on the `likeliness_measure` function, as you can see below : 
t1-volume
pet-volume
Is this expected ?

AliceJoubert:
> Thanks for your work @AliceJoubert !
> 
> Every tests have been done on MacOS.
> 
> * Unit tests surrounding `T1-volume` and `PET-volume` work well.
> * `clinica run t1-volume` and `clinica run pet-volume` commands have been tested individually and also work well.
> * However, non-regression tests on `T1-volume` and `PET-volume` doesn't seem to work, throwing an error on the `likeliness_measure` function, as you can see below :
> 
> t1-volume pet-volume Is this expected ?

Yes ! This could be expected. As long as the pipeline complete we can change the reference values.

Adam-Ismaili-92:
> > Thanks for your work @AliceJoubert !
> > Every tests have been done on MacOS.
> > 
> > * Unit tests surrounding `T1-volume` and `PET-volume` work well.
> > * `clinica run t1-volume` and `clinica run pet-volume` commands have been tested individually and also work well.
> > * However, non-regression tests on `T1-volume` and `PET-volume` doesn't seem to work, throwing an error on the `likeliness_measure` function, as you can see below :
> > 
> > t1-volume pet-volume Is this expected ?
> 
> Yes ! This could be expected. As long as the pipeline complete we can change the reference values.

Ok nice then !

PR #1617, opened 2026-01-13
[ENH] ADNI-to-BIDS : Code updated to handle `DXSUM_[DATE].csv`

### LINKED ISSUE #1572: [ENH] ADNI-to-BIDS : Code should be updated to handle `DXSUM_[DATE].csv`

This issue is particularly evident for the file `DXSUM_[DATE].csv`, previously `DXSUM_PDXCONV_ADNIALL.csv`, which is not handled correctly by the converter’s code.

Therefore, the code responsible for processing `DXSUM_[DATE].csv` should be updated accordingly.

- [ ] **MODIF 1** : Adapt the code to the name of the new file

- [ ] **MODIF 2** : Change attribute call from `.Phase` to `.PHASE`

Adam-Ismaili-92:
Closes #1572

### DIFF
--- a/clinica/converters/adni_to_bids/_utils.py
+++ b/clinica/converters/adni_to_bids/_utils.py
@@ -143,16 +143,9 @@ def correct_diagnosis_sc_adni3(
         Corrected participants_df.
     """
     diagnosis_dict = {1: "CN", 2: "MCI", 3: "AD"}
-    # DXSUM_PDXCONV_ADNIALL has been renamed to DXSUM_PDXCONV
-    # this ensures old ADNI downloads still work with recent versions of Clinica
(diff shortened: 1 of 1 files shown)


AliceJoubert (commented):
In the Excel you made you indicated that you need to change a call from `.Phase` to `.PHASE`. Do you not plan on doing it here?

Adam-Ismaili-92:
> In the Excel you made you indicated that you need to change a call from `.Phase` to `.PHASE`. Do you not plan on doing it here?

After further investigation, no attribute call was made to `.Phase` in the code in regards to this file, and it was not a blocking parameter when testing, so no modification will be made on this matter.

Issue #1623, opened 2026-01-15
[BUG] ADNI-to-BIDS : Empty and undefined columns in sessions specifications added to `*_sessions.tsv`


Adam-Ismaili-92:
**Describe the bug**
When converting an ADNI subset to BIDS using Clinica's ADNI-to-BIDS converter, an error can occur when processing clinical data if no subject IDs match the `bids_ids` list in the function `_filter_subj_bids` from `clinica/converters/adni-to-bids/_utils.py`.
In this case, an empty dataframe is returned with its columns not properly filtered against the sessions specifications in the function `_create_adni_sessions_dict` from `clinica/converters/adni-to-bids/_utils.py`, which can lead to unexpected and invalid columns being added to the final sessions dataframe.

**To Reproduce**
1. Use the ADNI-to-BIDS converter
1. Convert a single subject for which no rows match in some clinical data, such as `003_S_6432` in `clinica_data_ci`
3. Open the produced `*_sessions.tsv` in the BIDS folder
4. Observe the invalid columns present in `*_sessions.tsv`, such as `ADAS_QuestionnaireNotAttempted`, `ADAS_Q1_T1_WL1_W1_butter`, ...

**Expected behavior**
The resulting empty dataframe `df_filtered`, the final sessions dataframe `df_subj_session`, and the produced `*_sessions.tsv` file, should only keep columns defined in the `sessions.tsv` specifications.

**Screenshots**
This kind of data should be found in the produced `*_sessions.tsv` file :
<img width="942" height="70" alt="Image" src="https://github.com/user-attachments/assets/a75dd9ea-cc39-41c3-8959-91b601d559d7" />

**Desktop :**
 - OS: MacOS Sequoia 15.7.1
 - Clinica Version: 0.10.1

**Additional context**
A temporary and quick fix would be to filter `df_filtered` columns to keep only those defined in the `sessions.tsv` specifications before merging them into `df_subj_session`.

PR #1625, opened 2026-01-16
[FIX] ADNI-to-BIDS : Quick fix - Invalid columns removed from `df_filtered` when empty

Adam-Ismaili-92:
Adds a quick fix to #1623, does not close it

### DIFF
--- a/clinica/converters/adni_to_bids/_utils.py
+++ b/clinica/converters/adni_to_bids/_utils.py
@@ -455,10 +455,6 @@ def create_adni_sessions_dict(
                 ]["BIDS CLINICA"].tolist()
             )
 
-            df_filtered = df_filtered.loc[
-                :, (~df_filtered.columns.isin(df_subj_session.columns))
-            ]
-
             df_subj_session = pd.concat([df_subj_session, df_filtered], axis=1)
     if df_subj_session.empty:
         cprint(


clinica/converters/adni_to_bids/_utils.py
```diff
@@ -448,13 +448,19 @@ def create_adni_sessions_dict(
...
             )
+            renamed_cols = set(dict_column_correspondence.values())
+
             df_filtered.rename(columns=dict_column_correspondence, inplace=True)
+            df_filtered = df_filtered.loc[:, df_filtered.columns.isin(renamed_cols)]
+
             df_filtered = df_filtered.loc[
                 :, (~df_filtered.columns.isin(df_subj_session.columns))
             ]
+
```
AliceJoubert:
```suggestion
            df_filtered = pd.DataFrame(columns=df_sessions[df_sessions["ADNI location"]==location]["BIDS CLINICA"].tolist())

```

At this point I guess the goal is to have an empty dataframe with the columns corresponding to the bids names from the specifications, for "ITEM.csv" in this case (location in general)

AliceJoubert (commented):
Thanks for the fix @Adam-Ismaili-92 . Just a suggestion to simplify the code

clinica/converters/adni_to_bids/_utils.py
Adam-Ismaili-92: Yes, it would be simpler and clearer to do it with the suggestion you provided.

clinica/converters/adni_to_bids/_utils.py
```diff
@@ -448,13 +448,17 @@ def create_adni_sessions_dict(
...
+                columns=df_sessions[
+                    df_sessions["ADNI location"].str.split("/").str[0] == location
+                ]["BIDS CLINICA"].tolist()
             )
-            df_filtered.rename(columns=dict_column_correspondence, inplace=True)
+
             df_filtered = df_filtered.loc[
                 :, (~df_filtered.columns.isin(df_subj_session.columns))
             ]
+
```
AliceJoubert: I don't think you need these anymore either

Adam-Ismaili-92: Ok, it has also been removed

PR #1629, opened 2026-01-16
[ENH] ADNI-to-BIDS : Code updated not to handle `TAUQC3.csv` anymore

### LINKED ISSUE #1575: [ENH] ADNI-to-BIDS : Code should be updated to handle `TAUQC_[DATE].csv`

This issue is particularly evident for the file `TAUQC_[DATE].csv`, previously `TAUQC.csv` and `TAUQC3.csv`, which is not handled correctly by the converter’s code.

Therefore, the code responsible for processing `TAUQC_[DATE].csv` should be updated accordingly.

- [x] **MODIF 1** : Adapt the code to only select `TAUQC.csv`, not `TAUQC3.csv`

Adam-Ismaili-92:
Closes #1575

### DIFF
--- a/clinica/converters/adni_to_bids/modality_converters/_tau_pet.py
+++ b/clinica/converters/adni_to_bids/modality_converters/_tau_pet.py
@@ -115,7 +115,9 @@ def _compute_tau_pet_paths(
         if subject_pet_meta.empty:
             continue
         # QC for TAU PET images for ADNI 2 and 3
-        tau_qc_subj = tauqc[(tauqc.SCANQLTY == 1) & (tauqc.RID == int(subject[-4:]))]
+        tau_qc_subj = tauqc[
+            (tauqc.SCANQLTY == 1) & (tauqc.RID == int(subject.split("_S_")[-1]))
+        ]
         tau_qc_subj.rename(columns={"SCANDATE": "EXAMDATE"}, inplace=True)
         subj_dfs_list = get_images_pet(
             subject=subject,


clinica/converters/adni_to_bids/modality_converters/_tau_pet.py
```diff
@@ -107,24 +107,15 @@ def _compute_tau_pet_paths(
...
-        # QC for TAU PET images for ADNI 3
-        tau_qc3_subj = tauqc3[
-            (tauqc3.SCANQLTY == 1) & (tauqc3.RID == int(subject[-4:]))
-        ]
-        # Concatenating visits in both QC files
-        tau_qc_subj = pd.concat(
-            [tau_qc2_subj, tau_qc3_subj], axis=0, ignore_index=True, sort=False
-        )
+        # QC for TAU PET images for ADNI 2 and 3
+        tau_qc_subj = tauqc[(tauqc.SCANQLTY == 1) & (tauqc.RID == int(subject[-4:]))]
```
AliceJoubert: I think we can do a little better than `subject[-4:]` in terms of robustness. At this point, what is subject ? XSXXXX or X_S_XXXX ? Maybe no need to do a regex but you can split the subject string based on S or _ S _ : 
```python
subject.split('_S_') [-1]
```
This takes the RID (as in, the number after the _S_) whatever its size

Adam-Ismaili-92: Thanks for the suggestion, it should have been done.

PR #1634, opened 2026-01-26
[ENH] ADNI-to-BIDS : Regex updated to handle new clinical file names

### LINKED ISSUE #1633: [ENH] ADNI-to-BIDS : Regex should be updated when loading a clinical file

This issue is particularly evident for certain filenames, such as `[OPTIONAL_PREFIX]_All_Images_[DATE]` or `[OPTIONAL_PREFIX]_Manifest_[DATE]`, for which the pattern used to identify them is not suitable.

Therefore, the code responsible for processing clinical file names such as `[OPTIONAL_PREFIX]_All_Images_[DATE]` or `[OPTIONAL_PREFIX]_Manifest_[DATE]` should be updated accordingly.

- [x] **MODIF 1** : Update the function `load_clinical_csv` in `clinica/converters/_utils.py`

Adam-Ismaili-92:
Closes #1633

### DIFF
--- a/test/unittests/converters/adni_to_bids/test_adni_utils.py
+++ b/test/unittests/converters/adni_to_bids/test_adni_utils.py
@@ -267,17 +267,29 @@ def test_load_clinical_csv(
     from clinica.converters._utils import load_clinical_csv  # TODO: move this test
 
     input_df.to_csv(tmp_path / csv_name, index=False)
-    pd.DataFrame().to_csv(tmp_path / f".{csv_name}")
 
     assert_frame_equal(load_clinical_csv(tmp_path, csv_to_look_for), input_df)
 
 
-def test_load_clinical_csv_error(tmp_path):
+@pytest.mark.parametrize(
+    "csv_name,csv_to_look_for",
+    [
+        (".adnimerge.csv", "adnimerge"),
+        (".adnimerge_27Jan2026.csv", "adnimerge"),
+        (".foobar_all_images_27Jan2026.csv", "all_images"),
+    ],
+)
+def test_load_clinical_csv_error(
+    tmp_path, input_df: pd.DataFrame, csv_name: str, csv_to_look_for: str
+):
     import re
 
     from clinica.converters._utils import load_clinical_csv  # TODO: move this test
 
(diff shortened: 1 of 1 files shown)


clinica/converters/_utils.py
```diff
@@ -502,7 +502,9 @@ def load_clinical_csv(clinical_dir: Path, filename: str) -> pd.DataFrame:
     pd.DataFrame:
         Dataframe corresponding to the filename.
     """
-    pattern = filename + r"(_\d{1,2}[A-Za-z]{3}\d{4})?.csv"
+    pattern = (
+        r"^(?!\.)" + r"(?:.*_)?" + filename + r"(?:_\d{1,2}[A-Za-z]{3}\d{4})?\.csv$"
+    )
```
AliceJoubert: Was there a reason for separating the parts of the pattern with a `+` ?

```suggestion
    pattern = (
        r"^(?:[^.]*_)" + filename + r"(?:_\d{1,2}[A-Za-z]{3}\d{4})?\.csv$"
    )
```
--> `?:` for non-capturing group (we are not interested in the value for now)
--> The association `[^.]` indicates takes everything but `.`
--> No `?` since the association I just described is already non-greedy, _ie_ it will take anything (no limited to digits, letters...)

Basically : 
--> `.barba_FILENAME_30Oct2025.csv` ❌
--> `barbaFILENAME_30Oct2025.csv` ❌
--> `barba_truc_01_FILENAME_30Oct2025.csv` ✅

AliceJoubert: Also, do not hesitate to update the unit tests with the new names of the medata files that are very similar. I am thinking of `[...]_ALL_IMAGES_PET_[DATE]` and co.

Adam-Ismaili-92: I agree with your suggestion, I'll add it while correcting the unit test. However, a `?` at the end of `r"^(?:[^.]*_)"`, such as `r"^(?:[^.]*_)?"`, is required ; without it, the first underscore would be mandatory.

AliceJoubert: What underscore are you referring to ?

Adam-Ismaili-92: > What underscore are you referring to ?

The one in `r"^(?:[^.]*_)"`, it would always require filenames to get one underscore at the beginning, without the `?` at the end. This behavior has been tested locally.

PR #1641, opened 2026-01-30
[ENH] GENFI-to-BIDS : More clinical data added to tabular (`.tsv`) files

### LINKED ISSUE #1556: [ENH] GENFI-to-BIDS : More clinical data should be added in the tabular data (`*.tsv`) for subjects and sessions
Few clinical information are available regarding subjects and sessions in the tabular data (`*.tsv`) after the BIDS conversion : 
- Subject level : `participant_id`, `gender`, `handedness`, `education`, `genetic_status_1`, `genetic_status_2`
- Session level : `session_id`, `genfi_version`, `age_at_visit`, `date_of_scan`, `diagnosis`, `ftld-cdr-global`, `cdr-sob`
- Scan level : `modality`, `run_num`, `filename`, `source_path`, `manufacturer`, `number_of_parts`

This lack of information makes these tabular data not really useful, especially for clinicians and researchers.

Thus, more relevant clinical data should be added in these tabular files. 

Some should be mandatory while others should be optional. Here are some suggestions :

- Mandatory clinical data :
  - For participants : 
    - [x] `Blinded Code`
    - [x] `Blinded Site`
    - [x] `Blinded Family`
    - [x] `Gender`
    - [x]  `Genetic Group`
  - For sessions :
    - [x] `Visit`
    - [x] `Date of assessment`
    - [x] `Diagnosis`
    - [x] `Age at visit`
    - [x] `Genetic Status 1`
    - [x] `Genetic Status 2`
    - [x] `Education`
    - [x] `FTLD_CDR_NM_GLOBAL`
- Optional clinical data : 
  - `All remaining clinical data from all clinical files ...`

Adam-Ismaili-92:
Closes #1556

### DIFF
--- a/test/unittests/converters/test_genfi_to_bids_utils.py
+++ b/test/unittests/converters/test_genfi_to_bids_utils.py
@@ -490,28 +490,19 @@ def test_complete_clinical_data():
         df_imaging=df_imaging, df_clinical_list=df_clinical_list
     )
 
-    MERGE_KEY = ["blinded_code", "blinded_site", "visit"]
-
     assert_frame_equal(result, expected, check_like=True, check_dtype=False)
 
 
 @pytest.mark.parametrize(
-    ("full", "gif"),
+    ("full", "gif", "expected"),
     [
-        (False, False),
-        (True, True),
-        (True, False),
-        (False, True),
+        (False, False, "mandatory_specs"),
+        (True, True, "full_specs"),
+        (True, False, "full_specs"),
+        (False, True, "gif_specs"),
     ],
 )
-def test_specs_depending_on_option(full, gif):
+def test_specs_depending_on_option(full, gif, expected):
     from clinica.converters.genfi_to_bids._utils import _specs_depending_on_option
 
-    if full:
-        assert _specs_depending_on_option(full, gif) == "full_specs"
-
-    elif gif:
-        assert _specs_depending_on_option(full, gif) == "gif_specs"
-
-    else:
-        assert _specs_depending_on_option(full, gif) == "mandatory_specs"
+    assert _specs_depending_on_option(full, gif) == expected


clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,82 @@ def _find_clinical_data(
...
-def _read_file(data_file: Path) -> pd.DataFrame:
+def _read_file(data_file_pattern: tuple[Path, str]) -> pd.DataFrame:
+    data_file = data_file_pattern[0]
+    pattern = data_file_pattern[1]
+
+    df_genfi_1 = pd.read_excel(data_file)
+    df_genfi_2 = pd.DataFrame()
+
+    if pattern != "FINAL*GENETICS*.xlsx":
+        df_genfi_2 = pd.read_excel(data_file, sheet_name=1)
```
AliceJoubert: I am unsure you need the `pattern` variable here. If data_file is like `<path_to_file>/file_name` you can do :

```suggestion
    if "GENETICS" not in data_file.name:
        df_genfi_2 = pd.read_excel(data_file, sheet_name=1)
```

That way you also don't need to modify the function `_check_file`

```suggestion
    df_genfi_2 = pd.read_excel(data_file, sheet_name=1)  if pattern != "FINAL*GENETICS*.xlsx" else pd.DataFrame()
```

AliceJoubert: Is there a possibility there will be more than 2 sheets in these clinical data ? (if the study grows)

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,82 @@ def _find_clinical_data(
...
+    Sub-function to outer merge and coalesce duplicates :
+    - Keeps all keys from left_df OR right_df
+    - For overlapping non-key columns, fills left_df n/a values with right_df actual values
+    """
+    left_df = left_df.copy()
+    right_df = right_df.copy()
+
+    # Drops rows without keys
+    left_df = left_df.dropna(subset=on)
+    right_df = right_df.dropna(subset=on)
```
```suggestion
    left_df = left_df.copy().dropna(subset=on)
    right_df = right_df.copy().dropna(subset=on)
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,82 @@ def _find_clinical_data(
...
+    for c in right_cols:
+        dup_col = f"{c}_duplicate"
+        if dup_col in merged_df.columns:
+            if c in merged_df.columns:
+                merged_df[c] = merged_df[c].combine_first(
+                    merged_df[dup_col]
+                )  # Fills n/a in left_df with right_df values
+                merged_df.drop(columns=[dup_col], inplace=True)
+            else:
+                merged_df.rename(columns={dup_col: c}, inplace=True)
```
AliceJoubert: In which cases do you have overlapping columns ?

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,82 @@ def _find_clinical_data(
             "FINAL*CLINICAL*.xlsx",
             "FINAL*BIOSAMPLES*.xlsx",
             "FINAL*NEUROPSYCH*.xlsx",
+            "FINAL*GENETICS*.xlsx",
```
AliceJoubert: For later : you will probably need to update the CI data since this file is not in it

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -248,15 +310,21 @@ def prepare_dataset_to_bids_format(
...
-    specifications = pd.read_csv(Path(__file__).parent / "specifications.csv", sep=";")
-    if not gif:
-        specifications = specifications.head(8)
+
+    spec_filename = "mandatory_specs"
+    if full:
+        spec_filename = "full_specs"
+    specifications = pd.read_csv(
+        Path(__file__).parent / "specifications" / (spec_filename + ".csv"), sep=";"
+    )
```
```suggestion
    specifications = pd.read_csv(
        Path(__file__).parent / "specifications" / f"{'full_specs' if full else 'mandatory_specs'}.csv", sep=";"
    )
```

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: I agree, it can be done that way and be simpler in terms of modif, it should be done in the next commit.

Adam-Ismaili-92: I guess there could be more than two sheets if the study grows, especially with the release of the next phase.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: For example, in the case of the `neuropsych` and `genetics` clinical files that both contain the columns `Genetic Group`, `Genetic Status 1` and `Genetic Status 2`.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Yes, it should be done soon.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,77 @@ def _find_clinical_data(
...
+    left_df = left_df.copy().dropna(subset=on)
+    right_df = right_df.copy().dropna(subset=on)
+
+    # Outer merge dfs
+    merged_df = left_df.merge(
+        right_df,
+        how="outer",
+        on=on,
+        suffixes=("", "_duplicate"),
+        indicator=False,
```
```suggestion
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -163,29 +163,77 @@ def _find_clinical_data(
...
+    for c in right_cols:
+        dup_col = f"{c}_duplicate"
+        if dup_col in merged_df.columns:
+            if c in merged_df.columns:
+                merged_df[c] = merged_df[c].combine_first(
+                    merged_df[dup_col]
+                )  # Fills n/a in left_df with right_df values
+                merged_df.drop(columns=[dup_col], inplace=True)
+            else:
+                merged_df.rename(columns={dup_col: c}, inplace=True)
```
```suggestion
    for col in right_df.columns:
        if dup_col:=f"{col}_duplicate" in merged_df.columns:
                merged_df[col] = merged_df[col].combine_first(
                    merged_df[dup_col]
                )  # Fills n/a in left_df with right_df values
                merged_df.drop(columns=[dup_col], inplace=True)
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -206,28 +254,37 @@ def _complete_clinical_data(
...
-    df_clinical = df_clinical.dropna(subset=merge_key)
-    return df_clinical_complete.merge(df_clinical, how="inner", on=merge_key)
+
+    df_clinical_list = [
+        df_demographics,
+        df_biosamples,
+        df_neuropsych,
+        df_genetics,
+        df_clinical,
+    ]
```
AliceJoubert: I find this list and signature of function kinda unpractical. Here is what I would suggest : 

```python
def _find_clinical_data(
    clinical_data_directory: Path,
    pattern:str,
) -> pd.DataFrame:
    """Finds the clinical data associated with the pattern in the cdd.

    Parameters
    ----------
    clinical_data_directory: Path
        The path to the clinical data.

    pattern : str
         Pattern to find the clinical data

    Returns
    -------
    List[pd.DataFrame]
        Dataframes containing the clinical data
    """
    cprint("Looking for clinical data.", lvl="info")

    return _read_file(_check_file(clinical_data_directory, pattern))
```

that you use as such : 

```python
def parse_clinical_data(clinical_data_directory: Path) -> pd.DataFrame:
    return _complete_clinical_data(_find_clinical_data(clinical_data_directory, "FINAL*IMAGING*.xlsx"),
     [_find_clinical_data(clinical_data_directory, pattern) for pattern in ("FINAL*CLINICAL*.xlsx", "FINAL*DEMOGRAPHICS*.xlsx", "FINAL*BIOSAMPLES*.xlsx", "FINAL*NEUROPSYCH*.xlsx", "FINAL*GENETICS*.xlsx")])
```

so you can simplify the function in : 
```python
def _complete_clinical_data(
    df_imaging: pd.DataFrame,
    metadata_df_list:list[pd.DataFrame],
) -> pd.DataFrame:
     #TODO : replace list below
```

.pre-commit-config.yaml
```diff
@@ -13,6 +13,7 @@ repos:
     rev: v2.3.0
     hooks:
       - id: codespell
+        args: ["--ignore-words-list=ALS"]
```
AliceJoubert, line 16: Pourquoi est-ce qu'on a besoin de ça ?

Adam-Ismaili-92: Parce que le string "ALS", qui est utilisé comme valeur de test dans la fonction `test_merge_and_coalesce` par exemple, est considéré comme une erreur d'orthographe par le correcteur du pre-commit, donc je lui ai notifié d'ignorer le string.

D'ailleurs, si aucune autre modif n'est à signaler, j'attends la réponse d'Arabella avant de merge.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -135,99 +135,137 @@ def _check_file(directory: Path, pattern: str) -> Path:
...
+
+    # Coalesces overlapping columns
+    for col in right_df.columns:
+        if (dup_col := f"{col}_duplicate") in merged_df.columns:
+            merged_df[col] = merged_df[col].combine_first(
+                merged_df[dup_col]
+            )  # Fills n/a in left_df with right_df values
+            merged_df.drop(columns=[dup_col], inplace=True)
+        else:
+            merged_df.rename(columns={dup_col: col}, inplace=True)
```
```suggestion
```

You still don't need this : you won't be able to rename something in merged_df if it is not inside.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -135,99 +135,137 @@ def _check_file(directory: Path, pattern: str) -> Path:
...
-        df_biosamples, how="inner", on=merge_key
-    )
-    df_clinical_complete = df_clinical_complete.merge(
-        df_neuropsych, how="inner", on=merge_key
-    )
-    df_clinical = df_clinical.dropna(subset=merge_key)
-    return df_clinical_complete.merge(df_clinical, how="inner", on=merge_key)
+
+    df_clinical_complete = df_imaging.copy()
+
```
AliceJoubert, line 255: Last question here : why was it necessary before to remove the "diagnosis" column and why do we not do it anymore ?

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
     assert not str(row["source_path"]).startswith(str(source))
     assert row["manufacturer"] == "SIEMENS"
     assert row["number_of_parts"] == 1
+
+
+def test_find_clinical_data(monkeypatch, tmp_path):
+    import clinica.converters.genfi_to_bids._utils as genfi_utils
```
AliceJoubert: There you can fully make use of parametrize : 

```python
@pytest.mark.parametrize("file_name, expected_name",
[(filename1, expectedname1),
 (filename2,expectedname2), ...])
def test_truc(monkeypatch, tmp_path, file_name, expected_name):
     file_path = tmp_path / expected_name
     def _check_file(clinical_data_directory, pattern):
        return file_path
     ....
     out = genfi_utils._find_clinical_data(tmp_path, file_name)
     assert out[i].iloc[0, 0] == expected_name
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+    for i in range(length):
+        out.append(genfi_utils._find_clinical_data(tmp_path, clinical_file_names[i]))
+
+    assert len(out) == 6
+
+    for i in range(len(out)):
+        assert out[i].iloc[0, 0].endswith(clinical_file_names[i] + ".xlsx")
+
+
+@pytest.mark.parametrize(
```
AliceJoubert: There you don't need parametrize since you have only one value per variable each time.

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+        )
+    ],
+)
+def test_merge_and_coalesce(left_df, right_df, on, expected_df):
+    from clinica.converters.genfi_to_bids._utils import _merge_and_coalesce
+
+    result = _merge_and_coalesce(left_df, right_df, on=on)
+
+    # sort and normalize to properly compare
+    result = result.sort_values(by=on).reset_index(drop=True)
```
AliceJoubert: This could be done at the end of `_merge_and_coalesce` for more readability

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+    expected_df = expected_df.sort_values(by=on).reset_index(drop=True)
+    result = result[expected_df.columns]
```
AliceJoubert: Why do we need this ?

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+
+    assert_frame_equal(result, expected_df, check_dtype=False)
+
+    assert not any(col.endswith("_duplicate") for col in result.columns)
```
AliceJoubert: This can happen if naturally there is a column with "duplicate" in its name in the first dataframe. We don't need to remove these columns I guess. Plus, with the previous `assert_frame_equal` you don't need to check anything more on result vs expected

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+
+
+@pytest.mark.parametrize(
+    "df_imaging,df_clinical_list,expected",
```
AliceJoubert: Same you don't need parametrize here

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,176 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+    MERGE_KEY = ["blinded_code", "blinded_site", "visit"]
+
+    # sort and normalize to properly compare
+    result = result.sort_values(by=MERGE_KEY).reset_index(drop=True)
+    expected = expected.sort_values(by=MERGE_KEY).reset_index(drop=True)
+    result = result[expected.columns]
+
+    assert_frame_equal(result, expected, check_dtype=False)
+
+    assert not any(col.endswith("_duplicate") for col in result.columns)
```
AliceJoubert: Same comments as the previous function + with the data you use to test you don't introduce duplicates

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: It was removed to avoid duplicates since the column is present in both `df_clinical` and `df_demographics`. Since we now merge and coalesce, that logic is not needed anymore.

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: Those lines ensure the comparison checks for data correctness, not row order, index artifacts or column ordering differences. For example, it could fail on this kind of comparison without sorting and normalizing : 
```id | value            id | value```
```2  | B       vs       1  | A```
```1  | A                2  | B```

File: test/unittests/converters/test_genfi_to_bids_utils.py (hunk 14 above)
Comment (Adam-Ismaili-92) [reply test/unittests/converters/test_genfi_to_bids_utils.py]: Those lines of code are applied to both `result` and `expected` dataframes to properly compare (the data from) them, they're not really linked to `_merge_and_coalesce` functioning and they would still be used for the `expected` df, so I don't really see the point of moving them out.

Comment (AliceJoubert) [reply test/unittests/converters/test_genfi_to_bids_utils.py]: My point is : 
- if you do care about the order of columns / index you should put it at the end of merge_and_coalesce
- if you don't care about it you don't need to go out of your way to test it

File: test/unittests/converters/test_genfi_to_bids_utils.py (hunk 15 above)
Comment (AliceJoubert) [reply test/unittests/converters/test_genfi_to_bids_utils.py]: If you had additional unexpected columns in `result` you would miss them which is not ideal. You can look up the `check_like` flag of `assert_df_equal` to test without ordered index/columns

File: clinica/converters/genfi_to_bids/_utils.py [hunk 19]
```diff
@@ -135,99 +135,135 @@ def _check_file(directory: Path, pattern: str) -> Path:
...
 
 
+def _merge_and_coalesce(
+    left_df: pd.DataFrame,
+    right_df: pd.DataFrame,
+    on: List[str],
+) -> pd.DataFrame:
+    """
+    Sub-function to outer merge and coalesce duplicates :
+    - Keeps all keys from left_df OR right_df
```
```suggestion
    - Keeps all keys from both left_df and right_df
```

or something similar would be clearer

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,155 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
     assert row["manufacturer"] == "SIEMENS"
     assert row["number_of_parts"] == 1
+
+
+@pytest.mark.parametrize(
+    "filename",
+    [
+        "FINAL_IMAGING.xlsx",
+        "FINAL_DEMOGRAPHICS.xlsx",
+        "FINAL_CLINICAL.xslx" "FINAL_BIOSAMPLES.xslx",
```
```suggestion
        "FINAL_CLINICAL.xslx",
        "FINAL_BIOSAMPLES.xslx",
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,155 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+
+    def _read_file(data_file):
+        return pd.DataFrame({"source": [data_file.name]})
+
+    monkeypatch.setattr(genfi_utils, "_check_file", _check_file)
+    monkeypatch.setattr(genfi_utils, "_read_file", _read_file)
+
+    out = genfi_utils._find_clinical_data(tmp_path, filename)
+
+    assert out.iloc[0, 0].endswith(filename)
```
AliceJoubert, line 379:
```suggestion
    assert len(out) == 1
    assert out.iloc[0, 0].endswith(filename)
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,155 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+            }
+        ),
+        pd.DataFrame(
+            {
+                "blinded_code": ["C9ORF001", "C9ORF002"],
+                "blinded_site": ["GENFI_AA", "GENFI_AA"],
+                "visit": [1, 1],
+                "diagnosis": ["bvFTD", "ALS"],
+            }
+        ),
```
AliceJoubert, line 447: I think it would be interesting to add the case of 2 data frames in the list with overlapping column to verify the behavior works with more than 1. Ex : col "diagnosis_status" in imaging and 2 other df in the list.

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: Ok, I see better now, since it is not necessary, I'll not go out of my way and test it, I'll use `assert_df_equal` flag instead.

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: I agree with this suggestion. It should be done in the next commit.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -135,99 +135,136 @@ def _check_file(directory: Path, pattern: str) -> Path:
...
+            df_clinical_complete, df, on=merge_key
+        )
+
+    return df_clinical_complete
 
 
 def prepare_dataset_to_bids_format(
     complete_data_df: pd.DataFrame,
     gif: bool,
+    full: bool,
```
AliceJoubert: You can indicate the default (False)

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -248,15 +285,21 @@ def prepare_dataset_to_bids_format(
...
     )
-    specifications = pd.read_csv(Path(__file__).parent / "specifications.csv", sep=";")
-    if not gif:
-        specifications = specifications.head(8)
+
+    specifications = pd.read_csv(
+        Path(__file__).parent
+        / "specifications"
+        / f"{'full_specs' if full else ('gif_specs' if gif else 'mandatory_specs')}.csv",
+        sep=";",
```
AliceJoubert, line 328: I think at this point we need a more complex case check to make sure it behaves as intended. 

You can write a function like:

```python
def  _spec_depending_on_option(full:bool, gif:bool)->str:
     #TODO : unit test
     if full:
          return "full_specs"
     if gif:
          return "gif_specs"
     return "mandatary_specs"

# Then you use it : 
specifications = pd.read_csv(
        Path(__file__).parent
        / "specifications"
        / f"{_spec_depending_on_option(full, gif)}.csv",
        sep=";",...)
```

if I'm not wrong, returns "full_specs" as long as full is True which would be the behavior we want. WDYT ?

test/unittests/converters/test_genfi_to_bids_utils.py
AliceJoubert: It would be better to use different values "n/a", 2... to make sure it actually works (though it probably does but it is for future cases)

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: Plus unless I am wrong this function is not unit tested so it's best if not many functionalities are handled by it (--> better to write helper functions which are tested)

docs/Converters/GENFItoBIDS.md
```diff
@@ -95,9 +95,15 @@ where:
...
+    - `--clinical-data-dir/-cdd` is the path to the clinical data directory. Allows the user to add mandatory clinical data to `participants.tsv` and `sessions.tsv`. Mandatory clinical data are the following :
+        - For `participants.tsv` : `blinded_code`, `blinded_family`, `blinded_site`, `gender`.
+        - For `sessions.tsv` : `age_at_visit`, `date_of_assessment`, `diagnosis`, `education`, `ftld-cdr-nm-global`, `genetic_group`, `genetic_status_1`, `genetic_status_2`, `visit`. 
+        <br> All the remaining clinical data is optional and added through the use of the `-full` flag.
     - `--clinical-data-tsv/-cdt` is a tsv file containing the additional fields the user wants. The available data can be retrieved inside the installation of clincia at this location `your_path_to_clinica/clinica/iotools/data/genfi_data.tsv`.
-    - `-gif` Allows the user to add the value from gif to the session.tsv.
+    - `-gif` allows the user to add all the clinical data related to the imaging volumes (GIF, Geodesic Information Flow) to `session.tsv`.
+    - `-full` allows the user to add all clinical data (mandatory and optional) to `sessions.tsv`.
```
AliceJoubert: I would precise that gif also contains mandatory clinical data. And full contains gif data.

AliceJoubert (commented):
Thanks for the work @Adam-Ismaili-92 . Some comments about the new additions. Also I am unsure what you did in the CI data PR because I did not review it yet but it would be nice to have these options tested in non-regression (one or the other)

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: I agree with your suggestion. It should be done and unit tested in the next commit.

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,172 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+            "ds_f_score": [8, 11],
+            "genetic_status_1": ["P", "A"],
+        }
+    )
+
+    result = _complete_clinical_data(
+        df_imaging=df_imaging, df_clinical_list=df_clinical_list
+    )
+
+    MERGE_KEY = ["blinded_code", "blinded_site", "visit"]
```
```suggestion
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -346,3 +346,172 @@ def test_write_scans_and_niftis(tmp_path, mocker):
...
+    from clinica.converters.genfi_to_bids._utils import _specs_depending_on_option
+
+    if full:
+        assert _specs_depending_on_option(full, gif) == "full_specs"
+
+    elif gif:
+        assert _specs_depending_on_option(full, gif) == "gif_specs"
+
+    else:
+        assert _specs_depending_on_option(full, gif) == "mandatory_specs"
```
```suggestion
@pytest.mark.parametrize(
    ("full", "gif", "expected"),
    [
        (False, False, "mandatory_specs"),
        (True, True, "full_specs"),
        (True, False, "full_specs"),
        (False, True, "gif_specs"),
    ],
)
def test_specs_depending_on_option(full, gif, expected):
    from clinica.converters.genfi_to_bids._utils import _specs_depending_on_option

    assert _specs_depending_on_option(full, gif) == expected
```

AliceJoubert: Also because if your test function copies the behavior of the function (with `if, else...`) you do not really test if the behavior is correct

Adam-Ismaili-92:
> Almost perfect, thanks for the work. Next step is CI data PR and if it validates the non-regression tests with this PR I will merge both

[`clinica_data_ci` PR](https://github.com/aramis-lab/clinica_data_ci/pull/107), linked to this PR, has been pushed and validates the non-regression tests.

PR #1647, opened 2026-02-11
[FIX] GENFI-to-BIDS : Fix and simplification for `-cdt` option

### LINKED ISSUE #1646: [FIX] GENFI-to-BIDS : `-cdt` option doesn't work properly
**Describe the bug**
`-cdt` option doesn't work properly. This error is thrown when using it : 
`NotADirectoryError: [Errno 20] Not a directory: '/clinica/clinica/converters/genfi_to_bids/_utils.py/data_mapping.tsv'`
Additionally, the use of a `data_mapping.tsv` file different from the specifications for this option is not practical.

**To Reproduce**
1. Use Clinica's GENFI-to-BIDS converter by entering this command : `clinica convert genfi-to-bids [DATASET_DIR] [BIDS_DIR] -cdd [CLINICAL_DATA_DIR] -cdt [CDT_FILE]`
2. See the error

**Expected behavior**
`-cdt` usage should not throw any error. Also, the use of `data_mapping.tsv` should be removed to only use the specifications.

**Screenshots**
Image

**Desktop**
 - OS: MacOS Sequoia 15.7.1
 - Clinica's Version: 0.10.1

**Additional context**
A simple fix would be to add a `.parent` after the `Path(__file__)` in the line `data_mapping = pd.read_csv(Path(__file__) / "data_mapping.tsv", sep="\t")` in the file `clinica/clinica/converters/genfi_to_bids/_utils.py`.

Adam-Ismaili-92:
Closes #1646 
This PR is not complete, further modifications need the merge of PR #1641

### DIFF
--- a/test/nonregression/test_run_converters.py
+++ b/test/nonregression/test_run_converters.py
@@ -66,7 +66,7 @@ def test_genfi_converter_using_options(cmdopt, tmp_path, gif, full):
     output_dir = tmp_path / "bids"
     convert(
         study,
-        path_to_dataset=input_dir / "unorganized",
+        path_to_dataset=input_dir / "unorganized_for_options",
         bids_dir=output_dir,
         path_to_clinical=input_dir / "clinical_data",
         subjects=input_dir / "subjects.txt",


AliceJoubert:
> * I was rather thinking that `-cdt` could still be used independently from the other options, requiring only the list of data located in the `full_specs.csv` file. Obviously, any duplicates already provided by options such as `-gif` or `-full` will be removed.

Yes, `cdt` should be kept and can be used independently from the other options. However in the case the flag `full` is set to True then whatever is indicated in the option cdt becomes useless because `full` takes everything already.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,93 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+        for i, line in enumerate(f, start=1):
+            data = line.strip()
+
+            if not data:
+                continue  # Skip empty lines
+
+            if data not in specs_values:
+                raise ValueError(
+                    f"Error at line {i}: '{data}' not found in specifications."
+                )
```
AliceJoubert: That is a bit excessive to stop the whole conversion for one mistake ; here you can `cprint` a warning to the user with the same message but `continue`

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,93 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+            clinical_data_list.append(data)
+
+    if not clinical_data_list:
+        raise ValueError("'-clinical_data_txt/cdt' is empty (no valid entries found).")
```
AliceJoubert: Same here you don't really need to stop everything. A warning and the next function also needs to work for an empty list

docs/Converters/GENFItoBIDS.md
```diff
@@ -99,7 +99,7 @@ where:
         - For `participants.tsv` : `blinded_code`, `blinded_family`, `blinded_site`, `gender`.
         - For `sessions.tsv` : `age_at_visit`, `date_of_assessment`, `diagnosis`, `education`, `ftld-cdr-nm-global`, `genetic_group`, `genetic_status_1`, `genetic_status_2`, `visit`. 
         <br> All the remaining clinical data is optional and added through the use of the `-full` flag.
-    - `--clinical-data-tsv/-cdt` is a tsv file containing the additional fields the user wants. The available data can be retrieved inside the installation of clincia at this location `your_path_to_clinica/clinica/iotools/data/genfi_data.tsv`.
+    - `--clinical-data-txt/-cdt` is a txt file containing the additional fields the user wants, written one field per line. The available data can be retrieved inside the installation of clinica at this location : `your_path_to_clinica/clinica/converters/genfi_to_bids/specifications/full_specs.csv`.
```
AliceJoubert: It would be nice to give an example of what it looks like right after the txt explanation

Adam-Ismaili-92:
> Looking at the specifications files (full_specs VS mandatory_specs) it seems like the only place where it is possible to add values is in `sessions`. If true then you don't need to go to all the trouble of looping on specs columns and the whole dataframe. You only need to :
> 
> * check if the value exists (ie is inside full_specs sessions column)
> * check if it is not inside the used specifications file sessions column
> * add to this column if not
> 
> That should simplify the code

The vast majority of the added fields are indeed related to the `sessions` column of the specifications, but given that they were assigned on an indicative basis, without any real review from Arabella, I prefer to keep this logic in case she wishes to revisit certain assignments, by adding some to the other columns.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,98 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    List[str]
+        List of selected clinical data fields
+    """
+    clinical_data_list = []
+
+    specs_values = {
+        str(value).strip()
+        for value in specs_df.to_numpy().ravel()
+        if pd.notna(value) and str(value).strip() != ""
+    }
```
AliceJoubert: I am not sure why you need to : 
- specify it's strings
- strip the result
- check if you have empty values
since the specs_df is always going to be the specifications for full, which has been written by you. Normally it should already be under the right format. 

Also, you don't need to consider all the dataframe, the "sessions" part is enough

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,98 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+            if not data:
+                continue  # Skip empty lines
+
+            if data not in specs_values:
+                log_and_warn(
+                    f"Line {i}: '{data}' not found in specifications. It will be ignored.",
+                    UserWarning,
+                )
+                continue
```
AliceJoubert: When you can it is good to write small helper functions to avoid having several functionalities handled by the same function. It also makes it easier to test. An exemple here would be to have the "filtering" part in another function, called here. That way you separate the functionality "loading the text and the specs" and "filtering"

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,98 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+            clinical_data_list.append(data)
+
+    if not clinical_data_list:
+        log_and_warn(
+            "'-clinical_data_txt/cdt' is empty (no valid entries found).", UserWarning
+        )
```
AliceJoubert, line 327: You can be more precise : 
```suggestion
        log_and_warn(
            f"File for option -cdt at location {cdt_path} does not contain any valid entry.", UserWarning
        )
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,98 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+        Dataframe to complete
+    """
+    for value in clinical_data_list:
+        sessions_values = df_to_complete["sessions"].astype(str).tolist()
+
+        if value in sessions_values:
+            continue
+
+        if value in specs_df["sessions"].values:
+            last_valid_idx = df_to_complete["sessions"].last_valid_index()
```
AliceJoubert: Here `specs_df` is always going to be the specifications for full. I don't understand why you still need to refer to it after you filtered the list already.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,98 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+            next_idx = last_valid_idx + 1
+
+            if next_idx < len(df_to_complete):
+                df_to_complete.loc[next_idx, "sessions"] = value
```
AliceJoubert: Also I am unsure why you need to refer to a last "valid" index specifically and compare the length to it

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: - Specifying that the value is a string and stripping the result is just a form of good practice to make a general line of code that would give results of a single type, for easier filtering processing, and be cleaned of white spaces, in case the specifications would change or the function would be used for other purposes. But it is not necessary. I can remove it in the next commit.

- Checking if there are empty or NaN values was just there to assure that no empty strings or NaN values would be put inside the resulting list. I can however limit my checking to NaN values if I remove the conversion to `str`, which should be done in the next commit.

- All the filtering part, including the ones done on the `sessions` column from the function `_merge_clinical_data_list_into_df`, will be moved into helper functions used in this function in the next commit.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Because the last valid index can refer to the last real value before a series of NaN, not necessarily the end of the column.

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: I do understand the interest of checking these but not really if the input is a file that you wrote yourself. That would be different for a file given as input by a user over which you have no control originally

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,129 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
     return "mandatory_specs"
 
 
+def _filter_invalid_data(
+    data: str, index: int, specs_values: set, specs_df: pd.DataFrame
+) -> bool:
+    """Filter invalid (empty, unknown and not present in 'sessions') data from the user txt file.
```
AliceJoubert: Taking a look at what comes after it looks like it would be easier to return True in case the line is validated, False otherwise. To reflect that change on the na:e of the function you could name it `_validate_data_from_user` or something similar. WDYT ?

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,129 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    if data not in specs_values:
+        log_and_warn(
+            f"Line {index}: '{data}' not found in specifications. It will be ignored.",
+            UserWarning,
+        )
+        return True
+
+    # Skip lines not in 'sessions'
+    if data not in specs_df["sessions"].values:
+        return True
```
AliceJoubert: Is there any advantage to checking both specs_df and specs_values ? To be fair I would have expected only the first one (specs_df) since you do not need to check if that value exist in participants or scans anymore

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,129 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    specs_values = {value for value in full_specs.to_numpy().ravel() if pd.notna(value)}
+
+    with open(cdt_path, "r", encoding="utf-8") as f:
+        for i, line in enumerate(f, start=1):
+            data = line.strip()
+
+            if _filter_invalid_data(data, i, specs_values, full_specs):
+                continue
+
+            clinical_data_list.append(data)
```
AliceJoubert: You could : 
```python
if _filter... : #(function that returns True)
    clinical_data_list.append(data)
```

or 

```python
if not _filter... : #(function that returns False)
    clinical_data_list.append(data)
```

No need to make a special case for the one where you do nothing

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,129 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+        List of selected clinical data fields
+    """
+    clinical_data_list = []
+
+    full_specs = pd.read_csv(
+        Path(__file__).parent / "specifications/full_specs.csv",
+        sep=";",
+    )
+
+    specs_values = {value for value in full_specs.to_numpy().ravel() if pd.notna(value)}
```
AliceJoubert: Based on my comments about the filtering function you could go with : 

```python
specs_values = fulls_specs["sessions"].to_numpy()
```

or 

```python
specs_values = fulls_specs["sessions"].dropna().to_numpy()
```

if you want. WDYT ?

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,129 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    df_to_complete: pd.DataFrame
+        Specs like dataframe to complete
+
+    Returns
+    -------
+    pd.DataFrame
+        Dataframe to complete
+    """
+    for value in clinical_data_list:
+        sessions_values = df_to_complete["sessions"].astype(str).tolist()
```
AliceJoubert: you can get this out of the loop

clinica/converters/genfi_to_bids/_utils.py
AliceJoubert: Since we are talking about incrementing the values inside df_to_complete["sessions"] which refers to a file you created yourself and which refers to the column where there should be no nans if I'm not wrong I'm not sure you need to keep this. You can probably just take the end of the column. WDYT ?

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: The question is not really whether I am the original writer of the file or not, but to make the code durable over time, given that the specifications will surely have to be changed along with changes in GENFI phases, Arabella's interests, etc. This would allow to avoid encountering the same issue multiple times.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: As mentioned earlier, it will make the code a bit too specific and not really durable in time, but given the current files processed, I can indeed do it this way. This should be done in the next commit.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Yes I agree. It should be done this way in the next commit.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,108 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
     return "mandatory_specs"
 
 
+def _validate_data_from_user(
+    data: str, index: int, full_specs_values: List[str]
```
```suggestion
    data: str, index: int, full_specs_values: list[str]
```

You can use native typing here. I know it is not your doing but since we can now use `list` instead of `List` from `Typing` due to the python versions having evolved we can slowly change it in the base code of clinica as we go.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,108 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    # Skip empty lines
+    if not data:
+        return False
+
+    # Skip unknown lines in full_specs.sessions
+    if data not in full_specs_values:
+        log_and_warn(
+            f"Line {index}: '{data}' not found in 'full_specs[\"sessions\"]'. It will be ignored.",
+            UserWarning,
+        )
```
AliceJoubert: In terms of user experience it might be possible that it creates a lot of lines. I would suggest possibly storing the index/data in a list and format it at the end of the loop if need be to print to the user all the information at the same time.
That would mean outside of this function keep the index and data in a list if the result of this function is False. And I think in that case you don't need to pass index as a parameter anymore here. 
WDYT ?

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,108 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+        return False
+
+    return True
+
+
+def _load_clinical_data_list(cdt_path: Path) -> List[str]:
```
```suggestion
def _load_clinical_data_list(cdt_path: Path) -> list[str]:
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,108 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+        log_and_warn(
+            f"File for option '-clinical_data_txt/cdt' at location {cdt_path} does not contain any valid entry.",
+            UserWarning,
+        )
+
+    return clinical_data_list
+
+
+def _merge_clinical_data_list_into_df(
+    clinical_data_list: List[str], df_to_complete: pd.DataFrame
```
```suggestion
    clinical_data_list: list[str], df_to_complete: pd.DataFrame
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +286,108 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+    Returns
+    -------
+    pd.DataFrame
+        Dataframe to complete
+    """
+    sessions_values = df_to_complete["sessions"].astype(str).tolist()
+
+    for value in clinical_data_list:
+        if value not in sessions_values:
```
```suggestion
    for value in clinical_data_list:
        if value not in df_to_complete["sessions"]:
```

Correct me if I'm wrong but I think you can go directly with this

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -328,23 +427,22 @@ def prepare_dataset_to_bids_format(
...
+                "The '-full' flag is being used, "
+                "using the '-clinical_data_txt/-cdt' option is redundant and will be ignored.",
+                UserWarning,
+            )
+
+        else:
+            specifications = _merge_clinical_data_list_into_df(
+                _load_clinical_data_list(path_to_clinical_txt),
+                specifications.copy(),
+            )
```
AliceJoubert, line 424: Just out of curiosity from my part, is there any reason to copy the spec file if anyway you modify it and it is not used anywhere else ?

docs/Converters/GENFItoBIDS.md
```diff
@@ -99,7 +99,23 @@ where:
         - For `participants.tsv` : `blinded_code`, `blinded_family`, `blinded_site`, `gender`.
         - For `sessions.tsv` : `age_at_visit`, `date_of_assessment`, `diagnosis`, `education`, `ftld-cdr-nm-global`, `genetic_group`, `genetic_status_1`, `genetic_status_2`, `visit`. 
         <br> All the remaining clinical data is optional and added through the use of the `-full` flag.
-    - `--clinical-data-tsv/-cdt` is a tsv file containing the additional fields the user wants. The available data can be retrieved inside the installation of clincia at this location `your_path_to_clinica/clinica/iotools/data/genfi_data.tsv`.
+    - `--clinical-data-txt/-cdt` is a txt file containing the additional fields the user wants. The available data can be retrieved inside the installation of clinica at this location : `your_path_to_clinica/clinica/converters/genfi_to_bids/specifications/full_specs.csv`. This file should be written one field per line such as in the example below :
```
AliceJoubert:
```suggestion
    - `--clinical-data-txt/-cdt` is a txt file containing the additional fields the user wants. The available data can be retrieved inside the installation of clinica at this location : `your_path_to_clinica/clinica/converters/genfi_to_bids/specifications/full_specs.csv`. The txt file should be written one field per line such as in the example below :
```

Also maybe I would precise that the added field will be found in the sessions . tsv files in the BIDS. And that the option is useless in case of full being used (maybe not formulated that way).

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+
+def test_invalidate_unknown_data_from_user():
+    from clinica.converters.genfi_to_bids._utils import _validate_data_from_user
+
+    data = "invalid_data_0"
+
+    expected = f"Line {0}: '{data}' not found in 'full_specs[\"sessions\"]'. It will be ignored."
+
+    with pytest.warns(UserWarning, match=re.escape(expected)):
+        assert not _validate_data_from_user(data, 0, FULL_SPECS_VALUES)
```
AliceJoubert: I feel like `test_invalidate_unknown_data_from_user` and `test_invalidate_not_in_sessions_data_from_user` test for the same behavior and are redundant

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+@pytest.mark.parametrize(
+    ("clinical_data_list", "expected"),
+    [
+        (
+            ["blinded_family\n", "aad_1\n", "bids_filename"],
+            ["aad_1"],
+        ),
+        (
+            ["  blinded_site  \n", "\n", "aad\n", "\tbids_full_path\n"],
+            ["aad"],
```
AliceJoubert: You are kinda testing the same thing between these 2 cases :  
- the validated one is very similar in both cases as they present the same oddity ie `\n`.
- the format of the invalidated ones vary but you don't really care here since they are not in the SPECS FILES and would never get validated anyway (ex blinded_site)

Moreover you don't need to parametrize here, if you want to test several cases around a case (different oddities like whitespaces and \n or presence in the files or not) you can just have 1 list containing all these cases. Not 2 lists.

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+)
+def test_load_clinical_data_list_success(tmp_path, clinical_data_list, expected):
+    from clinica.converters.genfi_to_bids._utils import _load_clinical_data_list
+
+    cdt_path = tmp_path / "additional_clinical_data.txt"
+    cdt_path.write_text("".join(clinical_data_list), encoding="utf-8")
+
+    out = _load_clinical_data_list(cdt_path)
+
+    assert out == expected
```
```suggestion
    assert _load_clinical_data_list(cdt_path) == expected
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+
+
+def test_merge_clinical_data_list_into_df_no_duplicate():
+    from clinica.converters.genfi_to_bids._utils import (
+        _merge_clinical_data_list_into_df,
+    )
+
+    clinical_data_list = ["session_id", "genfi_version"]
+
+    out = _merge_clinical_data_list_into_df(
```
AliceJoubert, line 631: These 2 can be only one test since you test the exact same thing at the end. And if you do so I think you can define `TO_COMPLETE_SPECS_DF` inside the unique function testing it

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+    cdt_path = tmp_path / "additional_clinical_data.txt"
+    cdt_path.write_text("".join(clinical_data_list), encoding="utf-8")
+
+    expected = (
+        f"File for option '-clinical_data_txt/cdt' at location {cdt_path} "
+        "does not contain any valid entry."
+    )
+
+    with pytest.warns(UserWarning, match=re.escape(expected)):
+        _load_clinical_data_list(cdt_path)
```
AliceJoubert: In that case I think it's interesting to test that the output of _load_clinical_data_list is indeed empty. Else you are just testing the behavior of the intermediary function that is used inside. You can go with : 



```suggestion
    with pytest.warns(UserWarning, match=re.escape(expected)):
       assert not  _load_clinical_data_list(cdt_path)
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,213 @@ def test_specs_depending_on_option(full, gif, expected):
...
+def test_load_clinical_data_list_unknown_field(tmp_path):
+    from clinica.converters.genfi_to_bids._utils import _load_clinical_data_list
+
+    cdt_path = tmp_path / "additional_clinical_data.txt"
+    cdt_path.write_text("aad\nfalse_field\naad_1\n", encoding="utf-8")
+
+    expected = "Line 2: 'false_field' not found in 'full_specs[\"sessions\"]'. It will be ignored."
+
+    with pytest.warns(UserWarning, match=re.escape(expected)):
+        _load_clinical_data_list(cdt_path)
```
AliceJoubert: Same here you can test the output of the function to verify only one field was ignored

test/unittests/converters/test_genfi_to_bids_utils.py
Adam-Ismaili-92: I still would like to be sure that they are getting invalidated, that is way it is still interesting to put them, even if I agree that I can get everything into one list instead of using parametrize.

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Just for extra safety reasons since the dataframe is modified in place in the function, but it can be done without it.

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +287,109 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    Returns
+    -------
+    bool
+        Filtering result. False if the data line should be skipped. True otherwise.
+    """
+    # Skip unknown lines in full_specs.sessions
+    if data not in full_specs_values:
+        return False
+
+    return True
```
AliceJoubert: I think you can simplify as : 
```suggestion
    return data in full_specs_values
```

AliceJoubert: I am unsure it's necessary to keep the function if it is only a simple check though (before you also had the if data test)

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +287,109 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+    full_specs_values = pd.read_csv(
+        Path(__file__).parent / "specifications/full_specs.csv",
+        sep=";",
+    )["sessions"].tolist()
+
+    with open(cdt_path, "r", encoding="utf-8") as f:
+        for index, line in enumerate(f, start=1):
+            data = line.strip()
+
+            if data:
```
```suggestion
            if data:=line.strip():
```

clinica/converters/genfi_to_bids/_utils.py
```diff
@@ -286,9 +287,109 @@ def _specs_depending_on_option(full: bool, gif: bool) -> str:
...
+
+    if not clinical_data_list:
+        log_and_warn(
+            f"File for option '-clinical_data_txt/cdt' at location {cdt_path} does not contain any valid entry.",
+            UserWarning,
+        )
+
+    if index_data_list:
+        message = (
+            "Some data have not been found in 'full_specs[\"sessions\"]':\n"
```
AliceJoubert: It's unclear for the user which data you are talking about so I would precise we are talking about the text file from -cdt (like in the previous message). You can also indicate that every available option for data is written in full_specs sessions

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,162 @@ def test_specs_depending_on_option(full, gif, expected):
...
+
+FULL_SPECS_VALUES = FULL_SPECS_DF["sessions"].tolist()
+
+
+def test_invalidate_unknown_data_from_user():
+    from clinica.converters.genfi_to_bids._utils import _validate_data_from_user
+
+    data = "invalid_data_0"
+
+    assert not _validate_data_from_user(data, FULL_SPECS_VALUES)
```
```suggestion
    assert not _validate_data_from_user("invalid_data_0, FULL_SPECS_VALUES)
```

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,162 @@ def test_specs_depending_on_option(full, gif, expected):
...
+        "source",
+        "blinded_code",
+        "bids_filename",
+        "bids_full_path",
+    ],
+)
+def test_invalidate_not_in_sessions_data_from_user(data):
+    from clinica.converters.genfi_to_bids._utils import _validate_data_from_user
+
+    assert not _validate_data_from_user(data, FULL_SPECS_VALUES)
```
AliceJoubert: Still this function is testing the exact same thing as the previous one with "invalid_data_0"

test/unittests/converters/test_genfi_to_bids_utils.py
```diff
@@ -506,3 +507,162 @@ def test_specs_depending_on_option(full, gif, expected):
...
+                "genfi_version",
+            ],
+        }
+    )
+
+    clinical_data_list = ["aad", "session_id", "aad_1", "aad_2", "genfi_version"]
+
+    out = _merge_clinical_data_list_into_df(
+        clinical_data_list, TO_COMPLETE_SPECS_DF.copy()
+    )
```
```suggestion
    out = _merge_clinical_data_list_into_df(
        ["aad", "session_id", "aad_1", "aad_2", "genfi_version"], TO_COMPLETE_SPECS_DF.copy()
    )
```

test/unittests/converters/test_genfi_to_bids_utils.py
AliceJoubert: No problem with that, you just don't need to use parametrize and 2 different lists here !

clinica/converters/genfi_to_bids/_utils.py
Adam-Ismaili-92: Yes, I think that keeping the function now might be unnecessary, this suggested line will be added to the main function.

test/nonregression/test_run_converters.py
```diff
@@ -45,3 +45,44 @@ def test_converters(cmdopt, tmp_path, study: StudyName):
...
         compare_bids_tsv(output_dir, ref_dir / "bids")
+
+
+@pytest.mark.parametrize(
+    "gif,full",
+    [
+        (False, False),
+        (True, False),
+        (False, True),
+        (True, True),
```
AliceJoubert, line 55: GENFI tests take along time so I would just test the cases that should work and are really different  as we have extensively unit tested the function so (False, True) and (True, True) have the same behavior. If I had to choose maybe it's best to keep (True, True) as it is the one that could cause an issue if the code changes. WDYT ?

test/nonregression/test_run_converters.py
```diff
@@ -45,3 +45,44 @@ def test_converters(cmdopt, tmp_path, study: StudyName):
...
+        path_to_dataset=input_dir / "unorganized",
+        bids_dir=output_dir,
+        path_to_clinical=input_dir / "clinical_data",
+        subjects=input_dir / "subjects.txt",
+        path_to_clinical_txt=input_dir / "additional_clinical_data.txt",
+        gif=gif,
+        full=full,
+    )
+
+    convert(StudyName.GENFI, **kwargs)
```
AliceJoubert:
```suggestion
    study = StudyName.GENFI
    base_dir = Path(cmdopt["input"])
    input_dir, tmp_dir, ref_dir = configure_paths(
        base_dir, tmp_path, get_converter_name(study)
    )
    output_dir = tmp_path / "bids"

    convert(study,
        path_to_dataset=input_dir / "unorganized",
        bids_dir=output_dir,
        path_to_clinical=input_dir / "clinical_data",
        subjects=input_dir / "subjects.txt",
        path_to_clinical_txt=input_dir / "additional_clinical_data.txt",
        gif=gif,
        full=full)
```

test/nonregression/test_run_converters.py
```diff
@@ -45,3 +45,44 @@ def test_converters(cmdopt, tmp_path, study: StudyName):
...
+
+    if full:
+        compare_folders(output_dir, ref_dir / "bids_with_options/bids_full", output_dir)
+        compare_bids_tsv(output_dir, ref_dir / "bids_with_options/bids_full")
+    elif gif:
+        compare_folders(output_dir, ref_dir / "bids_with_options/bids_gif", output_dir)
+        compare_bids_tsv(output_dir, ref_dir / "bids_with_options/bids_gif")
+    else:
+        compare_folders(output_dir, ref_dir / "bids_with_options/bids_cdt", output_dir)
+        compare_bids_tsv(output_dir, ref_dir / "bids_with_options/bids_cdt")
```
AliceJoubert:
```suggestion
    options =  "full" if full else "gif" if gif else "cdt"
    compare_folders(output_dir, ref_dir / "bids_with_options"/f"bids_{options}", output_dir)
    compare_bids_tsv(output_dir, ref_dir / "bids_with_options"/f"bids_{options}")
```

test/nonregression/test_run_converters.py
Adam-Ismaili-92: I also agree that it would be a more general case to test that would make the testing quicker. It should be done in the next commit.

PR #1662, opened 2026-03-25
[FIX] OASIS3 converter adapted to current downloadable clinical data 

davidaquilue:
In this PR I have modified the clinical data loading so that the user can directly point to the currently downloadable OASIS3_data_files directory. This directory contiains a wide variety of .csv files with all the necessary clinical and metadata information to reconstruct very similar tables to the previously used ADRC table.

In this manner, new users of the converter do not need to ask for the workaround to convert their OASIS3 datasets to BIDS. 

Additionally, adds a couple of loggers to print out progress in BIDS conversion, in a similar approach as in the ADNI converter.

PR #1667, opened 2026-03-31
[ENH ]OASIS3-to-BIDS: Adds TAU PET modality to converter

davidaquilue:
Very simple extension after our work to update the OASIS3-to-BIDS converter to the new clinical data in #1662. 

Stemming from the last commit after the PR discussion, we add in a single line the PET Tau modality.


AliceJoubert:
I see your PR (and the other modality one) contains a lot of commits that were added by your first PR we merged already. That was to be expected if you created this branch from the other one before it was merged. The commits would appear twice if we merge now so you need to do a little something to correct that. All you need to do is **rebasing** this PR branch onto the `oasis3_new_modalities` one. I merged the other PR into it so you will its code in it. Tell me if you don't know how to do it, I can help ! 

_Note : I am currently working on refactoring the code (here #1669) after merging yours so you might need to rebase again after I merge_

davidaquilue:
Hello @AliceJoubert , still new to the rebasing, I typically use Git by myself (unfortunately) so have not encountered these situations. I believe I have now successfully rebased the two new branches onto the `oasis3_new_modalities`. However, it still seems like this PR contains a lot of commits. I am unsure whether all these commits will be added on top of your branch...

In worst case scenario, these are very simple PRs, I can re-create the branches and PRs.

Sorry for the inconvenience!

AliceJoubert:
Yep the commits will be added. If the modification is very easy to reproduce I think you can delete and re-create the branches, that would be easy to do !

PR #1673, opened 2026-04-09
[REF] ADNI-to-BIDS : add option to convert other processing steps for all PET modalities

### LINKED ISSUE #1665: [CONV] ADNI-to-BIDS converter does not use step 4 scans for PET TAU
Following user feedback on the converter ADNI-to-BIDS, it was found that for PET TAU data conversion only step 2 scans are considered. DCM2NIIX does not perform well for some scans with step 2 (ex : 041_S_6785, 041_S_6136, 041_S_6159, 041_S_6354) but does for step 4. We should allow conversion of step 4 scans.

AliceJoubert:
closes #1665

What was done : 

- Modality and PETProcessingStep modelisation in a new module `_modality.py`
- Add the possibility to choose which processing step is searched for for pet modalities. Default is ("Co-registered, Averaged"). It corresponds to what we were doing before except for PET FDG which also automatically converted STEP4_8. 
- What it implies : 
    - Filenames changed : the reconstruction is indicated whatever the step now, contrary to before
    - **One step at the same time** for all pet modalities --> if the user wants several, they would have to run the converter several times

TODO after :
- CI Data
- Harmonize Modalities : between user input and ADNIModalityConverter Class + ADNITracer + what is used in `get_images_pet`


clinica/converters/adni_to_bids/cli.py
```diff
@@ -36,6 +36,13 @@
 @click.option(
     "-xml", "--xml_path", help="Path to the root directory containing the xml metadata."
 )
+@click.option(
+    "-pps",
+    "--pet_preprocessing_step",
```
thibaultdvx, line 41: I'm wondering if this name with underscores is coherent with the others (e.g., ``force-new-extraction``, ``clinical-data-only``).

thibaultdvx (commented):
LGTM!

Adam-Ismaili-92:

I have a few remarks based on the tests that have been done.

Firstly, all tests have been conducted on both MacOS and Linux, and show the same results.

The subjects that have been used for testing are the following :
`041_S_6785, 041_S_6136, 041_S_6159, 041_S_6354`

Upon reviewing the results and outputs of this PR, I found that :
- Unit tests work
- CLI help text is showing the new option
- Images are well generated and effectively differ in quality between the different steps (eg: 2 vs 4)

However, a few things need to be corrected : 
  - The CLI help text may be updated to : `[OPTIONAL] ADNI PET **pre**processing step selected for images **conversion** [0 to 5]. See possibilities in the documentation. Default **value** depends on tracer.  [0<=x<=5]
  - The documentation needs to be updated, in particular : `--pet**-**preprocessing**-**step / -pps : this option allows to select the desired **ADNI** pet preprocessing step **which** is by default 2 / "Coregistered, Averaged".` (The last part surely needs to be removed.)
  - It would be interesting to warn the user when no image with the chosen preprocessing step has been found for one particular subject, because no message or error is currently thrown in that regard when there is at least one good subject having the chosen preproc step in the dataset, and the error thrown when all the subjects are wrong is too generic and not really explicit, as you can see below : Capture d’écran 2026-05-20 à 14 20 50
  - The automatic selection of a default value for the preprocessing step does not seem to work, the default seems to be `None`, as you can see below : Capture d’écran 2026-05-20 à 16 48 11

Adam-Ismaili-92 (changes requested):
Remarks and requested changes are described in a previous comment.

AliceJoubert:
Thanks for the review @Adam-Ismaili-92. About the warning, 

I touched a bit the information the user receives to be a little more indicative. However for the end message it is due to clinical data conversion at the end of the pipeline using subjects present in the bids. If nothing was converted it outputs the message you had. I agree it's not informative but IMO starting to rework this does not belong to this PR.

Adam-Ismaili-92 (commented):
Thanks for the corrections @AliceJoubert !

The CLI helps text message and the documentation are indeed better.

However, some work still needs to be done on some issues pointed out in the previous message.

The subjects that have been used for testing are the following :
`041_S_6785, 041_S_6136, 041_S_6159, 041_S_6354`

All tests have been conducted on Linux and MacOS.

Upon reviewing the results and outputs of this PR, I found that :

- Some unit tests work don't work anymore, as you can see below (this error might be outdated due to the recent commits) : Image
- This error may not be related to this PR, but among the subjects tested, the ECAT `.v` images could not be converted : Image
- After excluding the ECAT `.v` images, and using the generic command line for ADNI-to-BIDS `clinica convert adni-to-bids [dataset] [clinical_data] [bids]`, the preprocessing step automatically selected doesn't seem to be the correct one, as no images are converted : Image
The images are nonetheless well produced when the preprocessing step is manually set, by using the values 2 or 4.

PR #1676, opened 2026-04-20
[ENH] OASIS3 FTP Run Selection and Averaging Pipeline

davidaquilue:
As had been previously discussed with @AliceJoubert , this PR adds tackles a challenging issue in processing OASIS3 Tau PET data:
Different acquisition protocols results in different number of runs/files per Tau PET session, which can have a possible number of temporal PET acquisitions, depending on the protocol type.

The final goal is to obtain a final single 3D volume resulting from the co-registration and averaging of the adequate temporal slices, so that the available image has the same characteristics as other datasets (such as the ADNI dataset) and can be forwarded through clinica's PET pipelines.

- Explores available Tau PET scans in OASIS3 (18F-AV1451, Flortaucipir scans)
- Matches scan characteristics to protocol-specific filetype, determining an action (use for SUVR, ignore, extract slices before suvr)
- Applies determined actions, resulting in a single co-registered and average 3D Tau PET volume per available session.

I have the additional comments for reviewers:

1. Currently, pipeline is FTP- and OASIS3- specific. It seems that Amyloid PET also could use of the coreg+avg pipeline part. Should single-dataset-specificity be addressed and generalized?
2. Outputs are stored in a single directory, should the output be structured in BIDS-like structure?
3. Documentation is dataset-specific, where should I add the current [README.md](https://github.com/aramis-lab/clinica/blob/ebe3f9ea60d18762711271409508ca23a07143ca/clinica/pipelines/oasis3_ftp_select_coreg_avg/README.md) file?
4. Tests have not yet been created, will address them after these previous comments are sorted out.

All the best,
David


davidaquilue:
I believe my rebasing was yet again not correct... sorry about that, I think the branches diverged at a previous time point. I could recreate the PR with only latest changes if necessary...

PR #1677, opened 2026-04-20
[ENH] OASIS3 FTP Run Selection and Averaging Pipeline

davidaquilue:
As had been previously discussed with @AliceJoubert , this PR adds tackles a challenging issue in processing OASIS3 Tau PET data:
Different acquisition protocols results in different number of runs/files per Tau PET session, which can have a possible number of temporal PET acquisitions, depending on the protocol type.

The final goal is to obtain a final single 3D volume resulting from the co-registration and averaging of the adequate temporal slices, so that the available image has the same characteristics as other datasets (such as the ADNI dataset) and can be forwarded through clinica's PET pipelines.

Explores available Tau PET scans in OASIS3 (18F-AV1451, Flortaucipir scans)
Matches scan characteristics to protocol-specific filetype, determining an action (use for SUVR, ignore, extract slices before suvr)
Applies determined actions, resulting in a single co-registered and average 3D Tau PET volume per available session.
I have the additional comments for reviewers:

Currently, pipeline is FTP- and OASIS3- specific. It seems that Amyloid PET also could use of the coreg+avg pipeline part. Should single-dataset-specificity be addressed and generalized?
Outputs are stored in a single directory, should the output be structured in BIDS-like structure?
Documentation is dataset-specific, where should I add the current [README.md](https://github.com/aramis-lab/clinica/blob/ebe3f9ea60d18762711271409508ca23a07143ca/clinica/pipelines/oasis3_ftp_select_coreg_avg/README.md) file?
Tests have not yet been created, will address them after these previous comments are sorted out.

All the best,
David


davidaquilue:
Seeing how the OASIS3 PIB and AV45 PET scans are not co-registered and averaged either, I am working on generalizing the pipeline to extend to these other PET modalities.

Let me know if you prefer to directly review the extended version or if you would like to prefer to go step by step and first review the FTP-only pipeline.

Sorry for the spam!

AliceJoubert:
Hello @davidaquilue , thank you very much for your work ! Sorry for closing the PR (I didn't think it would do this). I merged what you already worked on in dev because it is totally finished, and deleted the original branch `oasis3_new_modalities`. This PR closed as a result. If it is not too much trouble, do you think you can reopen it based on `oasis3_iotools` which I just created ? Many thanks. I will do a first review of your work in the meantime

AliceJoubert:
For a first review : 

-  I would prefer if this work was moved to iotools instead of being in pipelines. Pipelines in clinica all use nypipe to order nodes (steps) and parallelize, which would not be the case here so I want to avoid confusion.
- You can put the readme if the documentation part of clinica : docs/IOTools/<your_new _md_file_with_the_pipeline_name>.md . I think you need to install the requirements for documentation with a `poetry install --with docs`. Tell me if this is not clear.
- When you say it is specific to OASIS3 and that Amyloid PET could benefit from it I am not sure I understand since there is PET amyloid data in OASIS3 ? 
- Though, if the input is a BIDS maybe you can generalize the tool, which we don't necessarily require. I guess what would change is the manual review of the scans you did that a user would have to perform by themselves to apply it to a new dataset. For now let's just review as is and we can decide later !
- For the output we want to keep the BIDS structure. A possibility would be to copy the input BIDS to the output path and just change the relevant pet files. We do something similar with the tool `center-nifti`

davidaquilue:
Hello @AliceJoubert 


> For a first review :
> 
> * I would prefer if this work was moved to iotools instead of being in pipelines. Pipelines in clinica all use nypipe to order nodes (steps) and parallelize, which would not be the case here so I want to avoid confusion.
> * You can put the readme if the documentation part of clinica : docs/IOTools/<your_new _md_file_with_the_pipeline_name>.md . I think you need to install the requirements for documentation with a `poetry install --with docs`. Tell me if this is not clear.

okay, good!

> * When you say it is specific to OASIS3 and that Amyloid PET could benefit from it I am not sure I understand since there is PET amyloid data in OASIS3 ?

What I meant is that the pipeline is only adapted to the Tau PET in OASIS3. Therefore, it can be easily extended to the PET existing amyloid data in OASIS3.

> * Though, if the input is a BIDS maybe you can generalize the tool, which we don't necessarily require. I guess what would change is the manual review of the scans you did that a user would have to perform by themselves to apply it to a new dataset. For now let's just review as is and we can decide later !

In the end, I managed to "hardcode" the manual review of the scans for the OASIS3. To generalize to any BIDS dataset, this hardcoded part should be modified. I do not know exactly how that could be made, I guess with some type of JSON, but I can point you later on to the part where the decisions are hardcoded and maybe your team can see how to adapt it. From my side, I will keep it as dataset-agnostic as possible.

> * For the output we want to keep the BIDS structure. A possibility would be to copy the input BIDS to the output path and just change the relevant pet files. We do something similar with the tool `center-nifti`

Okay, will take a look!

I will directly add a PR for the generic pipeline so that you can review it.

Thanks!

PR #1690, opened 2026-04-23
[ENH] OASIS3 PET IO - Run Selection and Averaging Pipeline

davidaquilue:
As discussed in #1677 with @AliceJoubert , this PR tackles a challenging issue in processing OASIS3 PET data:
different acquisition protocols results in different number of runs/files per PET session, which can have a possible number of temporal PET acquisitions, depending on the protocol type.

Since typical PET processing pipelines require to obtain a final single 3D volume resulting from the co-registration and averaging of adequate temporal slices, we implement here a solution for the OASIS3 dataset to transform the raw downloadable data into images with the same characteristics as other datasets (such as the ADNI dataset) and can be forwarded through clinica's PET pipelines.

1. Explores available PET scans of the requested tracer (PIB, AV45, AV1451) in OASIS3
2. Matches scan characteristics to protocol-specific filetype, determining an action (use for SUVR, ignore, extract slices before suvr)
3. Applies determined actions, resulting in a single co-registered and average 3D Tau PET volume per available session.

Remaining comments to be addressed or implemented:
- [x] Store documentation in ``docs/IOTools/``
- [ ] Tests have not yet been created, will address them after these previous comments are sorted out.
- [ ] Improve storage of outputs, in BIDS-formatted derivative with copy of the input. Check the ``center-nifti`` function.

All the best,
David


clinica/iotools/oasis3_pet_select_coreg_avg/__init__.py
```diff
@@ -0,0 +1 @@
+
```
AliceJoubert, line 1: In `init` you can put all the function that you wish to expose to the user. Here I would say it's run_pipeline, so  : 


```python
from ._pipeline import run_pipeline

__all__ = ["run_pipeline"]
```

:warning: `run_pipeline` is a bit too generic a name. maybe something like `oasis3_scan_selection_averaging` would be better. WDYT ?

clinica/iotools/oasis3_pet_select_coreg_avg/__main__.py
```diff
@@ -0,0 +1,3 @@
+from .cli import cli
```
AliceJoubert, line 1: This will be handled by the file `iotools/cli.py` so you can delete the main here !

clinica/iotools/oasis3_pet_select_coreg_avg/cli.py
```diff
@@ -0,0 +1,68 @@
+from pathlib import Path
```
AliceJoubert, line 1: You can move all this to iotools/cli.py, with the following considerations : 
- use `@cli.command()` to allow the pipeline to be found by the client (verify this by typing `clinica iotools --help` --> if the new pipeline doesn't appear it won't work)
- there is already a bids_directory option defined at the top of the file so you can replace bids_dir by bids directory (`@bids_directory`, name to change in your function)
- same remark as in the close PR for inventory dir
- same remark as in the closed PR for bids_dir, you can validate it with the function check_bids_dataset from clinica.dataset.bids._validation
- `tracer:  Union[str, SUPPORTED_TRACERS]`
- :warning: `oasis3-pet-pipeline` is also too generic of a name, maybe you can put the same name as the function previously (🙏) called `run_pipeline`

AliceJoubert (commented):
I think I will need to iterate several times to avoid drowning you in information so bear with me ! Thanks already for making the requested change. Here are some pointers to make the client work

docs/IOTools/OASIS3_PET_Run_Selection.md
```diff
@@ -0,0 +1,127 @@
+# OASIS-3 PET Run Selection Pipeline
```
AliceJoubert, line 1: If you want to add this page to the website, you also need to indicate it in the `mkdocs.yml` page, in the `nav` section (end of file). Then, you will be able to deploy the documentation and see the outcome with : `poetry run mkdocs serve`

clinica/iotools/oasis3_pet_select_coreg_avg/cli.py
```diff
@@ -0,0 +1,68 @@
...
+@click.argument(
+    "output_dir",
+    type=click.Path(writable=True, file_okay=False, resolve_path=True),
+)
+@click.option(
+    "--tracer",
+    type=click.Choice(SUPPORTED_TRACERS, case_sensitive=False),
+    required=True,
+    help="PET tracer to process (AV1451, AV45, or PIB).",
+)
```
AliceJoubert, line 23: If you use "required" then it's not an option anymore, it's an argument (like the input and output). You can also put a default value

clinica/iotools/oasis3_pet_select_coreg_avg/_tracer_config.py
```diff
@@ -0,0 +1,234 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Callable
+
+from clinica.utils.pet import Tracer
+
+
+@dataclass(frozen=True)
+class TracerConfig:
```
AliceJoubert, line 18: Technically it works. If you have time, I would suggest : 
- passing `bids_tracer_tag`, as a property built on `self.tracer.value`
- passing `classify_fn` as an abstract class method (requires building on class abc.ABC), which would be overridden in children classes.


```python
@dataclass(frozen=True)
class Oasis3TracerConfig(ABC):
    ....
    @abstractmethod
    def classify_fn(...)->...:
        pass

class AV1451Config(Oasis3TracerConfig):
    tracer = ...
    late_phase_start_min = ...
    late_phase_end_min = ...

    def classify_fn(...)->...:
        # what was in classify_av1451

```
    
- making use of `Enum` :
    - invalid values are handled automatically
    - no need to use `clinica.pet.utils.Tracer` anymore
    - you can directly use this class in the client

```python
class Oasis3SupportedTracers(str, Enum):
    "AV1451" = AV1451
    "AV45" = AV45
    "PIB" = PIB

def get_tracer_config(tracer : Oasis3SupportedTracers)->Oasis3TracerConfig:
    if tracer == Oasis3SupportedTracers.PIB: 
        return PIBConfig
    if tracer == Oasis3SupportedTracers.AV1451: 
        return AV1451Config
    if tracer == Oasis3SupportedTracers.AV45: 
        return AV45Config
```

clinica/iotools/oasis3_pet_select_coreg_avg/_session_filtering.py
```diff
@@ -0,0 +1,78 @@
+"""Session-level usability filtering and cprint-based logging."""
+
+from __future__ import annotations
```
AliceJoubert, line 3:
```suggestion
```

This is deprecated in python 3.14 so we should not use it ! if it causes issues we will find other ways to handle it

clinica/iotools/oasis3_pet_select_coreg_avg/_scan_classification.py
```diff
@@ -0,0 +1,163 @@
+"""PET scan classification: frame-timing parsing and protocol classification."""
+
+from __future__ import annotations
```
AliceJoubert, line 3:
```suggestion
```

clinica/iotools/oasis3_pet_select_coreg_avg/_scan_classification.py
```diff
@@ -0,0 +1,163 @@
...
+    values = None
+    for parent_key in ("FrameTimes", "Time"):
+        ft = metadata.get(parent_key)
+        if ft and isinstance(ft, dict):
+            inner = ft.get("FrameTimes")
+            if inner and isinstance(inner, dict):
+                values = inner.get("Values")
+                if values:
+                    break
+    if not values or not isinstance(values, list) or len(values) == 0:
```
AliceJoubert, line 45: What is your expectations for `values` ? If I understand it should be a list, not empty (not a single value for ex).

Do you have an example of subject where metadata uses the "Time" parent_key ?

clinica/iotools/oasis3_pet_select_coreg_avg/_scan_classification.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    values = None
+    for parent_key in ("FrameTimes", "Time"):
+        ft = metadata.get(parent_key)
+        if ft and isinstance(ft, dict):
+            inner = ft.get("FrameTimes")
+            if inner and isinstance(inner, dict):
+                values = inner.get("Values")
+                if values:
+                    break
```
AliceJoubert, line 44: Generally we avoid using break as much as possible : 

```python
try : 
    values = metadata.get(parent_key).get("FrameTimes").get("Values")
except AttributeError:
    continue
```

clinica/iotools/oasis3_pet_select_coreg_avg/_scan_classification.py
```diff
@@ -0,0 +1,163 @@
...
+    current_dur = durations[0]
+    current_count = 1
+    for d in durations[1:]:
+        if d == current_dur:
+            current_count += 1
+        else:
+            dur_counts.append(f"{current_count}x{current_dur:.0f}s")
+            current_dur = d
+            current_count = 1
+    dur_counts.append(f"{current_count}x{current_dur:.0f}s")
```
AliceJoubert, line 65: This could be a helper function ; though can you explain your logic behind it ? Maybe there is an easier way to do what you want

clinica/iotools/oasis3_pet_select_coreg_avg/_scan_classification.py
```diff
@@ -0,0 +1,163 @@
...
+    for json_path in sorted(pet_dir.glob(f"*{tracer_cfg.bids_tracer_tag}*_pet.json")):
+        with open(json_path, "r", errors="ignore") as f:
+            try:
+                metadata = json.load(f)
+            except json.JSONDecodeError:
+                continue
+
+        modality = metadata.get("Modality", "PET")
+        frame_info = parse_frame_times(metadata)
+        nii_frames = None if frame_info is not None else get_nii_frame_count(json_path)
```
AliceJoubert, line 108: Can you also explain the logic behind this line to me ? So I am sure I understood properly (like, the info is either in the json or the nifti ?)

AliceJoubert: Additionnally, if nii_frames is None should it be added to records ?

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+    session_id = row["Session_ID"]
+    file_name = row["File_Name"]
+    action = row["Action"]
+    tag = f"{subject_id}/{session_id}/{file_name}"
+
+    if bids_output:
+        out_path = _bids_output_path(bids_dir, subject_id, session_id, file_name)
+    else:
+        out_name = Path(file_name).stem + "_coreg_avg.nii.gz"
+        out_path = output_dir / out_name
```
AliceJoubert, line 137: bids_output is mandatory here

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+    if out_path.exists():
+        cprint(f"[EXISTS] Already processed, skipping: {tag}", lvl="debug")
+        return True
+
+    json_path = find_bids_json(bids_dir, subject_id, session_id, file_name)
+    if json_path is None:
+        cprint(f"[SKIP] JSON not found: {tag}", lvl="warning")
+        return False
```
AliceJoubert, line 145: Unless I missed something if the json is missing it will cause issues earlier on  because the first steps of the pipelines are based on json keys ?

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+
+    nii_path = find_bids_nifti(json_path)
+    if nii_path is None:
+        cprint(f"[SKIP] NIfTI not found: {tag}", lvl="warning")
+        return False
```
AliceJoubert, line 150: Same here you already check for the nifti earlier, no ?

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+        metadata = json.load(f)
+
+    values = None
+    for parent_key in ("FrameTimes", "Time"):
+        ft = metadata.get(parent_key)
+        if ft and isinstance(ft, dict):
+            inner = ft.get("FrameTimes")
+            if inner and isinstance(inner, dict):
+                values = inner.get("Values")
+                if values:
```
AliceJoubert, line 45: You can also use the try / except syntax here

clinica/iotools/oasis3_pet_select_coreg_avg/_nii_reading.py
```diff
@@ -0,0 +1,92 @@
...
+    if frame_indices is None:
+        frame_indices = list(range(data.shape[3]))
+
+    affine = img.affine
+    header = img.header.copy()
+    frames = [
+        sanitize_nifti_header(nib.Nifti1Image(data[..., idx], affine, header))
+        for idx in frame_indices
+    ]
+    return frames, frame_indices
```
AliceJoubert, line 92: Do you need to be returning frame_indices ?

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
+"""PET image processing: late-frame extraction, coregistration, and averaging."""
+
+from __future__ import annotations
```
AliceJoubert, line 3:
```suggestion
```

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+def coregister_and_average(frames: list[nib.Nifti1Image]) -> nib.Nifti1Image:
+    """Rigid-body coregister all frames to the first frame, then average them.
+
+    When only one frame is provided it is returned unchanged.
+    """
+    if len(frames) == 1:
+        return frames[0]
+
+    fixed_nib = frames[0]
+    fixed_ants = ants.from_nibabel(fixed_nib)
```
AliceJoubert: All the `ants.from_nibabel` or `ants.to_nibabel` won't work because these methods are from ants.utils

AliceJoubert (commented):
Hello @davidaquilue ,

Lots of comments/questions from my part. Take your time going over them :-)

davidaquilue:
TODO: Modify the PIB windows, as per the OASIS3 Data Dictionary, most analyses on the OASIS take the frames in the 30-60 windows, instead of the 40-60.

clinica/iotools/oasis3_pet_select_coreg_avg/_pet_processing.py
```diff
@@ -0,0 +1,205 @@
...
+    fixed_nib = frames[0]
+    fixed_ants = ants.from_nibabel_nifti(fixed_nib)
+
+    aligned_data = [np.asarray(fixed_nib.dataobj, dtype=np.float32)]
+    for frame_nib in frames[1:]:
+        moving_ants = ants.from_nibabel_nifti(frame_nib)
+        result = ants.registration(
+            fixed=fixed_ants,
+            moving=moving_ants,
+            type_of_transform="Affine",
```
jianfeixu95, line 71: Hi, @davidaquilue 

Probably, the type_of_transform parameter value might be 'Rigid'. _type_of_transform="Rigid"_

This step is intra-acquisition frame-to-frame coregistration (same subject, same session), a 6-DOF rigid transform should be sufficient — the brain doesn't scale or shear across frames within one scan. The extra 6 affine DOFs can fit noise in low-SNR early frames and introduce spurious scaling/shearing in the averaged image.

Some papers for your reference:
- [Costes N, Dagher A, Larcher K, Evans AC, Collins DL, Reilhac A. Motion correction of multi-frame PET data in neuroreceptor mapping: simulation based validation. Neuroimage. 2009 Oct 1;47(4):1496-505. doi: 10.1016/j.neuroimage.2009.05.052. Epub 2009 May 27. PMID: 19481154.](https://pubmed.ncbi.nlm.nih.gov/19481154/)
- [A Rapid and Accurate Method to Realign PET Scans Utilizing Image Edge Information](https://jnm.snmjournals.org/content/36/4/657)

CHEERS,

PR #1705, opened 2026-06-22
[DOC] Update information for mrtrix3 and convert3d

AliceJoubert:
## Description
The conda channels used for these tools do not support Mac (ARM) installation. Documentation was modified to indicate how to install clinica considering this. An additional yml for conda environment was added to handle this specific case.

## To Check :
- Command line : `conda env create -n test -f environment(...).yml`
  - [x] Works for Linux / Mac Intel (should not change existing behavior)
  - [x] Now works for Mac ARM
- Documentation builds properly

### DIFF
--- a/docs/Software/Installation.md
+++ b/docs/Software/Installation.md
@@ -18,9 +18,15 @@ To install Miniconda, open a new terminal and type the following commands:
     bash /tmp/miniconda-installer.sh
     ```
 
-=== "MacOS"
+=== "MacOS (Intell)"
     ```{.sourceCode .bash .copy}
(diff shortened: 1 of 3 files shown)


Adam-Ismaili-92 (commented):

The documentation builds properly.

And the command `conda env create -n test -f environment-macarm.yml -k` works well on both Linux and MacOS.

However, the command `conda env create -n test -f environment.yml -k` doesn't work on Linux on the ICM Workstations, is it normal? (I am guessing that the firewall policies could be the problem)

Apart from that, LGTM !

Adam-Ismaili-92 (approved):

PR #1712, opened 2026-07-21
[FIX] PET-Volume : Add `build_petpvc_name` as a task

AliceJoubert:
## Description
When using option `--pvc_psf_tsv` with an actual file, code crashes because of improper  Nipype formatting (input name, type). This option is currently not tested in the CI.

## Key Changes Made
- Added the function `build_pet_pvc_name` from pet surface utils in the tasks with type checking compatible with Nipype
- Fixed input format for one node

## Checklist
- [x] Coding style respected (eg: '_' at the beginning of variable names, private or public functions, docstrings, ...)
- [ ] Unit tests added and passing
- [x] Functional testing completed (on Linux and macOS)
- [x] Instantiation tests passing

## How to Test

- Run the pet-volume pipeline with the option `--pvc_psf_tsv`

### DIFF
--- a/clinica/pipelines/pet/volume/pipeline.py
+++ b/clinica/pipelines/pet/volume/pipeline.py
@@ -473,13 +473,13 @@ class PETVolume(GroupPipeline, PETPipeline):
 
         from .tasks import (
             apply_binary_mask_task,
+            build_pet_pvc_name_task,
             compute_atlas_statistics_task,
             create_binary_mask_task,
             create_pvc_mask_task,
             normalize_to_reference_task,
         )
         from .utils import (
-            build_pet_pvc_name,
             get_from_list,
             init_input_node,
         )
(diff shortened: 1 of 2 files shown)


Adam-Ismaili-92:
This PR has been tested on both MacOS and Linux:

- This command has been used on both os: `clinica run pet-volume [BIDS] [CAPS] [GROUP] AV45 PONS --pvc_psf_tsv [PVC_PSF_TSV]`, and works well in each case.
- The unit tests are all passing.
- The non regression tests are all passing.
- The code seems to respect the coding style.

I have just one concern: several warnings regarding wrong calculation values which do not appear without using the option are visible in the logs, is this normal?
Capture d’écran 2026-08-25 à 11 27 05

Apart from that, LGTM, thanks for your work @AliceJoubert !

PR #1715, opened 2026-07-23
[REF] PET-Linear : Integrate Hugues Roy's work

AliceJoubert:
## Description
Based on Hugues Roy's work. Removes usage of Synquick script for rigid registration. To prevent the skull intensities from messing with the registration, a new "skull-stripping" (multiplication between registered-to-T1 MNI and T1) node is added. 

## Key Changes Made
- Introduce a node for skull-stripping using Ants Math method
- Removes Synquick Registration script in favor of the regular Ants Registration method

## Checklist
- [x] Coding style respected (eg: '_' at the beginning of variable names, private or public functions, docstrings, ...)
- [x] Unit tests added and passing
- [x] Functional testing completed (on Linux and macOS)
- [x] Instantiation tests passing
- [x] Non-regression tests passing (on Linux and macOS)
- [x] CI database updated (if required)
- [ ] Documentation updated (if required)

## Further work
When T1-Linear integrates skull-stripping, its output could be used instead of using a MNI reference here.

### DIFF
--- a/clinica/pipelines/pet/linear/pipeline.py
+++ b/clinica/pipelines/pet/linear/pipeline.py
@@ -315,6 +315,7 @@ class PETLinear(PETPipeline):
 
         from .tasks import (
             clip_task,
+            get_item_from_list,
             get_skull_stripping_from_reference_task,
             perform_suvr_normalization_task,
         )
@@ -376,7 +377,6 @@ class PETLinear(PETPipeline):
         ants_registration_node = npe.Node(
             name="antsRegistration", interface=ants.Registration()
         )
-        ants_registration_node.inputs.write_composite_transform = True
         ## image dimension
         ants_registration_node.inputs.dimension = 3
         ## type of transform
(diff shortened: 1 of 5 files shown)


Adam-Ismaili-92 (commented):
This PR has been tested on both MacOS and Linux.

On both systems:
- The unit tests worked well.

However, some errors occurred:
- The non regression test regarding PET Linear could not terminate, running indefinitely.
- The same situation happened for the manual test done using selected subjects and sessions. The encountered error is described in the screenshot below:
Image

It seems that the line `output_value = func(first_arg, *list(args))` returned an error because of a missing file in the working directory that has not been produced: `/WD/pet-linear/117019e1ccae174dc04f3d9db0edd60d235a3642/renameFileCAPS/result_renameFileCAPS.pklz`. 
The folder `renameFileCAPS` is also absent from the working directory.
- Output images could not be compared since no output image was generated using PET Linear from this PR.

To further help to debug, here are the hierarchy of the BIDS and CAPS folders:
Image
Capture d’écran 2026-09-04 à 11 54 35

PR #1719, opened 2026-08-03
[DOC] Simplify ITK and PETPVC download instructions

AliceJoubert:
## Key Changes Made

- Group together ITK and PETPVC download instructions
- Indicate easiest way to install them

## Checklist
- [x] Documentation updated (if required)

## How to Test

- Use a bare OS (Mac or Linux), meaning without a prior installation of ITK or PETPVC.
- Create a conda environment with clinica + petpvc in it (just follow the clinica documentation for this step)
- Test that a pipeline like pet-volume or pet-surface still works regarding petpvc steps.

### DIFF
--- a/docs/Software/Third-party.md
+++ b/docs/Software/Third-party.md
@@ -169,9 +169,25 @@ We recommend installing [**FSL 6.0**](https://fsl.fmrib.ox.ac.uk/).
 
 ___
 
-### ITK
+### PETPVC + ITK
 
-You can find more details about `ITK` on their [website](https://itk.org/). To install it :
+In clinica `PETPVC` (more info on their [github](https://github.com/UCL/PETPVC)) and `ITK` (more info on their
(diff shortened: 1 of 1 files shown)


Adam-Ismaili-92 (approved):
I still approved the PR because I don't think that point is a deal-breaker, but wouldn't it be better to have two separate blocks for PETPVC and ITK? That way, users wouldn't get confused about how to download them independently, as it has been done in the other blocks, even though they are linked regarding their use within Clinica.

AliceJoubert:
> I still approved the PR because I don't think that point is a deal-breaker, but wouldn't it be better to have two separate blocks for PETPVC and ITK? That way, users wouldn't get confused about how to download them independently, as it has been done in the other blocks, even though they are linked regarding their use within Clinica.

We are giving tips to download external software because they are linked to Clinica, not in any use case. Also ITK is optional if you download PETPVC the "right" way so that made sense to group them together.

PR #1723, opened 2026-09-07
[FIX] FLAIR/T1-Linear: use brain-only references for skull-stripped images

### LINKED ISSUE #1720: [BUG] FLAIR/T1-Linear : reference for already skull-stripped images
**Describe the bug**
If inputting already skull-stripped images (ex for FLAIR: Brats) the registration can go wrong as it will consider the skull of the. This is not systematic. 

**To reproduce**
Run the pipeline on an already skull-stripped image. Ex Brats n°0006. ⚠️ In this case the bias field correction was skipped.

**To fix**
Add an option to the client
Give a skull-stripped reference to the registration node

**To go further**
These anat pipelines are not very flexible. One could want only the registration for example. Maybe we could

Sawlyer:
Closes #1720

When processing already skull-stripped T1 or FLAIR images, the linear pipelines can register against a reference that still contains the skull.

This PR adds a `--skull-stripped` option to `t1-linear` and `flair-linear`. When enabled, Clinica uses brain-only MNI references for registration. The references are included in the package and retrieved through Clinica’s existing checksum-verified download mechanism if they are not available locally.

The default behavior remains unchanged.


AliceJoubert:
Hello @Sawlyer, thank you very much for your contribution ! I will be reviewing it this week. Best,

